Issue #7, opened 2017-06-13
Handle duplicate video file names


mbertrand:
An uploaded file will replace any existing video file with the same name, regardless of user.  A duplicate file name should be modified slightly (append _01, _02, for example); perhaps also place videos in a S3 subfolder for each user.

Issue #2, opened 2017-06-13
'Access Denied' for signed URL's to uploaded video files with spaces in name


mbertrand:
Steps to reproduce:

- Upload a video file with spaces in its name, like 'Linus and Lucy.mp4'
- Generate a signed URL for the video
- Open the signed URL in a browser

```
<Error>
<Code>AccessDenied</Code>
    <Message>Access denied</Message>
</Error>
```

PR #11, opened 2017-06-16
Fix filename-related issues

mbertrand:
closes https://github.com/mitodl/odl-video-service/issues/2 
closes https://github.com/mitodl/odl-video-service/issues/7

#### What's this PR do?
- Avoid overwriting of videos/S3 files by ensuring each file upload has a unique filename/S3 key
  - Use virtual S3 subfolders for each user id
  - Add a unique UUID to the S3 key
  - Example: "1/9fa9d051-bb10-4698-a9c9-fcd7c622b589/lastjedi.mp4"
  - Creates a new model 'VideoFile', and moves most of the Video model's S3-related functionality to that.
- Use urllib.quote(s3-key) when generating cloudfront URL's to prevent 'Access Denied' error
- Move/rename test files

You'd need to get this running in docker (docker-compose build, docker-compose up) but that requires an MIT X509 certificate and AWS access keys as describe in the README.  There are several new unit tests in this PR.


ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
-        """s3 url"""
-        bucket_name = os.environ.get("VIDEO_S3_BUCKET", "odl-video-service")
-        encoded_key = urllib.parse.quote_plus(self.s3_object_key)
+        """
+        URL for the Video S3 object
+
+        Returns:
+            str: URL
+        """
         return "https://s3.amazonaws.com/{bucket}/{key}".format(
```
giocalitri: this is defaulting to US East Region: shouldn't make this configurable and move it to the Settings?

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
     @property
     def cloudfront_url(self):
-        """cloudfront url"""
+        """
+        Get the Cloudfront URL for the video
+
+        Returns:
+            str: Cloudfront unsigned URL
+        """
         distribution = os.environ.get("VIDEO_CLOUDFRONT_DIST")
```
giocalitri: This should be moved to the settings, but feel free to open an issue to address all these cases in a separated PR

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
+            key=self.s3_object_key
         )
 
+    @property
+    def cloudfront_presigned_url(self):
+        """
+        Get a signed Cloudfront URL with a default expiration date of 2 hours from
+        when this property is called.
+        """
+        expires = datetime.datetime.utcnow() + datetime.timedelta(hours=2)
```
giocalitri: I would use an approach similar to MM or bootcamp to define times in UTC

giocalitri: Unless there is a specific use for having the `cloudfront_signed_url` method, we should probably move the code here

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
+            str: Cloudfront signed URL
+        """
         return make_cloudfront_signed_url(
             key=self.s3_object_key,
             expires=expires,
         )
 
     def __str__(self):
-        return self.title or "<untitled video>"
+        return self.s3_object_key
```
giocalitri: not sure why you made the `__str__` so minimal and the `__repr__` that detailed, but I would make them pretty similar

ui/serializers.py
```diff
@@ -51,10 +51,19 @@ class VideoSerializer(serializers.ModelSerializer):
     class Meta:
         model = Video
         fields = (
-            'id', 'created_at', 'title', 'description', 'moira_lists',
-            's3_object_key',
+            'id', 'created_at', 'title', 'description', 'moira_lists', 'videofile_set',
```
giocalitri: Please, one entry per line or just one line (it should be long enough)

ui/serializers.py
```diff
@@ -51,10 +51,19 @@ class VideoSerializer(serializers.ModelSerializer):
...
-        read_only_fields = ('id', 'created_at', 's3_object_key')
+        read_only_fields = ('id', 'created_at', 'videofile_set')
+
+
+class VideoFileSerializer(serializers.ModelSerializer):
+    """Video File Serializer"""
+    class Meta:
+        model = VideoFile
+        fields = (
+            'id', 'created_at', 's3_object_key', 'encoding', 'bucket_name'
```
giocalitri: Please, one entry per line or just one line (it should be long enough)

ui/util.py
```diff
@@ -86,7 +88,7 @@ def cloudfront_signed_url(key, expires):
     if not cloudfront_dist:
         raise RuntimeError("Missing required env var: VIDEO_CLOUDFRONT_DIST")
     url = "https://{dist}.cloudfront.net/{key}".format(
-        dist=cloudfront_dist, key=key,
+        dist=cloudfront_dist, key=quote(key),
```
giocalitri: Please, one entry per line or just one line

ui/views.py
```diff
@@ -79,22 +79,22 @@ def stream(request):
...
-        video, _ = Video.objects.get_or_create(
-            s3_object_key=dropbox_file["name"],
-            defaults={
-                "source_url": dropbox_file["link"],
-                "creator": request.user,
-            }
+        video = Video.objects.create(
+            source_url=dropbox_file["link"],
+            creator=request.user,
+            title=dropbox_file["name"]
```
giocalitri: what happens if this raises for any reason?

ui/views.py
```diff
@@ -79,22 +79,22 @@ def stream(request):
...
-        video.save()
-        task_result = stream_to_s3.delay(video.source_url)
+
+        task_result = stream_to_s3.delay(video.source_url, video.s3_key)
         response[video.id] = {
-            "key": video.s3_object_key,
+            "key": video.s3_key,
             "task": task_result.id,
         }
+        VideoFile.objects.get_or_create(
```
giocalitri: why `get_or_create`?

ui/views.py
```diff
@@ -107,17 +107,19 @@ class VideoViewSet(viewsets.ModelViewSet):
     def destroy(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         video = self.get_object()
         if request.GET.get("s3"):
-            video.s3_object.delete()
+            for file in video.videofile_set.all():
+                file.s3_object.delete()
```
giocalitri: why do you need to delete them one at the time in a loop?

ui/views.py
```diff
@@ -107,17 +107,19 @@ class VideoViewSet(viewsets.ModelViewSet):
...
         self.perform_destroy(video)
         return Response(status=status.HTTP_204_NO_CONTENT)
 
-    @detail_route(permission_classes=(IsAdminOrHasMoiraPermissions,))
-    def signed_url(self, request, pk=None):  # pylint: disable=unused-argument
+    @detail_route(permission_classes=(IsAdminOrHasMoiraPermissions,), url_path=r'signed_url/(?P<encoding>\w+)')
+    def signed_url(self, request, pk=None, encoding=None):  # pylint: disable=unused-argument
```
giocalitri: I do not understand this: what is `pk` and why both `pk` and `encoding` are defaulted to `None`?

giocalitri:
addresses the issues you liked, you should use text like
```
closes #2
closes #7
```
so when this PR is merged the liked issues get closed automatically

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
     description = models.TextField(blank=True)
     source_url = models.URLField()
     moira_lists = models.ManyToManyField(MoiraList)
+    status = models.TextField(max_length=24, blank=True)
+    s3_subkey = models.UUIDField(unique=True, null=False, blank=False, default=uuid4)
+
+    @property
+    def s3_key(self):
+        """
+        Avoid duplicate S3 keys/filenames when transferring videos rom Dropbox
```
mbertrand: It is used to [generate the 's3_object_key' for a new VideoFile object](https://github.com/mitodl/odl-video-service/blob/mb/filebugs/ui/views.py#L94) representing the file being uploaded to S3.

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
+        return newkey
+
+
+class VideoFile(models.Model):
+    """
+    A file associated with a Video object, either the original upload or a transcoded file.
+    """
+    created_at = models.DateTimeField(auto_now_add=True)
+    s3_object_key = models.TextField(unique=True)
+    bucket_name = models.CharField(max_length=128, blank=False, null=False)
```
mbertrand: Looks like [bucket names can't be more than 63 characters long](http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html) so I will change it to that.

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
+    video = models.ForeignKey(Video, on_delete=models.CASCADE)
+    encoding = models.CharField(max_length=128, default='original')
```
mbertrand: The name will be up to us, and associated with an AWS preset ID (up to AWS).  To be stored in settings.py from environment variable.  Example:

```
ET_PRESET_IDS=small:1351620000001-000061,medium:1351620000001-000040,large:1351620000001-000020
```

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
+    preset_id = models.CharField(max_length=128, blank=True, null=True)
```
mbertrand: preset_id will be the AWS encoding preset used for the VideoFile, not used here yet.  I can take it out from this PR but it will be needed for the transcoding PR.

ui/models.py
```diff
@@ -42,45 +44,104 @@ class Video(models.Model):
...
-        bucket_name = os.environ.get("VIDEO_S3_BUCKET", "odl-video-service")
+        """
+        Get the S3 Object for the video
+
+            Returns:
+                s3.Object: Video file's S3 object
+        """
         s3 = boto3.resource('s3')
-        return s3.Object(bucket_name, self.s3_object_key)
+        return s3.Object(self.bucket_name, self.s3_object_key)
```
mbertrand: If bucket name or s3_object_key is None then an exception will be raised here.  I can add "null=False, blank=False" to both fields.  No exception will be raised here if the object doesn't exist  on S3.

ui/models.py
mbertrand: Sure, will do

ui/views.py
mbertrand: The upload will continue via celery but no VideoFile object will be created for it (and currently an 'Error' popup shows on the screen.  Maybe it would be better to try creating the VideoFile object before the upload starts, immediately after the Video object is created, and if an exception is raised here for some reason, catch it, delete the Video object, then raise the exception and don't proceed with the upload.

ui/views.py
mbertrand: `.create` would make more sense, I'll change it.

ui/views.py
mbertrand: In my transcode branch I removed this and made it a signal, should I do that in this PR?
```
@receiver(pre_delete, sender=VideoFile)
@receiver(pre_delete, sender=VideoThumbnail)
def delete_s3_files(sender, **kwargs):
    """
    Make sure S3 files are deleted along with associated video file/thumbnail object
    """
    kwargs['instance'].s3_object.delete()
```

ui/views.py
mbertrand: pk is the id of the Video object.  I could make both 'pk' and 'encoding' args instead of kwargs:
```def signed_url(self, request, pk, encoding)```

This is for generating a REST URL, for example: http://192.168.99.101:8079/api/videos/3/signed_url/original/

File: ui/models.py (hunk 3 above)
Comment (mbertrand) [inline ui/models.py]: Like [this](https://github.com/mitodl/micromasters/blob/master/micromasters/utils.py#L225)?
```
def now_in_utc():
    """
    Get the current time in UTC
    Returns:
        datetime.datetime: A datetime object for the current time
    """
    return datetime.datetime.now(tz=pytz.UTC)
```

Comment (mbertrand) [inline ui/models.py]: The cloudfront_signed_url method allows for an expiration date as input, but I don't think we'll need that as far as I know so I will get rid of it and move its code here.

File: ui/models.py (hunk 12 above)
Comment (giocalitri) [inline ui/models.py]: ok... I asked because maybe it is not the best to have it as a property: it looks more like a function to generate an unique key.
Being a property it is more like the key of an object that already exists (not sure I am clear).
Anyway feel free to leave as is if you think it is clear enough

File: ui/models.py (hunk 14 above)
Comment (giocalitri) [inline ui/models.py]: well... if the names are up to us, I would define them formally in a class.

File: ui/models.py (hunk 16 above)
Comment (giocalitri) [inline ui/models.py]: the question is mostly about if we need to handle the exception

File: ui/models.py (hunk 3 above)
Comment (giocalitri) [inline ui/models.py]: well... we do not need the separated function, but yes

File: ui/views.py (hunk 8 above)
Comment (giocalitri) [inline ui/views.py]: yes, I agree

File: ui/views.py (hunk 10 above)
Comment (giocalitri) [inline ui/views.py]: no, if you have done the work in another branch, keep it

File: cloudsync/tasks.py [hunk 17]
```diff
@@ -1,16 +1,14 @@
 """
 Tasks for cloudsync app
 """
-
 import os
 import re
-from urllib.parse import unquote
-
-import boto3
+import urllib
```
giocalitri: why this change?

ui/conftest.py
```diff
@@ -25,10 +27,25 @@ def video(user):  # pylint: disable=redefined-outer-name
...
+@pytest.mark.django_db
+def videofile(video):  # pylint: disable=redefined-outer-name
+    """
+    Fixture to create a video file
+    """
+    obj = VideoFile(
+        video=video,
+        s3_object_key=video.s3_key,
+        encoding='original',
+        bucket_name=settings.VIDEO_S3_BUCKET
```
giocalitri: do we have 1 bucket for all videos and thumbnails?

ui/serializers.py
```diff
@@ -67,8 +71,5 @@ class DropboxFileSerializer(serializers.Serializer):
     isDir = serializers.BooleanField()
 
     def create(self, validated_data):
-        video, _ = Video.objects.get_or_create(
-            s3_object_key=validated_data["name"],
-            defaults={'source_url': validated_data["link"]},
-        )
+        video, _ = Video.objects.get_or_create(defaults={'source_url': validated_data["link"]},)
```
giocalitri: can you explain this change?

ui/models.py
mbertrand: I think by ensuring that bucket_name and object_key database fields are never null, there shouldn't be any exception here to handle, only later when this is used to access the object in some way, for example ```videofile.s3_object.delete()```

cloudsync/tasks.py
mbertrand: When I moved ```from urllib.parse import unquote``` after ```import boto3```, pylint complained that urllib had to be imported 1st. So I changed it to ```import urllib``` in order to keep all the ```from ``` statements after the ```import``` statements.

ui/conftest.py
mbertrand: In addition to the bucket for original uploads, there will be a bucket for transcoded videos, and another bucket for thumbnails.

ui/serializers.py
mbertrand: The Video object no longer has an 's3_object_key', that has been moved to VideoFile model.  But this create method isn't used anywhere, maybe I should remove it altogether.

mbertrand: The Video model no longer has an 's3_object_key' attribute, that was moved to VideoFile.  But this ```create``` method isn't used anywhere, maybe I should remove it altogether.

ui/models.py
mbertrand: Makes sense, I will change it to not be a property

ui/models.py
mbertrand: Something like this?

```
class EncodingNames:
    ORIGINAL="original"
    SMALL="small"
    MEDIUM="medium"
    LARGE="large"
    HLS="hls"
```

```
ET_PRESET_IDS={
    EncodingNames.SMALL: '1351620000001-000061',
    EncodingNames.MEDIUM:'1351620000001-000040',
    EncodingNames.LARGE:'1351620000001-000020'
}
```

giocalitri: yes

ui/serializers.py
giocalitri: I still do not understand what we are trying to do here: what's the case when the object already exists?

ui/conftest.py
giocalitri: ok

ui/serializers.py
mbertrand: I think this  'create' method was intended to allow for creating a video via the REST API, but it's not actually used in the app anywhere.

ui/views.py
mbertrand: Done

ui/models.py
mbertrand: Done

cloudsync/tasks.py
giocalitri: yes, but the reason was that urllib is in the standard library and both is installed...

the sequence of imports should be:
```
import <standard library package>
from <standard library package> import <function/class/constants>

import <installed library package>
from < installed library package> import <function/class/constants>

import <local app package>
from < local app package> import <function/class/constants>
```

you can switch the `from` with the imports (other people in the group refer one or the other) as long as it is consistent in the module.

ui/models.py
giocalitri: I think there can still be a problem with the connectivity, but We can postpone handling that.

cloudsync/tasks.py
mbertrand: Thanks, I misunderstood the earlier comments to mean putting all 'import' statements before 'from' statements regardless of package source, I'll update this.

giocalitri:
👍

Issue #19, opened 2017-06-22
transcoding video files


pdpinch:
As a user, I want to upload video files in any format and have them converted to formats suitable for streaming. 

### acceptance criteria

- [x] #3 asynchronous process, since it can take some time
- [x] #3 user needs to see status in app (done, in progress, failure)
- [x] #23 retry on failures
- [x] #24 devops needs to be alerted on repeated failures
- [x] #3 need to choose output formats based on front-end requirements
- [x] #3 video output formats should be configurable at the application level
- [x] #77 choose transcode presets

### questions

- [x] Is HLS sufficient, or do we still need multiple, individual bit rates?
  - HLS on AWS Cloudfront requires multiple individual files

### nice to have

- [x] notify user (email) when transcode is complete

mbertrand:
> need to choose output formats based on front-end requirements
> video output formats should be configurable

My assumption was that all videos would be transcoded the same way using the same presets, and at most the user could choose to play back the video in the player at different resolutions like 360P vs HD, but the default would be for the browser to choose the best resolution based on the client's bandwidth.

>Is HLS sufficient, or do we still need multiple, individual bit rates?

Based on the example I walked through, multiple bit rates are still generated, but HLS has a 'master playlist' file that links to the different encodings to allow for adaptive streaming: https://www.jwplayer.com/blog/encoding-hls-with-amazon-elastic-transcoder/

pdpinch:
Sorry, I wasn't clear about "configurable." I want the same transcode settings for the entire app, but I'd like a way to be able to change those settings if we need to at some point in the future. I updated the description to specify "configurable at the application level".

I removed the question about HLS. It appears that cloudfront only supports the HLS configuration where a manifest file points to multiple files in different bitrates.

rhysyngsun:
Reopening because #23 and #24 are still open.

pdpinch:
This seems done to me

emmello:
Do you want to limit what formats can be uploaded?  This was a question with TechTV because of issues with filesizes - m4vs, ProRes formats, etc.

pdpinch:
I think we're taking a pragmatic approach: if AWS can transcode it, we'll take it. 

If it can't transcode it, we'll return an error to the user.

emmello:
Sounds good.  Some of the issues may have been with the custom transcode for TechTV, so probably doesn't apply with Elastic Transcode.

Issue #3, opened 2017-06-13
Transcode videos after uploading


mbertrand:
- Use AWS Elastic Transcoder.
- Transcode to multiple bitrates/resolutions? Robert mentioned transcoding at 4 bit-rates for current TechTV

Ferdi:
@mbertrand can we copy the settings from the current techtv elastic transcoder ?

mbertrand:
@Ferdi  Reviewing my notes, looks like Robert told me the multiple bit-rates were for lecture capture (1.0,1.8,2,2.8 mb/sec), I will ask if he has the full transcoding configuration available for TechTV.  When I tried out the current TechTV site I didn't see links to alternate video resolutions/bit-rates after the transcoding completed for my video, so there may be just one.

mbertrand:
Robert: 

> "TechTV processes the original file into 5 different “flavors.”:
> Small.mp4
> Medium.mp4
> Basic.mp4
> Large.mp4
> HD.mp4
>  
> Small.mp4 serves clients with approx. 100kbps connectivity (low-end ISDN). The other extreme, HD.mp4, is approx. 2500kbps and is suitable for most US-based broadband access. The three in-between encodes are available as detected bandwidth allows (our heavily-customised FlowPlayer uses their standard bandwidth detection plug-in.)"

mbertrand:
@pdpinch Robert mentioned that there is a private "StarCluster/Transcluster" repo that has the AWS Elastic Transcode profiles for each of the above video sizes, could you give me access to those?

pdpinch:
I'm hoping he means: https://github.com/TechTV/transcluster

Let me know if that doesn't answer your questions.

mbertrand:
Thanks, that's it: https://github.com/TechTV/transcluster/blob/develop/configs/et/transcluster-presets.json

PR #25, opened 2017-06-29
Transcoding Videos

mbertrand:
closes https://github.com/mitodl/odl-video-service/issues/19
closes https://github.com/mitodl/odl-video-service/issues/3

[The django-elastictranscoder PR](https://github.com/mitodl/django-elastic-transcoder/pull/1) should probably be merged first.

#### What's this PR do?
- Based on preset ID's in settings, an AWS ElasticTranscode job will be created and run immediately after a video is uploaded (and runs asynchronously), generating an HLS playlist and thumbnails.  The current default presets are 2M, 1M, and 600K HLS video, but we should probably create our own higher-resolution presets.

- When a video detail page is loaded, the transcode job's status for that video is checked (TODO: make use of SNS notifications to trigger this).  If the job is complete, a VideoFile model object is created for the playlist, and VideoThumbnail objects are created for each thumbnail (1 or more depending on length of video).

- When a video is deleted, all the associated VideoFile and VideoThumbnail objects along with associated S3 objects will be deleted.

You will need AWS credentials, Dropbox application credentials, etc as described in the README to run this application.  Once running, upload a video, then click 'view your uploaded videos', and click the link to your uploaded video.  This will load the video detail page.  At first you will probably just see links to the 'Original' video (the one uploaded).  If you refresh the page after a few minutes, you should see links for an 'HLS' video and one or more thumbnails.


ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
...
+        Args:
+            encoding(str): The encoding preset to use (should be one included in settings.ET_PRESET_IDS)
+
+        Returns:
+            str: The S3 key to used for the encoded file.
+        """
+        original_s3_key = self.videofile_set.get(encoding='original').s3_object_key
+        if not preset:
+            return original_s3_key
+        output_template = 'transcoded/{}_{}'
```
rhysyngsun: Could you move this to a module-level constant?

ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
...
+        for playlist in job['Playlists']:
+            VideoFile.objects.get_or_create(
+                bucket_name=settings.VIDEO_S3_TRANSCODE_BUCKET,
+                s3_object_key='{}.m3u8'.format(playlist['Name']),  # This assumes HLS encoding
+                encoding=EncodingNames.HLS,
+                preset_id=','.join([output['PresetId'] for output in job['Outputs']]),
+                video=self
+            )
+        for output in job['Outputs']:
+            if 'ThumbnailPattern' in output:
```
rhysyngsun: Take it or leave it: `if 'ThumbnailPattern' not in output: continue` to avoid an indent.

ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
...
+                thumbnail_pattern = output['ThumbnailPattern'].replace("{count}", "")
+                preset = get_et_preset(output['PresetId'])
+                s3 = boto3.resource("s3")
```
rhysyngsun: Can move `s3` and `bucket` outside the loop`, maybe even to their own property for easier mocking.

ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
...
+
     def __str__(self):
         return self.title or "<untitled video>"
 
     def __repr__(self):
         return '<Video: {self.title!r} {self.s3_subkey!r}>'.format(self=self)
 
 
-class VideoFile(models.Model):
+class VideoS3Mixin(object):
```
rhysyngsun: How about keeping this a subclass of `models.model`, marking it an abstract model, and retaining these fields: `created_at`, `s3_object_key`, `s3_object_key` and the method `delete_s3_objects` as implemented in `VideoThumbnail`.

ui/util.py
```diff
@@ -56,8 +60,16 @@ def get_moira_client():
 # http://boto3.readthedocs.io/en/stable/reference/services/cloudfront.html#generate-a-signed-url-for-amazon-cloudfront
 
 def rsa_signer(message):
-    """rsa signer"""
-    private_key_file_path = "/run/secrets/cloudfront-key"
+    """
+    Create an RSA signature for use in a signed URL
+
+    Args:
+        message():
```
rhysyngsun: Could you finish the docstring?

ui/util.py
```diff
@@ -107,3 +119,31 @@ def get_expiration(query_params, default_duration=timedelta(hours=2)):
...
+    """
+    Get the status of an ElasticTranscode job
+
+    Args:
+        job_id(str): ID of ElasticTranscode Job
+
+    Returns:
+        JSON representation of job status/details
+    """
+    et = boto3.client('elastictranscoder', settings.AWS_REGION)
```
rhysyngsun: Should probably refactor `boto3.client('elastictranscoder', settings.AWS_REGION)` to a method since it's used in a few places.

cloudsync/tasks.py
```diff
@@ -53,6 +56,45 @@ def callback(bytes_uploaded):
     )
 
 
+@shared_task(bind=True)
+def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
+    """
+    Given an S3 object key, transcode that object using a video pipeline, overwrite the original when done.
+    :param self: The celery task
```
rhysyngsun: Could you switch this docstring to our standard style?

cloudsync/tasks.py
```diff
@@ -53,6 +56,45 @@ def callback(bytes_uploaded):
...
+        'SegmentDuration': '10.0'
+    } for preset in settings.ET_PRESET_IDS]
+
+    playlists = [{'Format': 'HLSv4',
+                  'Name': video.transcode_key('_index'),
+                  'OutputKeys': [output['Key'] for output in outputs]}]
+
+    # Generate thumbnails for the 1st encoding (no point in doing so for each).
+    basename, _ = os.path.splitext(video_file.s3_object_key)
+    outputs[0]['ThumbnailPattern'] = "thumbnails/{}_thumbnail_{{count}}".format(basename)
```
rhysyngsun: Should probably validate the length of `settings.ET_PRESET_IDS` and raise an appropriate error at the start of the method since you're depending on it to be nonzero here.

cloudsync/tasks.py
```diff
@@ -53,6 +56,45 @@ def callback(bytes_uploaded):
...
+    transcoder = Transcoder(
+        settings.ET_PIPELINE_ID,
```
rhysyngsun: What happens if any of these aren't set?

cloudsync/tasks.py
```diff
@@ -53,6 +56,45 @@ def callback(bytes_uploaded):
...
+        'PresetId': preset,
+        'SegmentDuration': '10.0'
+    } for preset in settings.ET_PRESET_IDS]
+
+    playlists = [{'Format': 'HLSv4',
+                  'Name': video.transcode_key('_index'),
+                  'OutputKeys': [output['Key'] for output in outputs]}]
+
+    # Generate thumbnails for the 1st encoding (no point in doing so for each).
+    basename, _ = os.path.splitext(video_file.s3_object_key)
```
rhysyngsun: This might be useful as a class property on the mixin/base class.

cloudsync/tasks.py
```diff
@@ -53,6 +56,45 @@ def callback(bytes_uploaded):
...
+    }
+
+    # Generate an output video file for each encoding (assumed to be HLS)
+    outputs = [{
+        'Key': video.transcode_key(preset),
+        'PresetId': preset,
+        'SegmentDuration': '10.0'
+    } for preset in settings.ET_PRESET_IDS]
+
+    playlists = [{'Format': 'HLSv4',
```
rhysyngsun: Could you format this comprehension like the one above for consistency?

ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
...
+            str: The S3 key to used for the encoded file.
+        """
+        original_s3_key = self.videofile_set.get(encoding='original').s3_object_key
+        if not preset:
+            return original_s3_key
+        output_template = 'transcoded/{}_{}'
+        basename, _ = os.path.splitext(original_s3_key)
+        return output_template.format(basename, preset)
+
+    def process_transcode_results(self, job):
```
rhysyngsun: I don't think a model is an appropriate place for this method. Usually we'd put these in a method like `api.py`.

ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
             user=self.creator.id, uuid=str(self.s3_subkey), base=basename, ext=extension)
         return newkey
 
+    @property
```
rhysyngsun: I don't think this should be a `property` if it's having a side effect, particularly because it is getting invoked in the template. Also, it'd probably be best to separate the data pull from the rest. I'd probably refactor the refresh into an api method and call the refresh in the view code.

ui/models.py
```diff
@@ -66,24 +69,85 @@ def s3_key(self):
...
+        """
+        Create VideoFile and VideoThumbnail objects for a Video based on AWS ET job output
+
+        Args:
+            job(JSON): JSON representation of AWS ET job output
+        """
+
+        for playlist in job['Playlists']:
+            VideoFile.objects.get_or_create(
+                bucket_name=settings.VIDEO_S3_TRANSCODE_BUCKET,
```
rhysyngsun: This is a lot of args to pass into `get_or_create`, usually you'd just pass those that uniquely identify the object, and pass the rest as a `dict` to the `defaults` arg. Ideally those unique args should have a unique constraint on them at the DB level as well, unless there's a good reason not to. Same for the `VideoThumbnail` code below.

cloudsync/conftest.py
```diff
@@ -67,3 +68,33 @@ def stub_aws_upload():
...
+
+
+class MockClientET:
+    """
+    Mock boto3 ElasticTranscoder client, because ElasticTranscode isn't supported by moto yet
+    """
+    job = None
+    preset = None
+
+    def __init(self, *args, **kwargs):  # pylint: disable=unused-argument
```
rhysyngsun: Typo here and in docstring: `__init__`

ui/signals.py
```diff
@@ -0,0 +1,16 @@
...
+from ui.models import VideoFile, VideoThumbnail
+
+
+@receiver(pre_delete, sender=VideoFile)
+@receiver(pre_delete, sender=VideoThumbnail)
+def delete_s3_files(sender, **kwargs):  # pylint:disable=unused-argument
+    """
+    Make sure S3 files are deleted along with associated video file/thumbnail object
+    """
+    kwargs['instance'].delete_s3_objects()
```
rhysyngsun: Any concern here about the length of this operation causing the request to time out and the object not end up completely cleaned up / deleted? If so, I'd maybe defer cleanup of the S3 assets to a celery task.

requirements.txt
```diff
@@ -53,6 +56,7 @@ requests==2.18.1
 s3transfer==0.1.10        # via boto3
 simplegeneric==0.8.1      # via ipython
 six==1.10.0               # via cryptography, packaging, prompt-toolkit, python-dateutil, traitlets, zeep
+south==1.0.2
```
mbertrand: I think it was added here as a result of django-elastic-transcoder being a dependency, from which south was just removed last week. I'll get rid of it here.

scripts/create_presets.py
```diff
@@ -0,0 +1,41 @@
...
+    with open(os.path.join(script_path, '../config/et_presets.json')) as injson:
+        presets = json.load(injson)
+
+    client = boto3.client('elastictranscoder',
+                          region_name=region, aws_access_key_id=aws_key, aws_secret_access_key=aws_secret)
+    for preset in presets:
+        created = client.create_preset(**preset)
+        print("{}:{}".format(created['Preset']['Id'], created['Preset']['Name']))
+
+if __name__ == "__main__":
```
mbertrand: Done

ui/models.py
mbertrand: Move to ui.api.refresh_status, called from ui.views.VideoDetail

ui/models.py
mbertrand: Done

ui/models.py
mbertrand: Moved to ui.api.process_transcode_results

ui/models.py
mbertrand: Done

cloudsync/tasks.py
mbertrand: AWS will raise an error, and transcoding task will fail.

ui/util.py
```diff
@@ -64,7 +64,7 @@ def rsa_signer(message):
     Create an RSA signature for use in a signed URL
 
     Args:
-        message():
+        message(): The message to be signed
```
rhysyngsun: Is there a type for the message?

ui/views.py
```diff
@@ -63,7 +64,10 @@ class VideoDetail(DetailView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context['form'] = VideoForm(instance=context["object"])
+        video = context["object"]
+        video.status = refresh_status(video)
```
rhysyngsun: This line basically amounts to `video.status = video.status`, so i think you can remove it.

ui/models.py
```diff
@@ -201,30 +170,31 @@ def cloudfront_signed_url(self):
...
-            s3 = boto3.resource("s3")
-            bucket = s3.Bucket(settings.VIDEO_S3_TRANSCODE_BUCKET)
-            prefix = self.s3_object_key[:self.s3_object_key.rfind('/')]
-            for obj in bucket.objects.filter(Prefix=prefix):
-                obj.delete()
+            key = self.s3_object_key[:self.s3_object_key.rfind('/')]
+            delete_s3_objects.delay(self.bucket_name, key, as_filter=True)
         else:
-            self.s3_object.delete()
+            delete_s3_objects.delay(self.bucket_name, self.s3_object_key)
```
rhysyngsun: Could be just `super().delete_from_s3()`

ui/api.py
```diff
@@ -0,0 +1,67 @@
...
+
+def refresh_status(video):
+    """
+    Check the encode job status & if not complete, update the status via a query to AWS.
+
+    Returns:
+        str: The video's latest encode job status
+
+    """
+    if video.status in ['Complete', 'Error']:
```
rhysyngsun: i think you can just remove all the return code and just execute lines 62-66 if the status isn't complete or error. See my other comment in the view.

ui/models.py
```diff
@@ -201,30 +170,31 @@ def cloudfront_signed_url(self):
...
         """
         HLS encoding creates multiple S3 objects, use this method to delete them all.
         """
         if self.encoding == EncodingNames.HLS:
-            s3 = boto3.resource("s3")
-            bucket = s3.Bucket(settings.VIDEO_S3_TRANSCODE_BUCKET)
-            prefix = self.s3_object_key[:self.s3_object_key.rfind('/')]
-            for obj in bucket.objects.filter(Prefix=prefix):
-                obj.delete()
+            key = self.s3_object_key[:self.s3_object_key.rfind('/')]
```
rhysyngsun: `os.path.dirname(self.s3_object_key)` should give the same result.

ui/api.py
```diff
@@ -0,0 +1,67 @@
...
+                    'preset_id': output['PresetId'],
+                    'max_height': int(preset['Thumbnails']['MaxHeight']),
+                    'max_width': int(preset['Thumbnails']['MaxWidth'])
+                }
+            )
+
+
+def refresh_status(video):
+    """
+    Check the encode job status & if not complete, update the status via a query to AWS.
```
mbertrand: Done

rhysyngsun:
Code looks good, moving on to manual testing.

Issue #18, opened 2017-06-22
Upload video files


pdpinch:
As a user I need to upload files from dropbox to the video service. 

### acceptance criteria:

- [x] upload needs to be asynchronous and not dependent on the current browser session, because they can take a long time
- [x] uploading must be robust
  - [x] we should retry if it fails
  - [x] if it fails repeatedly, the user needs to know and devops needs to know 
  - [x] if it fails repeatedly, devops needs to know
  - [x] the file on S3 should be the exact copy of the file on dropbox
- [x] avoid collisions: the users should be allowed to upload the same video multiple times having multiple instances of the same video both on Amazon and in the Portal

### Questions:

- [x] Can we ensure that the files are actually video files? (how hard do we need to try?)
- [x] What happens / what do we do if files are too large for dropbox?

mbertrand:
> Can we ensure that the files are actually video files? (how hard do we need to try?)

I had some luck using the python-magic module to do this yesterday, it could tell the difference between a real mp4 file and a text file that I renamed with an .mp4 extension.  But I doubt it's 100% foolproof. Maybe we could do some basic validation, but if a bad/corrupt file still gets through that the transcoding will fail, and we could notify the user with a message similar to that used by Youtube: "The video has failed to process. Please make sure you are uploading a supported file type."

>What happens / what do we do if files are too large for dropbox?

20 GB is the current dropbox limit.  Youtube handles direct browser uploads of up to 128 GB (if finished < 12 hours), so maybe we could try to emulate that, or have them send us the video via FTP/SD-card and then we use a script to process it.

Ferdi:
I think what I was talking about is basically this part of the code https://goo.gl/aKPn9E

Can we do a review about what that code and see what best practices are for uploading / streaming to S3. For example, we we are concerned about large files can we do something like this https://goo.gl/aMRmL2 and/or use parallel upload in some way ?

mbertrand:
Sure, a code review sounds good.

Once a video file reaches the web server, it will be uploaded to S3 with the [Bucket.file_upload_obj](http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Bucket.upload_fileobj) method which uses multiple threads "if necessary" depending on the transfer configuration.  

But first the file needs to get to the web server.  If it is too big to fit on Dropbox, there are Javascript libraries that could be used to upload them direct through the browser in parallel chunks, for example this one: https://github.com/23/resumable.js

[...]

PR #28, opened 2017-07-06
Added explicit configuration for the upload to S3

giocalitri:
N/A but probably of #18

#### What's this PR do?
Exposes configuration for upload to S3 so we can tune it in the future

The upload to S3 should be slightly faster using the local environment, but functionally it should simply work as before.

### DIFF
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -9,10 +9,48 @@ https://docs.djangoproject.com/en/1.11/topics/settings/
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/1.11/ref/settings/
 """
-
+import ast
 import os
 from urllib.parse import urljoin
+
 import dj_database_url
+import yaml
+
+
+VERSION = "0.0.0"
+
+CONFIG_PATHS = [
+    os.environ.get('ODL_VIDEO_CONFIG', ''),
+    os.path.join(os.getcwd(), 'odl_video.yml'),
+    os.path.join(os.path.expanduser('~'), 'odl_video.yml'),
+    '/etc/odl_video.yml',
+]
+
+
+def load_fallback():
+    """Load optional yaml config"""
+    fallback_config = {}
+    config_file_path = None
+    for config_path in CONFIG_PATHS:
+        if os.path.isfile(config_path):
+            config_file_path = config_path
+            break
+    if config_file_path is not None:
+        with open(config_file_path) as config_file:
+            fallback_config = yaml.safe_load(config_file)
+    return fallback_config
+
+FALLBACK_CONFIG = load_fallback()
+
+
+def get_var(name, default):
+    """Return the settings in a precedence way with default"""
+    try:
+        value = os.environ.get(name, FALLBACK_CONFIG.get(name, default))
+        return ast.literal_eval(value)
+    except (SyntaxError, ValueError):
+        return value
+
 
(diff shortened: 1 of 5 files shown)


noisecapella:
Can you rebase please?

giocalitri:
Done

odl_video/settings.py
```diff
@@ -182,3 +182,27 @@
...
 VIDEO_S3_TRANSCODE_BUCKET = os.environ.get('VIDEO_S3_TRANSCODE_BUCKET', '{}_transcoded'.format(VIDEO_S3_BUCKET))
 VIDEO_S3_THUMBNAIL_BUCKET = os.environ.get('VIDEO_S3_THUMBNAIL_BUCKET', '{}_thumbnails'.format(VIDEO_S3_BUCKET))
+
+# AWS S3 upload settings
+# the defaults values come from the default configuration in boto3.s3.transfer.TransferConfig
+# apart from the first 2
+KB = 1024
+MB = KB * KB
+
+AWS_S3_UPLOAD_MULTIPART_THRESHOLD_MB = os.environ.get('AWS_S3_UPLOAD_MULTIPART_THRESHOLD_MB', 32)
```
noisecapella: These environment variables will be interpreted as strings. Can you parse these into ints here and below, and into a bool at the end? Maybe we should use `get_var` from the cookiecutter here

odl_video/settings.py
```diff
@@ -182,3 +182,27 @@
...
+AWS_S3_UPLOAD_MULTIPART_CHUNKSIZE_MB = os.environ.get('AWS_S3_UPLOAD_MULTIPART_CHUNKSIZE_MB', 32)
+AWS_S3_UPLOAD_MAX_CONCURRENCY = os.environ.get('AWS_S3_UPLOAD_MAX_CONCURRENCY', 10)
+AWS_S3_UPLOAD_NUM_DOWNLOAD_ATTEMPTS = os.environ.get('AWS_S3_UPLOAD_MULTIPART_CHUNKSIZE_MB', 5)
```
noisecapella: The variable name and environment variable name don't match

odl_video/settings.py
```diff
@@ -182,3 +182,27 @@
 VIDEO_S3_BUCKET = os.environ.get('VIDEO_S3_BUCKET', 'odl-video-service')
 VIDEO_S3_TRANSCODE_BUCKET = os.environ.get('VIDEO_S3_TRANSCODE_BUCKET', '{}_transcoded'.format(VIDEO_S3_BUCKET))
 VIDEO_S3_THUMBNAIL_BUCKET = os.environ.get('VIDEO_S3_THUMBNAIL_BUCKET', '{}_thumbnails'.format(VIDEO_S3_BUCKET))
+
+# AWS S3 upload settings
+# the defaults values come from the default configuration in boto3.s3.transfer.TransferConfig
+# apart from the first 2
```
noisecapella: Why are the first two different?

giocalitri: I made some tests and the default value of 8MB seems to not be optimal for the upload.

odl_video/settings.py
giocalitri: sure

odl_video/settings.py
giocalitri: copy/paste error

giocalitri:
changes done

Issue #36, opened 2017-07-12
celery requires the ODL_VIDEO_DB_DISABLE_SSL setting for docker-compose


mbertrand:
If ODL_VIDEO_DB_DISABLE_SSL is not set, celery will throw an error when trying to access the database.  The docker-compose.yml file needs that setting added to the celery section.

PR #37, opened 2017-07-12
Add 'ODL_VIDEO_DB_DISABLE_SSL' to celery in docker-compose.yml

mbertrand:
Fixes https://github.com/mitodl/odl-video-service/issues/36

#### What's this PR do?
Adds 'ODL_VIDEO_DB_DISABLE_SSL' to the celery configuration in docker-compose.yml

- Make sure 'ODL_VIDEO_DB_DISABLE_SSL' is not in your .env file
- ```docker-compose build```
- ```docker-compose run celery python manage.py shell```
- ```import os; os.environ.get('ODL_VIDEO_DB_DISABLE_SSL')```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,6 +71,9 @@ services:
       - redis
     env_file: .env
     environment:
+      DEV_ENV: 'True'
+      ODL_VIDEO_SECURE_SSL_REDIRECT: 'False'
+      ODL_VIDEO_DB_DISABLE_SSL: 'True'
       DATABASE_URL: postgres://postgres@db:5432/postgres
       REDIS_URL: redis://redis:6379/0
 


noisecapella:
Instead of doing this fix can we change the overall `docker-compose.yml` layout to match the cookiecutter? There there is a `python` container with all environment variables set, and the web and celery containers extend it

mbertrand:
I can change the overall yaml file, but some sections will need to be rewritten since it currently uses docker-composer version 3 and makes use of 'secrets' (not available in version 2.x), while the cookiecutter version uses 'extends' (not available in version 3.x)

rhysyngsun:
^ Giving up assignment of this PR until that's sorted out.

noisecapella:
@rhysyngsun @mbertrand and I talked and we decided to keep the version `3` file since it supports `secrets`. We'll figure out how to fix the docker-compose files to match each other later, but for this PR it will just keep the environment variables the same for `web` and `celery`

mbertrand:
@noisecapella @rhysyngsun I just made another commit which copies over a couple more environment settings from web to celery that seemed relevant.

Issue #38, opened 2017-07-13
HLS video URL's in playlists can be wrong w/non-alphanumeric characters


mbertrand:
If an S3 key contains certain non-alphanumeric characters, the URL for the associated video file that is saved to an HLS playlist file may be incorrect, and consequently the playlist will load in the browser but the video will not play.  Best solution is probably to avoid using the original upload filename altogether in S3 keys, and just use the video's unique UUID in the key.

PR #39, opened 2017-07-13
Stop using uploaded filenames for S3 keys, just use UUID instead.

mbertrand:
Fixes https://github.com/mitodl/odl-video-service/issues/38

#### What's this PR do?
Creates an S3 key for a video file as {user.id}/{uuid}/{uuid}.{extension} instead of  {user.id}/{uuid}/{base-filename}.{extension}

Upload a video and then check the S3 key link.  It should contain a UUID (2x) but not the filename.

### DIFF
--- a/ui/models.py
+++ b/ui/models.py
@@ -67,9 +67,9 @@ class Video(models.Model):
         Returns:
             str: A unique S3 key including the user id as a virtual subfolder
         """
-        basename, extension = os.path.splitext(self.source_url.split('/')[-1])
-        newkey = '{user}/{uuid}/{base}{ext}'.format(
-            user=self.creator.id, uuid=str(self.s3_subkey), base=basename, ext=extension)
+        _, extension = os.path.splitext(self.source_url.split('/')[-1])
+        newkey = '{user}/{uuid}/video{ext}'.format(
+            user=self.creator.id, uuid=str(self.s3_subkey), ext=extension)
         return newkey
 
     def transcode_key(self, preset=None):
--- a/ui/models_test.py
+++ b/ui/models_test.py
@@ -43,7 +43,7 @@ def test_video_model_s3keys(user):
     assert isinstance(new_video.s3_subkey, uuid.UUID)
     s3key = new_video.s3_key()
     assert s3key is not None
-    assert s3key == '{}/{}/fake.mp4'.format(user.id, new_video.s3_subkey)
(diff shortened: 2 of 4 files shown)


giocalitri:
A question: why repeating the uuid twice? cannot you just use a standard string for the filename?
Or can there be multiple ones?

giocalitri:
Meaning is the second uuid different from the first one?

mbertrand:
@giocalitri The uuid would be the same. I should be able to change to a standard string like 'file', or the video.id, whatever your preference is.  For example ```1/{uuid}/file{extension}``` or ```1/{uuid}/{video.id}{extension}```

giocalitri:
I do not have any preference, but I was wondering if too much anonymizatio could hurt us in the long term.
For example knowing that inside the directory identified by the UUID the file name is readable can help to understand what it is.
This is just a thought, honestly

mbertrand:
@giocalitri The extension will still be included so that should be helpful in knowing what the file is (.m3u8: hls playlist, .ts: hls video file, .jpg: thumbnail, etc.)  

I agree having the UUID in there twice seems unnecessary.  How about just 'video'? So for example, for a transcoded video the s3 keys would be:

```
transcoded/1/bc492aee-e7cf-4cab-89bf-c9e7a92a92d3/video__index.m3u8
transcoded/1/bc492aee-e7cf-4cab-89bf-c9e7a92a92d3/video_1351620000001-200015.ts
transcoded/1/bc492aee-e7cf-4cab-89bf-c9e7a92a92d3/video_1351620000001-200025.ts
transcoded/1/bc492aee-e7cf-4cab-89bf-c9e7a92a92d3/video_1351620000001-200055.ts
thumbnails/1/bc492aee-e7cf-4cab-89bf-c9e7a92a92d3/video_thumbnail_00001.png
```

mbertrand:
@giocalitri mind if I go ahead and replace the 2nd uuid with just 'video' as in the example above?

giocalitri:
go ahead

giocalitri:
Just one question:
If we decide to change the file names again in the future, is the global functionality preserved with the old file names?

mbertrand:
@giocalitri the functionality should be preserved with old filenames, since this change should affect only the creation of new videos/videofiles.  Methods to retrieve/update videos and videofiles should continue to work regardless of how the S3 keys were originally generated.

giocalitri:
👍

Issue #20, opened 2017-06-22
simple video playback


pdpinch:
As a video creator, I need to be able to stream my videos

### requirements

- [x] Choose video player: recommend Video.js or AblePlayer
- [x] Create HTML template that embeds video player
- [x] HTML page requires authentication
  - [ ] link to signed Cloudfront URL for video stream
- [ ] must work in mobile (iOS, Android) 
- [ ] subtitle support 

### nice to have

- [ ] chapter support

mbertrand:
Video Player:
- AblePlayer does not support HLS so it is out.
- VideoJS supports HLS and subtitles, BUT:
   - subtitles must be in vtt format not srt (python can be used to convert)
   - the HLS plugin is a bit flaky (compatibility issues with webpack; also encoding issues described below).
  - Supposedly works in mobile but not yet tested

Authentication:
- It is possible to generate a signed Cloudfront URL to the HLS playlist file. However, the playlist contains links to the different bitrate video files, and those links will not be signed.  Therefore, if the Cloudfront distribution is set up to require signed URL's, the playlist will be loaded but the video will not be able to play.  Flowplayer exhibits the same behavior.  Some other options:
  - Use signed cookies instead of URL's.  This will require registering a CNAME to the cloudfront distribution of <something>.odl.mit.edu so that cookies can be sent to cloudfront from the production app.  Theoretically this will work but hard to know without testing.
  - Use encrypted HLS.  I experimented with this using VideoJS and unfortunately it only plays part of the video and finishes prematurely.  Tried with Flowplayer and it wouldnt play at all. 
  - Store the original playlist(s) in the app, and when the video is requested, create a copy with new signed URL's for each link and send that to the page instead. Question: will an HLS playlist with absolute URL's work (default is relative)?

mbertrand:
- Subtitles are supported by video.js but code for uploading/adding them still needs to be added to the app.  Chapters supported by videojs too.  Example with both [here](http://www.aa.org/js/video-js/examples/elephantsdream/).

- I think signed cookies instead of signed url's are probably the best security solution.

pdpinch:
@mbertrand @giocalitri can we close this now that #40 is merged?

mbertrand:
@pdpinch Although videojs does claim to support mobile playback, I haven't verified that it does work yet, and subtitle support is not yet integrated into the app, but if that's okay then I'm fine with closing it now.

pdpinch:
[...]

PR #40, opened 2017-07-13
Simple video player

mbertrand:
https://github.com/mitodl/odl-video-service/issues/20

#### What's this PR do?
Adds a 'VideoPlayer' React component using video.js to play a video on the video detail page.

Upload a new video file.  Wait a few minutes for it to be transcoded. View the video's detail page.  The video should appear and automatically start playing.  There should be a button to manually choose the resolution.


noisecapella:
Can you rebase please?

mbertrand:
@noisecapella rebased

noisecapella:
It looks like `176p` is showing up three times in the resolution selector menu. Is there a way to configure that based on what resolutions are available? 
screenshot from 2017-07-17 15-22-30

mbertrand:
I think the current default transcoding presets don't upscale small videos.  The content of the playlist file is:
```
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=2194000,RESOLUTION=320x176,CODECS="avc1.4d001f"
jellies_1351620000001-200015_v4.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1036000,RESOLUTION=320x176,CODECS="avc1.4d001e"
jellies_1351620000001-200035_v4.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=344000,RESOLUTION=320x176,CODECS="avc1.42001e"
jellies_1351620000001-200055_v4.m3u8
```
since the max resolution for all 3 is "320x176", all that shows up in the options is 176P.  Higher-resolution videos should have different options.

noisecapella:
It seems like the sound is missing from the video. If I play the video by the cloudfront signed link I can hear it

mbertrand:
Updated with different default presets and HlSv3 output, should be able to hear sound now.

noisecapella:
The sound works for me now and I can confirm the quality selector changes the quality. I'm a little confused by the options for quality but I'm fine leaving that for another PR

ui/views_test.py
```diff
@@ -37,3 +40,30 @@ def test_upload_admin(admin_client, mocker):
     assert response.status_code == 200
     assert response.context_data['dropbox_key'] == 'dbkey'
     assert response.template_name == ['ui/upload.html']
+
+
+def test_video_detail_hls(admin_client, video, videofileHLS, mocker):  # pylint: disable=unused-argument
+    """Test video detail page when HLS videofile is available"""
+    mocker.patch('ui.models.make_cloudfront_signed_url', return_value=videofileHLS.cloudfront_url)
+    video.status = 'Complete'
+    url = '/videos/%d/' % video.id
```
noisecapella: Can you use `'/videos/{}/'.format(video.id)` here and below?

package.json
```diff
@@ -73,6 +74,8 @@
     "sinon": "^2.3.1",
     "style-loader": "^0.18.1",
     "url-loader": "^0.5.8",
+    "video.js": "5.20.1",
+    "videojs5-hlsjs-source-handler": "git+https://github.com/streamroot/videojs5-hlsjs-source-handler.git#1a66611",
```
mbertrand: Git version is the [latest release](https://github.com/streamroot/videojs5-hlsjs-source-handler/releases/tag/v0.1.2).  Unfortunately it is not available via npm.

cloudsync/tasks.py
```diff
@@ -88,7 +88,7 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
     } for preset in settings.ET_PRESET_IDS]
 
     playlists = [{
-        'Format': 'HLSv4',
+        'Format': 'HLSv3',
```
noisecapella: Just curious, why does this fix the sound issue?

odl_video/settings.py
```diff
@@ -461,7 +461,7 @@ def get_all_config_keys():
 # not using get_var for the following because the values in the strings would be interpreted as single instrunctions
```
noisecapella: Can you change the last word to `instructions`?

cloudsync/tasks.py
mbertrand: It is a combination of this and different default presets.  The presets handle both audio and video.  HLSv4 requires sound and video to be in separate presets (so 3 video presets and 1 audio preset), which is fine and I could have just added an additional audio preset to the transcode job, but I think HLSv3 may also be more compatible with older iOS/Android phones.

static/js/components/VideoPlayer.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+export default class VideoPlayer extends React.Component {
+  /*
+    Basic component based on https://github.com/videojs/video.js/blob/master/docs/guides/react.md
+    TODO: Add more configuration options, including optional rendering of <TextTrack> elements for subtitles.
+   */
+  componentDidMount() {
+    // instantiate video.js
+    this.player = videojs(this.videoNode, this.props, function onPlayerReady() { // eslint-disable-line no-undef
```
noisecapella: Since `videojs` is a global here can you add `/* SETTINGS videojs: false */` to the top line of this file? At some point we should configure the linter to check for this

ui/conftest.py
```diff
@@ -41,6 +41,23 @@ def video(user):  # pylint: disable=redefined-outer-name
```
noisecapella: At the top of this file there is an import from `settings` like `from odl_video import settings`. Can you change that to `from django.conf import settings`?

noisecapella: Can you also make a similar change in `models_test.py`?

ui/views_test.py
```diff
@@ -37,3 +40,30 @@ def test_upload_admin(admin_client, mocker):
     assert response.status_code == 200
     assert response.context_data['dropbox_key'] == 'dbkey'
     assert response.template_name == ['ui/upload.html']
+
+
+def test_video_detail_hls(admin_client, video, videofileHLS, mocker):  # pylint: disable=unused-argument
+    """Test video detail page when HLS videofile is available"""
+    mocker.patch('ui.models.make_cloudfront_signed_url', return_value=videofileHLS.cloudfront_url)
+    video.status = 'Complete'
+    url = '/videos/{}/'.format(video.id)
```
noisecapella: Can you use `reverse` from `django.core.urlresolvers` here and below? For example `url = reverse('video-detail', kwargs={'pk': video.id})`

ui/views_test.py
```diff
@@ -37,3 +40,30 @@ def test_upload_admin(admin_client, mocker):
...
+    response = admin_client.get(url)
+    assert 'videofile' in response.context_data
+    assert response.context_data is not None
+    js_settings_json = json.loads(response.context_data['js_settings_json'])
+    assert 'videofile' in js_settings_json
+    assert settings.VIDEO_CLOUDFRONT_DIST in js_settings_json['videofile']
```
noisecapella: I think this should be `==` instead of `in`, since the values should always be exactly the same

ui/views_test.py
```diff
@@ -37,3 +40,30 @@ def test_upload_admin(admin_client, mocker):
...
+
+
+def test_video_detail_hls(admin_client, video, videofileHLS, mocker):  # pylint: disable=unused-argument
+    """Test video detail page when HLS videofile is available"""
+    mocker.patch('ui.models.make_cloudfront_signed_url', return_value=videofileHLS.cloudfront_url)
+    video.status = 'Complete'
+    url = '/videos/{}/'.format(video.id)
+    response = admin_client.get(url)
+    assert 'videofile' in response.context_data
+    assert response.context_data is not None
```
noisecapella: This is already checked due to the previous assertion. Also, I'm not sure the check is necessary since the check below will fail if it's not present.

ui/views_test.py
```diff
@@ -37,3 +40,30 @@ def test_upload_admin(admin_client, mocker):
...
+    js_settings_json = json.loads(response.context_data['js_settings_json'])
+    assert 'videofile' in js_settings_json
```
noisecapella: Maybe we should just assert the entire `js_settings_json` dict here. What do you think?

ui/views.py
```diff
@@ -91,8 +92,14 @@ def get_context_data(self, **kwargs):
         video = context["object"]
         refresh_status(video)
         context['form'] = VideoForm(instance=video)
-        context['videofile'] = context['object'].videofile_set.filter(encoding=EncodingNames.HLS)
-        context["js_settings_json"] = json.dumps(default_js_settings(self.request))
+        try:
+            videofile = video.videofile_set.get(encoding=EncodingNames.HLS)
+        except ObjectDoesNotExist:
```
noisecapella: Can we use `Video.DoesNotExist` here?

ui/views.py
```diff
@@ -91,8 +92,14 @@ def get_context_data(self, **kwargs):
...
+            videofile = None
```
noisecapella: When does the video not exist when the user is on this page?

ui/templates/ui/upload.html
```diff
@@ -51,7 +51,7 @@
       let videoId;
       for (videoId in videos) {
```
noisecapella: Can you change this to `videoId of videos`?

noisecapella: Nevermind, maybe we should wait on refactoring this until later

ui/templates/ui/upload.html
```diff
@@ -51,7 +51,7 @@
...
         let video = videos[videoId];
-        let videoFile = video.title;
+        let videoFile = video.key.split("/")[2];
```
noisecapella: Can you grab the last item in the array instead of `2`?

ui/views.py
mbertrand: When the video transcoding is not yet complete there will be no videofile object with encoding=HLS

ui/templates/ui/upload.html
mbertrand: That should actually be video.title now

ui/templates/ui/upload.html
mbertrand: Doesn't work:

```
videos[Symbol.iterator] is not a function
```

ui/conftest.py
mbertrand: Done

static/js/components/VideoPlayer.js
mbertrand: Done

static/js/entry/video.js
```diff
@@ -0,0 +1,26 @@
+require('react-hot-loader/patch');
+require('videojs5-hlsjs-source-handler');
+import 'expose-loader?videojs!video.js'; // Needs to be available as a global
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import VideoPlayer from '../components/VideoPlayer';
+const videoSource = SETTINGS.videofile; // eslint-disable-line no-undef
```
mbertrand: Done

static/js/components/VideoPlayer.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
mbertrand: Done

ui/views_test.py
mbertrand: Done

ui/templates/ui/upload.html
noisecapella: Right, `in` iterates over keys in the object

PR #41, opened 2017-07-14
Refactored rest apis for UI

### LINKED ISSUE #27: Update views to return JSON instead of templates for use as an API in UI components
In order for us to expose a rich client UI we will be implementing the front-end as a Javascript SPA (Single Page App). To that end we will need to update the existing view functions to return a JSON API response instead of the current rendered HTML templates.

giocalitri:
closes #27 

#### What's this PR do?
Refactors the REST views and changes some URLs

Everything should work as before.

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -5,17 +5,23 @@ from celery import chain
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth import login
+from django.contrib.auth.views import login as login_view
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.forms import AuthenticationForm
-from django.contrib.auth.views import login as login_view
-from django.http import JsonResponse, HttpResponseRedirect
+from django.http import HttpResponseRedirect
+from django.db import transaction
 from django.shortcuts import render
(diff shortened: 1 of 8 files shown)


ui/views.py
```diff
@@ -90,60 +96,77 @@ def get_context_data(self, **kwargs):
...
-    return JsonResponse(response)
+class UploadVideosFromDropbox(APIView):
+    """
+    Class based view for uploading videos from dropbox to S3.
+    """
+    authentication_classes = (
+        authentication.SessionAuthentication,
+    )
+    permission_classes = (
+        permissions.IsAuthenticated,
```
mbertrand: The original upload view required admin permissions but I think just being an authenticated user makes more sense.

giocalitri: Yes, I did this on purpose.

giocalitri:
> Looks good to me! One question, should both codecov tests pass before approving a PR? 

this is up to you: I did not write tests for the views I refactored because there were no tests before. 
We can open another issue to write more tests or I can write them in this PR.
I have no preferences

Issue #24, opened 2017-06-27
notify on repeated transcode failures


pdpinch:
When transcodes fail, we will retry them (#23), but if they keep failing we should notify the devops team. 

I defer to @mitodl/devops how they would like this to work, but Sentry seems a reasonable option.

pdpinch:
From https://github.com/mitodl/odl-video-service/issues/23#issuecomment-315925476

> For users, list the status if the transcoding job on the video detail page. Maybe show details as well on why it failed?

Issue #17, opened 2017-06-21
Add more unit tests for existing functionality


mbertrand:
There should be more unit tests for existing functionality & models.

pdpinch:
This would need to be more specific before we can start working on it.

Can I suggest that we add unit tests as we go, when we add new functionality, by backfilling them to areas that are being touched?

pdpinch:
I'm going to close this because we've added a lot of tests and we're adding more as we add features and refactor code. 

If there are specific tests that you think are still lacking, please open specific issue(s).

PR #50, opened 2017-07-18
 Improved error handling, especially for celery tasks

mbertrand:
Related to https://github.com/mitodl/odl-video-service/issues/24 and https://github.com/mitodl/odl-video-service/issues/58

#### What's this PR do?
- New statuses for tasks to better track progress (Uploading, Transcoding, Upload Failed, Transcode Failed).
- Overrides default Transcoder.create_job class to record more detailed results info in EncodeJob objects
- Uses try-catch blocks to update statuses of videos and tasks when failures occur in celery tasks.
- Fixes progress bar on upload page.

- Upload a video file that's about 50 MB or larger.  Progress bar should begin filling in shortly.  Look at network requests and responses: responses should show amount of bytes transferred. When upload is complete, requests should stop being sent and the last response should show a status of 'Transcoding'.  Go to the video detail page, the video 'Encoding Status' should also be 'Transcoding'.  When transcoding is finished, the status should be shown as 'Complete'
- Choose a file to upload from dropbox, then rename the dropbox file after it appears on the upload list but before the progress bar starts filling in.  Eventually, after 15 minutes or so, the upload should fail after AWS gives up trying. Go to the video detail page and the 'Encoding Status' should be 'Upload failed'
- Rename a non-video dropbox file as "myvideo.mp4" and upload it.  Upload should complete successfully but transcoding should eventually fail, and the 'Encoding status' on the video detail page should be 'Transcoding failed'
- Go to ../admin/dj_elastictranscoder/encodejob/ and you should see EncodeJob objects for all videos whose transcoding tasks passed or failed, along with messages that contain the job results details.

### DIFF
--- a/cloudsync/tasks.py
+++ b/cloudsync/tasks.py
@@ -8,12 +8,16 @@ from urllib.parse import unquote
 import requests
 import boto3
 from boto3.s3.transfer import TransferConfig
-from celery import shared_task
+from botocore.exceptions import ClientError
+from celery import shared_task, states, Task
 from celery.utils.log import get_task_logger
+from dj_elastictranscoder.models import EncodeJob
 from django.conf import settings
-from dj_elastictranscoder.transcoder import Transcoder
 
-from ui.models import VideoFile
+from cloudsync.utils import VideoTranscoder
+from ui.api import refresh_status
+from ui.models import VideoFile, Video
+from ui.constants import VideoStatus
 
 log = get_task_logger(__name__)
 CONTENT_DISPOSITION_RE = re.compile(
@@ -23,46 +27,91 @@ CONTENT_DISPOSITION_RE = re.compile(
 THUMBNAIL_PATTERN = "thumbnails/{}_thumbnail_{{count}}"
 
 
-@shared_task(bind=True)
-def stream_to_s3(self, url, s3_key):
+class VideoTask(Task):
+    """
+    Custom Celery Task class for video uploads and transcodes
+    """
+
+    def get_task_id(self):
+        """
+        Get the task id (depending on whether the task is chained or not)
+
+        Args:
+            request(Task.request): The task request
+
+        Returns:
+            The request id
+        """
+        if self.request.chain:
+            try:
+                return self.request.chain[0]['options']['task_id']
+            except (IndexError, KeyError,) as exc:
+                # Log the error and continue, using self.request.id instead
+                # The worst that will happen is that progress bar won't work.
+                log.error("Could not find task_id in chain: %s", str(exc))
+                return
+        return self.request.id
+
+
+@shared_task(bind=True, base=VideoTask)
+def stream_to_s3(self, video_id):
     """
     Stream the contents of the given URL to Amazon S3
     """
-    if not url:
+
+    if not video_id:
         return False
-    response = requests.get(url, stream=True)
-    response.raise_for_status()
+    try:
+        video = Video.objects.get(id=video_id)
+    except (Video.DoesNotExist, Video.MultipleObjectsReturned):
+        log.error("Exception retrieving video with id %d", video_id)
+        raise
+    video.update_status(VideoStatus.UPLOADING)
+
+    task_id = self.get_task_id()
+    try:
+        response = requests.get(video.source_url, stream=True, timeout=60)
+        response.raise_for_status()
+    except Exception:
+        video.update_status(VideoStatus.UPLOAD_FAILED)
+        self.update_state(task_id=task_id, state=states.FAILURE)
+        raise
+
     _, content_type, content_length = parse_content_metadata(response)
 
     s3 = boto3.resource('s3')
     bucket_name = settings.VIDEO_S3_BUCKET
     bucket = s3.Bucket(bucket_name)
-
-    # Need to bind this here, because otherwise it gets lost in the callback somehow
-    task_id = self.request.id
+    total_bytes_uploaded = 0
 
     def callback(bytes_uploaded):
         """
         Callback function after upload
         """
+        nonlocal total_bytes_uploaded
+        total_bytes_uploaded += bytes_uploaded
         data = {
-            "uploaded": bytes_uploaded,
+            "uploaded": total_bytes_uploaded,
             "total": content_length,
         }
         self.update_state(task_id=task_id, state="PROGRESS", meta=data)
 
     config = TransferConfig(**settings.AWS_S3_UPLOAD_TRANSFER_CONFIG)
+    try:
+        bucket.upload_fileobj(
+            Fileobj=response.raw,
+            Key=video.get_s3_key(),
+            ExtraArgs={"ContentType": content_type},
+            Callback=callback,
+            Config=config
+        )
+    except Exception:
+        video.update_status(VideoStatus.UPLOAD_FAILED)
+        self.update_state(task_id=task_id, state=states.FAILURE)
+        raise
 
-    bucket.upload_fileobj(
-        Fileobj=response.raw,
-        Key=s3_key,
-        ExtraArgs={"ContentType": content_type},
-        Callback=callback,
-        Config=config
-    )
 
-
-@shared_task(bind=True)
+@shared_task(bind=True, base=VideoTask)
 def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
     """
     Given an S3 object key, transcode that object using a video pipeline, overwrite the original when done.
@@ -70,11 +119,11 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
     Args:
         video_id(int): The video primary key
     """
-    if not settings.ET_PRESET_IDS:
-        raise ValueError("At least one transcode preset required in settings")
+    video = Video.objects.get(id=video_id)
+    task_id = self.get_task_id()
+    self.update_state(task_id=task_id, state=VideoStatus.TRANSCODING)
 
     video_file = VideoFile.objects.get(video__id=video_id, encoding='original')
-    video = video_file.video
 
     video_input = {
         'Key': video_file.s3_object_key,
@@ -95,14 +144,44 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
 
     # Generate thumbnails for the 1st encoding (no point in doing so for each).
     outputs[0]['ThumbnailPattern'] = THUMBNAIL_PATTERN.format(video_file.s3_basename)
-    transcoder = Transcoder(
+    transcoder = VideoTranscoder(
         settings.ET_PIPELINE_ID,
         settings.AWS_REGION,
         settings.AWS_ACCESS_KEY_ID,
         settings.AWS_SECRET_ACCESS_KEY
     )
-    transcoder.encode(video_input, outputs, Playlists=playlists)
-    transcoder.create_job_for_object(video)
+    try:
+        transcoder.encode(video_input, outputs, Playlists=playlists)
+    except ClientError as exc:
+        log.error('Transcode job creation failed for video %s', video_id)
+        video.update_status(VideoStatus.TRANSCODE_FAILED)
+        self.update_state(task_id=task_id, state=states.FAILURE)
+        if hasattr(exc, 'response'):
+            transcoder.message = exc.response
+        raise
+    finally:
+        transcoder.create_job_for_object(video)
+        if video.status != VideoStatus.TRANSCODE_FAILED:
+            video.update_status(VideoStatus.TRANSCODING)
+
+
+@shared_task(bind=True)
+def update_video_statuses(self):  # pylint: disable=unused-argument
+    """
+    Check on statuses of all transcoding videos and update their status if appropriate
+    """
+    transcoding_videos = Video.objects.filter(status=VideoStatus.TRANSCODING)
+    for video in transcoding_videos:
+        try:
+            refresh_status(video)
+        except EncodeJob.DoesNotExist:
+            # Log the exception but don't raise it so other videos can be checked.
+            log.exception("No EncodeJob object exists for video id %d", video.id)
+            video.update_status(VideoStatus.TRANSCODE_FAILED)
+        except ClientError as exc:
+            # Log the exception but don't raise it so other videos can be checked.
+            log.exception("AWS error when refreshing job status for video %d: %s", video.id, exc.response)
+            video.update_status(VideoStatus.TRANSCODE_FAILED)
 
 
 def parse_content_metadata(response):
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -7,6 +7,7 @@ import platform
 from urllib.parse import urljoin
 
 import dj_database_url
+from django.core.exceptions import ImproperlyConfigured
 
 from odl_video.envs import (
     get_any,
@@ -163,7 +164,12 @@ CELERY_TASK_SERIALIZER = 'json'
 CELERY_RESULT_SERIALIZER = 'json'
 CELERY_ACCEPT_CONTENT = ['json']
 CELERY_TIMEZONE = 'UTC'
-
+CELERY_BEAT_SCHEDULE = {
+    'update-statuses': {
+        'task': 'cloudsync.tasks.update_video_statuses',
+        'schedule': get_int('VIDEO_STATUS_UPDATE_FREQUENCY', 60)
+    }
+}
 
 # django cache back-ends
 CACHES = {
@@ -440,10 +446,14 @@ ET_PRESET_IDS = get_string(
     '1351620000001-200010,1351620000001-200020,1351620000001-200050'
 ).split(',')
 
+if ET_PRESET_IDS == ['']:  # This may happen if `ET_PRESET_IDS=` is in .env file.
+    raise ImproperlyConfigured('ET_PRESET_IDS cannot be blank, please check your settings & environment')
+
 VIDEO_CLOUDFRONT_DIST = get_string('VIDEO_CLOUDFRONT_DIST', '')
 VIDEO_S3_BUCKET = get_string('VIDEO_S3_BUCKET', 'odl-video-service')
 VIDEO_S3_TRANSCODE_BUCKET = get_string('VIDEO_S3_TRANSCODE_BUCKET', '{}-transcoded'.format(VIDEO_S3_BUCKET))
 VIDEO_S3_THUMBNAIL_BUCKET = get_string('VIDEO_S3_THUMBNAIL_BUCKET', '{}-thumbnails'.format(VIDEO_S3_BUCKET))
+
 
 # server-status
 STATUS_TOKEN = get_string("STATUS_TOKEN", "")
--- a/ui/api.py
+++ b/ui/api.py
@@ -3,10 +3,10 @@ API methods
 """
 import logging
 
-from dj_elastictranscoder.models import EncodeJob
 from django.conf import settings
 
 from ui.encodings import EncodingNames
+from ui.constants import VideoStatus
 from ui.utils import get_et_preset, get_bucket, get_et_job
 from ui.models import VideoFile, VideoThumbnail
 
@@ -53,19 +53,23 @@ def process_transcode_results(video, job):
             )
 
 
-def refresh_status(video):
+def refresh_status(video, encode_job=None):
     """
     Check the encode job status & if not complete, update the status via a query to AWS.
 
     Args:
         video(ui.models.Video): Video object to refresh status of.
+        encode_job(dj_elastictranscoder.models.EncodeJob): EncodeJob associated with Video
     """
-    if video.status not in ['Complete', 'Error']:
-        try:
-            job = get_et_job(video.encode_jobs.latest("created_at").id)
-            if job['Status'] == 'Complete':
-                process_transcode_results(video, job)
-            video.status = job['Status']
-            video.save()
-        except EncodeJob.DoesNotExist:
-            log.error("No encoding job exists for video id %d", video.id)
+    if video.status == VideoStatus.TRANSCODING:
+        if not encode_job:
+            encode_job = video.encode_jobs.latest("created_at")
+        et_job = get_et_job(encode_job.id)
+        if et_job['Status'] == VideoStatus.COMPLETE:
+            process_transcode_results(video, et_job)
+            video.update_status(VideoStatus.COMPLETE)
+        elif et_job['Status'] == VideoStatus.ERROR:
+            video.update_status(VideoStatus.TRANSCODE_FAILED)
+            log.error('Transcoding failed for video %d', video.id)
+        encode_job.message = et_job
+        encode_job.save()
--- a/ui/api_test.py
+++ b/ui/api_test.py
@@ -11,18 +11,45 @@ from moto import mock_s3
 
 from cloudsync.conftest import MockClientET, MockBoto
 from ui.api import process_transcode_results, refresh_status
+from ui.constants import VideoStatus
 
 pytestmark = pytest.mark.django_db
 
 
-def test_video_job_status(mocker, video, encodejob):  # pylint: disable=unused-argument
+def test_video_job_status_error(mocker, video, encodejob):  # pylint: disable=unused-argument
     """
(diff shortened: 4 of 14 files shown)


cloudsync/tasks.py
```diff
@@ -23,58 +26,92 @@
...
+        response = requests.get(video.source_url, stream=True, timeout=60)
+        response.raise_for_status()
+        _, content_type, content_length = parse_content_metadata(response)
+
+        s3 = boto3.resource('s3')
+        bucket_name = settings.VIDEO_S3_BUCKET
+        bucket = s3.Bucket(bucket_name)
+        total_bytes_uploaded = 0
+
+        def callback(bytes_uploaded):
```
giocalitri: Should we move this function definition outside the scope of the other function? It does not make any sense to have it here.

cloudsync/tasks.py
```diff
@@ -23,58 +26,92 @@
...
+            Key=video.s3_key(),
+            ExtraArgs={"ContentType": content_type},
+            Callback=callback,
+            Config=config
+        )
+    except Exception as e:
+        log.exception('Upload failed for video %d', video_id)
+        update_video_status(video, VideoStatus.UPLOAD_FAILED)
+        self.update_state(task_id=task_id, state=states.FAILURE)
+        raise e
```
giocalitri: I do not think that the approach of wrapping the entire code in a try/except is a good one.
I think you need to wrap just the single call you are afraid to fail.
In this way it would be complicated to find where the code actually failed.

cloudsync/tasks.py
```diff
@@ -23,58 +26,92 @@
...
+@shared_task(bind=True, base=VideoTask)
 def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
     """
     Given an S3 object key, transcode that object using a video pipeline, overwrite the original when done.
 
     Args:
         video_id(int): The video primary key
     """
+    video = Video.objects.get(id=video_id)
+    update_video_status(video, VideoStatus.TRANSCODING)
```
giocalitri: cannot `update_video_status` simple be a method of the Video model?

cloudsync/tasks.py
```diff
@@ -95,14 +132,36 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
-    transcoder.encode(video_input, outputs, Playlists=playlists)
-    transcoder.create_job_for_object(video)
+    try:
+        transcoder.encode(video_input, outputs, Playlists=playlists)
+    except ClientError as e:
+        log.exception('Transcode job creation failed for video %s: %s', video_id, e.response)
+        update_video_status(video, VideoStatus.TRANSCODE_FAILED)
+        self.update_state(task_id=task_id, state=states.FAILURE)
+        transcoder.message = e.response
+        raise e
```
giocalitri: if you raise anyway, why do you `'log.exception`?

cloudsync/tasks.py
```diff
@@ -95,14 +132,36 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+    finally:
+        job = transcoder.create_job_for_object(video)
```
giocalitri: what does this do? A retry?

cloudsync/tasks.py
```diff
@@ -95,14 +132,36 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+
+    # Wait for the transcoding job to complete
+    waiter = transcoder.client.get_waiter('job_complete')
+    try:
+        waiter.wait(Id=transcoder.message['Job']['Id'])
```
giocalitri: is this locking on the request?

ui/api.py
```diff
@@ -53,19 +55,42 @@ def process_transcode_results(video, job):
...
+            if not encode_job:
+                encode_job = video.encode_jobs.latest("created_at")
+            et_job = get_et_job(encode_job.id)
+            if et_job['Status'] == VideoStatus.COMPLETE:
+                process_transcode_results(video, et_job)
+                encode_job.message = et_job
+                encode_job.save()
+                update_video_status(video, VideoStatus.COMPLETE)
+            elif et_job['Status'] == VideoStatus.ERROR:
+                update_video_status(video, VideoStatus.TRANSCODE_FAILED)
```
giocalitri: I do not think that wrapping all this code in a try/except is a good idea.
the `EncodeJob.DoesNotExist` seems to be relate to just one call and so does `ClientError`

ui/api.py
```diff
@@ -53,19 +55,42 @@ def process_transcode_results(video, job):
...
+
+
+def update_video_status(video, status):
+    """
+    Assign and save the status of a Video
+    :param video: The video to update
+    :param status: The status to assign the video
+    """
+    video.status = status
+    video.save()
```
giocalitri: yes, this should be a method of the video model

ui/api_test.py
```diff
@@ -19,10 +20,24 @@ def test_video_job_status(mocker, video, encodejob):  # pylint: disable=unused-a
...
+    """
+    Verify that refresh_status does not raise ClientError
+    """
+    video.status = VideoStatus.TRANSCODING
+    mocker.patch('ui.util.boto3', MockBoto)
+    error = Exception("unexpected exception")
+    mocker.patch('ui.util.get_transcoder_client',
+                 return_value=MockClientET(error=error))
+    with pytest.raises(Exception):
+        refresh_status(video)
```
giocalitri: Question: are there already tests for all the use cases of the API functions?

ui/statuses.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class VideoStatus:
+    """Simple class for possible video statuses"""
+    UPLOADING = 'Uploading'
+    UPLOAD_FAILED = 'Upload failed'
+    TRANSCODING = 'Transcoding'
+    TRANSCODE_FAILED = 'Transcode failed'
+    COMPLETE = 'Complete'
+    ERROR = 'Error'
```
giocalitri: What about rename the module from `statuses.py` to `constants.py`

cloudsync/tasks.py
```diff
@@ -23,58 +26,92 @@
...
-
-    bucket.upload_fileobj(
-        Fileobj=response.raw,
-        Key=s3_key,
-        ExtraArgs={"ContentType": content_type},
-        Callback=callback,
-        Config=config
-    )
+        if self.request.chain:
+            return self.request.chain[0]['options']['task_id']
```
mbertrand: self.request.chain[0]['options']['task_id'] is what correctly returns the id of the task during upload and allows the progress bar to accurately reflect the percentage of the file uploaded from dropbox to s3.  Using request.id doesn't work when the upload and transcode tasks are chained.

cloudsync/tasks.py
mbertrand: It is an inner function due to scope issues, it needs access to `video`, `content_length`, and `total_bytes_uploaded`.

cloudsync/tasks.py
mbertrand: ok, I will put try/except just around  `bucket.upload_fileobj` and `response.raise_for_status()`.

cloudsync/tasks.py
mbertrand: This creates an [EncodeJob](https://github.com/mitodl/django-elastic-transcoder/blob/master/dj_elastictranscoder/models.py#L7) object to represent the transcode job and its status/results.

cloudsync/tasks.py
mbertrand: It makes `read_job` requests to AWS every 30 seconds to see if the transcode job has finished.  So it keeps the celery task running until then.  The more I think about it the less I like this approach.  I would prefer to have a separate periodic celery task that would run every X seconds/minutes to check on the job status instead.

cloudsync/tasks.py
mbertrand: sure

cloudsync/tasks.py
mbertrand: I'll get rid of the log statements

ui/api_test.py
mbertrand: No: https://github.com/mitodl/odl-video-service/issues/17

ui/statuses.py
mbertrand: sure

ui/api.py
mbertrand: ClientError could happen in either `get_et_job` or `process_transcode_results` method calls.  I'll make a separate try/catch block for `EncodeJob.DoesNotExist`

cloudsync/tasks.py
mbertrand: Actually the log statements provide info on the video id for which the exception occurred, which may be useful for troubleshooting. Should I get rid of them anyway?

cloudsync/tasks.py
giocalitri: I understand that `request.id` doesn't work when the upload and transcode tasks are chained, but again, the documentation seems to suggest that the current task is not `0` in the chain

cloudsync/tasks.py
```diff
@@ -23,58 +26,95 @@
...
-    response.raise_for_status()
+
+    video = Video.objects.get(id=video_id)
+    video.update_status(VideoStatus.UPLOADING)
+
+    task_id = self.get_task_id()
+    try:
+        response = requests.get(video.source_url, stream=True, timeout=60)
+        response.raise_for_status()
+    except Exception as e:
```
giocalitri: nit: please do not use single letters as variable names

giocalitri: this comment also for the rest of the PR

cloudsync/tasks.py
```diff
@@ -95,14 +135,37 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
         settings.AWS_ACCESS_KEY_ID,
         settings.AWS_SECRET_ACCESS_KEY
     )
-    transcoder.encode(video_input, outputs, Playlists=playlists)
-    transcoder.create_job_for_object(video)
+    try:
+        if not settings.ET_PRESET_IDS:
+            raise ValueError("At least one transcode preset required in settings")
+        transcoder.encode(video_input, outputs, Playlists=playlists)
+    except (ValueError, ClientError) as e:
```
giocalitri: nit: if you are using tuples, please end the "list" with a comma to make clear that it is a tuple
`(ValueError, ClientError,)`

cloudsync/tasks.py
```diff
@@ -95,14 +135,37 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+        log.exception('Transcode job creation failed for video %s: %s', video_id, e.response)
```
giocalitri: doesn't this duplicate the stack trace if later you raise anyway?

cloudsync/tasks.py
giocalitri: and why it is a method of the transcoder?

cloudsync/tasks.py
giocalitri: what you described is locking the celery worker.
we probably should skip the execution and enqueue a new task with a delay: something like this
http://docs.celeryproject.org/en/latest/userguide/calling.html#eta-and-countdown

cloudsync/tasks.py
giocalitri: `"1a859e5a-8f71-4e01-9349-5ebc6dc66631"` is the parent task?

cloudsync/tasks.py
mbertrand: Related to #58 

If the `waiter.wait` method is put in a new delayed task, that task will then be locked until `waiter` finally gets a response, right?  What do you think of a periodic task that checks the status of all videos that have not yet completed or failed?  Or we could make use of AWS SNS with django-elastic-transcoder's endpoint API, though that makes things harder to test locally.

cloudsync/tasks.py
giocalitri: isn't a way to pass arguments to the callback?

cloudsync/tasks.py
mbertrand: I think so, or it effectively seems to function as one anyway.  That is the task id returned by the uploader view response.  There is a `parent_id` attribute in the task.request but it is `None`, and also an `id`, `root_id` and `correlation_id` but none of them are equal to `1a859e5a-8f71-4e01-9349-5ebc6dc66631`

cloudsync/tasks.py
giocalitri: maybe change it to a `log.error` if you want to keep the info

ui/api.py
giocalitri: ok

cloudsync/tasks.py
mbertrand: The only argument that seems to be passed by `Bucket.upload_fileobj` is the number of bytes transferred: http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Bucket.upload_fileobj

cloudsync/tasks.py
giocalitri: ok then
but please wrap the call with a try/except so you can log what went wrong in case the structure changes.
Also you should write tests for this

cloudsync/tasks.py
mbertrand: Right, I'll take out the stack trace from the log message

cloudsync/tasks.py
```diff
@@ -95,14 +135,37 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+        video.update_status(VideoStatus.TRANSCODE_FAILED)
+        self.update_state(task_id=task_id, state=states.FAILURE)
+        transcoder.message = e.response
```
noisecapella: Sorry for the drive by review, but `e.response` won't exist for `ValueError`

cloudsync/tasks.py
```diff
@@ -95,14 +135,37 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+        raise e
```
noisecapella: You can just do `raise` here

cloudsync/tasks.py
mbertrand: Thanks, will fix

cloudsync/tasks.py
mbertrand: @giocalitri what do you think of something like this, running every x minutes:

```
@shared_task(bind=True)
def update_video_statuses(self):
    transcoding_videos = Video.objects.filter(status=VideoStatus.TRANSCODING)
    for video in transcoding_videos:
        try:
            encode_job = video.encode_jobs.latest("created_at")
        except EncodeJob.DoesNotExist as e:
            # This is OK if video is still being uploaded
            if video.status != VideoStatus.UPLOADING:
                log.exception("No EncodeJob object exists for video id %d", video.id)
                raise
        try:
            et_status = get_et_job(encode_job.id)
            encode_job.message = et_status
            encode_job.save()
            if et_status['Status'] == VideoStatus.COMPLETE:
                process_transcode_results(video, et_status)
                video.update_status(VideoStatus.COMPLETE)
            elif et_status['Status'] == VideoStatus.ERROR:
                log.exception('Encoding failed for video %d', video.id)
                video.update(VideoStatus.TRANSCODE_FAILED)
        except ClientError as e:
            log.exception("AWS error when refreshing job status for video %d", video.id)
            raise
```

mbertrand: Maybe this should be moved to a separate PR?

mbertrand: Decided to go with option 3 from #58 based on our discussion this AM

cloudsync/tasks.py
```diff
@@ -23,58 +27,100 @@
...
+        Returns:
+            The request id
+        """
+        if self.request.chain:
+            try:
+                return self.request.chain[0]['options']['task_id']
+            except (IndexError, KeyError,) as exc:
+                # Log the error and continue, using self.request.id instead
+                # The worst that will happen is that progress bar won't work.
+                log.error("Could not find task_id in chain: %s", str(exc))
```
giocalitri: this will make the function return `self.request.id` that I do not think it is what should do

cloudsync/tasks.py
```diff
@@ -23,58 +27,100 @@
...
     Stream the contents of the given URL to Amazon S3
     """
-    if not url:
+
+    if not video_id:
         return False
-    response = requests.get(url, stream=True)
-    response.raise_for_status()
+
+    video = Video.objects.get(id=video_id)
```
giocalitri: this could theoretically raise: if you want to avoid, use try/except or `.filter` and `.first`

cloudsync/tasks.py
mbertrand: Make it return `None` instead?  This is only required for making the progress bar work.

cloudsync/tasks.py
```diff
@@ -23,58 +27,100 @@
...
 
     def callback(bytes_uploaded):
         """
         Callback function after upload
         """
+        nonlocal total_bytes_uploaded
+        total_bytes_uploaded = total_bytes_uploaded + bytes_uploaded
         data = {
-            "uploaded": bytes_uploaded,
+            "uploaded": total_bytes_uploaded + bytes_uploaded,
```
giocalitri: why do you sum again here?

cloudsync/tasks.py
```diff
@@ -23,58 +27,100 @@
...
-        Callback=callback,
-        Config=config
-    )
-
-
-@shared_task(bind=True)
+    try:
+        bucket.upload_fileobj(
+            Fileobj=response.raw,
+            Key=video.s3_key(),
```
giocalitri: we should probably make `Video.s3_key()` a property method

cloudsync/tasks.py
mbertrand: To get the number of bytes transferred since the upload began, so it can be compared to the file size (total), and percentage progress calculated.

cloudsync/tasks.py
```diff
@@ -95,14 +141,42 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
         settings.ET_PIPELINE_ID,
         settings.AWS_REGION,
         settings.AWS_ACCESS_KEY_ID,
         settings.AWS_SECRET_ACCESS_KEY
     )
-    transcoder.encode(video_input, outputs, Playlists=playlists)
-    transcoder.create_job_for_object(video)
+    try:
+        if not settings.ET_PRESET_IDS:
+            raise ValueError("At least one transcode preset required in settings")
```
giocalitri: why this is not in its own try/except?

giocalitri: to reuse the same code in the except?

cloudsync/tasks.py
```diff
@@ -95,14 +141,42 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+    """
+    Check on statuses of all transcoding videos and update their status if appropriate
+    """
+    transcoding_videos = Video.objects.filter(status=VideoStatus.TRANSCODING)
+    for video in transcoding_videos:
+        try:
+            refresh_status(video)
+        except EncodeJob.DoesNotExist:
+            # Log the exception but don't raise it so other videos can be checked.
+            log.exception("No EncodeJob object exists for video id %d", video.id)
```
giocalitri: don't we need to do anything else in this case?

cloudsync/tasks.py
mbertrand: Originally it was, but [based on feedback from an earlier code review](https://github.com/mitodl/odl-video-service/pull/11#discussion_r124569764), it was changed to a function instead.  But I can change it back to a property if you prefer.

cloudsync/tasks.py
giocalitri: not sure how it is used: if None does not crash anything else, go for it

cloudsync/tasks.py
mbertrand: I don't want to raise an exception here because it might block status checking for other videos.  Sentry.io should alert us of the error in the log, right?

cloudsync/tasks.py
giocalitri: you are right, sorry, but maybe a name like `get_s3_key` would be better for a function

cloudsync/tasks.py
mbertrand: Yes, to avoid having much of the same code in the except parts (`video.update()`, `self.update_state()`)

cloudsync/tasks.py
mbertrand: sure, I'll rename it

cloudsync/tasks.py
mbertrand: I could change the video status to 'Failed' here if there is no EncodeJob

cloudsync/tasks.py
mbertrand: I think I'll move this check directly to the `settings.py` file, so it will raise an error immediately on startup if missing.

noisecapella: I'm not sure you can do that, it interferes with `docker-compose build` because it runs a management command `collectstatic` which will import the settings

mbertrand: Since there are defaults for `ET_PRESET_IDS` in settings now, this shouldn't ever happen anyway. So it should probably just be removed.

cloudsync/tasks.py
```diff
@@ -23,58 +27,103 @@
...
+            The request id
+        """
+        if self.request.chain:
+            try:
+                return self.request.chain[0]['options']['task_id']
+            except (IndexError, KeyError,) as exc:
+                # Log the error and continue, using self.request.id instead
+                # The worst that will happen is that progress bar won't work.
+                log.error("Could not find task_id in chain: %s", str(exc))
+                return None
```
giocalitri: nit: just `return` (None is implicit)

cloudsync/tasks.py
giocalitri: ok

cloudsync/util.py
```diff
@@ -0,0 +1,34 @@
+"""Utility classes/methods for cloudsync"""
```
giocalitri: please name the module to `utils`

odl_video/settings.py
```diff
@@ -440,11 +445,15 @@ def get_all_config_keys():
     '1351620000001-200010,1351620000001-200020,1351620000001-200050'
 ).split(',')
 
+if ET_PRESET_IDS == ['']:  # This may happen if `ET_PRESET_IDS=` is in .env file.
+    raise ValueError('ET_PRESET_IDS cannot be blank, please check your settings & environment')
```
giocalitri: probably [ImproperlyConfigured](https://docs.djangoproject.com/en/1.11/ref/exceptions/#improperlyconfigured) is better here

ui/api.py
```diff
@@ -53,19 +53,22 @@ def process_transcode_results(video, job):
...
-            log.error("No encoding job exists for video id %d", video.id)
+    if video.status == VideoStatus.TRANSCODING:
+        if not encode_job:
+            encode_job = video.encode_jobs.latest("created_at")
+        et_job = get_et_job(encode_job.id)
+        if et_job['Status'] == VideoStatus.COMPLETE:
+            process_transcode_results(video, et_job)
+            video.update_status(VideoStatus.COMPLETE)
+        elif et_job['Status'] == VideoStatus.ERROR:
+            video.update_status(VideoStatus.TRANSCODE_FAILED)
```
giocalitri: do we need to log anything here?

ui/models.py
```diff
@@ -88,6 +88,16 @@ def transcode_key(self, preset=None):
...
 
+    def update_status(self, status):
+        """
+        Assign and save the status of a Video
+
+        Args:
+            status(str): The status to assign the video
+        """
+        self.status = status
+        self.save()
```
giocalitri: we should probably open an issue to modify the status field as a list of options

giocalitri:
There are a couple of very minor stuff, but I see in codecov that `cloudsync/task.py` is missing some tests.

ui/models.py
mbertrand: https://github.com/mitodl/odl-video-service/issues/87

ui/api.py
mbertrand: Added a log.error statement

giocalitri:
👍 after  squashing

Issue #56, opened 2017-07-18
The current logout functionality needs to be changed or removed


mbertrand:
The logout link redirects to the admin logged out page, even when touchstone integration is enabled (see http://video.odl.mit.edu).  This behavior needs to change:

- Log out of Touchstone and redirect to Touchstone login page? (not sure if this is possible)
- Remove the logout link altogether?

pdpinch:
IIRC, there's no way to log out of TouchStone. 

If possible, it would be nice to keep a logout button, for logging out of the video application.

PR #60, opened 2017-07-19
Custom logout page

mbertrand:
Fixes https://github.com/mitodl/odl-video-service/issues/56

#### What's this PR do?
- Adds a custom logout HTML template and changes the order of settings.INSTALLED_APPS so that a user is not redirected to the admin logout page after logging out.
- Changes the 'login' link on the base template to redirect to the index page (if touchstone integration is on, they will logged in via that; otherwise there will be a login form on the index page).

Log in, then click the 'log out' link.  You should be redirected to a simple logout page and not the admin logout page. Click the 'Login' link.  You should be redirected to the home page with a login form.

### DIFF
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -104,6 +104,9 @@ WEBPACK_DEV_SERVER_PORT = get_var('WEBPACK_DEV_SERVER_PORT', '8082')
 # Application definition
 
 INSTALLED_APPS = [
+    'ui.apps.UIConfig',
+    'cloudsync.apps.CloudSyncConfig',
+    'dj_elastictranscoder',
     'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
@@ -113,9 +116,6 @@ INSTALLED_APPS = [
     'rest_framework',
     'server_status',
     'raven.contrib.django.raven_compat',
-    'ui.apps.UIConfig',
-    'cloudsync.apps.CloudSyncConfig',
-    'dj_elastictranscoder',
 ]
 
 SESSION_ENGINE = 'django.contrib.sessions.backends.signed_cookies'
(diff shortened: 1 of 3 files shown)


giocalitri:
Why did you choose this approach instead of using the standard django logout feature with simple redirect to the index page via the `next` get parameter (if I recall well)? Basically what we are doing in Micromasters.

ui/templates/registration/logged_out.html
```diff
@@ -0,0 +1,6 @@
+{% extends "base.html" %}
+{% block title %}Logout{% endblock %}
+{% block pagetitle %}Logout{% endblock %}
+{% block content %}
+You are now logged out.
+{% endblock %}
```
giocalitri: please put an extra blank line at the end

odl_video/settings.py
```diff
@@ -104,6 +104,9 @@ def get_var(name, default):
 # Application definition
 
 INSTALLED_APPS = [
+    'ui.apps.UIConfig',
+    'cloudsync.apps.CloudSyncConfig',
+    'dj_elastictranscoder',
```
giocalitri: can you please explain me why the change of order and if you see any side effect by doing this?

mbertrand: `ui` in particular had to be before `django.contrib.admin` to override the admin logout page.  I decided to also place `cloudsync` and `dj_elastictranscoder` above the standard django apps to preemptively avoid similar issues.  I ran unit tests and manually tested the app to see if there were any unintended side effects but did not see any.  If you prefer I can put `cloudsync` and `dj_elastictranscoder` back in their original order.

giocalitri: I have no preferences, but I was wondering if you looked for possible ones.
I remember from reading documentation for unrelated issues that the sequence is pretty important, so I thought you looked into it.
Anyway I think we will find out.

ui/templates/registration/logged_out.html
mbertrand: Blank line added

giocalitri:
👍

PR #80, opened 2017-07-26
Removed secrets files and using only env variables

### LINKED ISSUE #78: Secrets cannot be in file
We do not have a way in heroku to load secrets from files, so we  need to move everything to environment variables.

giocalitri:
closes #78 

#### What's this PR do?
Removes all the secrets from files and puts them in environment variables

move the secrets to the env variables and test that everything works as before


README.rst
```diff
@@ -14,34 +14,38 @@ Installation
...
 ~~~~~~
-Create a secret key for Django, and store it in the file
-``secrets/django-secret-key``. You can run this code to do so:
+Create a secret key for Django, and store it in the ``.env`` file as SECRET_KEY.
+You can run this code to create a key:
 
 .. code-block:: bash
 
-    head -c 50 /dev/urandom > secrets/django-secret-key
+    head -c 50 /dev/urandom
```
noisecapella: This produces a sequence of random bytes. Maybe we should add a `| base64` to this to get something they can paste into `.env`

README.rst
```diff
@@ -14,34 +14,38 @@ Installation
...
 
-    [default]
-    aws_access_key_id=foo
-    aws_secret_access_key=bar
+    AWS_ACCESS_KEY_ID=foo
+    AWS_SECRET_ACCESS_KEY=bar
 
 You'll also need a CloudFront private key for generating signed URLs for
-CloudFront. Store the private key file in ``secrets/cloudfront-key.pem``.
+CloudFront. Store the private key file in ``.env`` in one singl string
```
noisecapella: `singl` should be `single`

README.rst
```diff
@@ -93,24 +97,26 @@ Dropbox
...
-``secrets/dropbox-credentials.ini``, like this:
+and store the app key in the file
+``.env``, like this:
 
 .. code-block:: ini
 
-    [default]
-    dropbox_app_key=foo
-    dropbox_app_secret=bar
+    DROPBOX_KEY=foo
```
noisecapella: Do we not need the secret for our app?

odl_video/envs.py
```diff
@@ -134,3 +134,22 @@ def get_any(name, default):
...
+
+     Args:
+        name (str): An environment variable name
+        default (str): The default value to use if the environment variable doesn't exist.
+
+    Returns:
+        bytes: The environment variable value, or the default as bytestring
+    """
+    value = get_string(name, default)
+    if not value:
```
noisecapella: Can we check `isinstance(value, str)` instead so that the default value can be parsed if it's a string?

README.rst
giocalitri: sure

README.rst
giocalitri: sure

README.rst
giocalitri: As far as I know we are not using it. I tested my branch and I am able to upload without it.

odl_video/envs.py
giocalitri: sure

ui/views.py
```diff
@@ -87,6 +85,7 @@ class VideoDetail(DetailView):
     model = Video
 
     def get_context_data(self, **kwargs):
+        # import pdb; pdb.set_trace()
```
noisecapella: Stray comment

noisecapella:
Code looks good, just waiting for the video to transcode to do a functional test

PR #86, opened 2017-07-27
Fix videojs error on heroku

### LINKED ISSUE #85: 'e is not defined' when loading video with production webpack JS
On the video detail page, in a heroku environment, the video will not play and the error 'e is not defined' appears in the JS console.

mbertrand:
Fixes #85 

#### What's this PR do?
Modifies `webpack.config.prod.js` by adding a DefinePlugin to alias `typeof global` to `undefined` based on recommendation here: https://github.com/videojs/videojs-contrib-hls/issues/600#issuecomment-295730581

### DIFF
--- a/webpack.config.prod.js
+++ b/webpack.config.prod.js
@@ -27,6 +27,11 @@ module.exports = Object.assign(prodConfig, {
         'NODE_ENV': '"production"'
       }
     }),
+    // This is necessary to make videojs & uglify work together:
+    // https://github.com/videojs/videojs-contrib-hls/issues/600#issuecomment-295730581
+    new webpack.DefinePlugin({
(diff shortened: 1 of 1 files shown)


webpack.config.prod.js
```diff
@@ -27,6 +27,9 @@ module.exports = Object.assign(prodConfig, {
         'NODE_ENV': '"production"'
       }
     }),
+    new webpack.DefinePlugin({
+      'typeof global': JSON.stringify('undefined')
+    }),
```
giocalitri: nit please put a comment about why this

giocalitri:
👍  after the comment is addressed

webpack.config.prod.js
noisecapella: Ah, good catch

Issue #58, opened 2017-07-19
Mechanism for updating the status of video transcoding jobs


mbertrand:
Currently, when a user visits a video detail page, if the video status is not yet 'Complete' or 'Error', a call will be made to AWS ElasticTranscoder to check on the status of the transcode job, and then update the video status based on the result.  This should be done automatically instead, and there are a few options for this:

1. Use SNS in conjunction with django-elastic-transcoder's 'endpoint' URL to send transcoding status updates directly to the app.  This should work fine on a live site, but would have to be manually simulated on a local development build.

2. Have SNS send status updates to an SQS message queue instead, and then have a recurring celery task to process the messages in the queue. 

3. Have a recurring celery task that queries the database for all videos whose status is not yet completed or failed, and then queries AWS ElasticTranscoder for the status of each one's transcode job for any possible updates.

4. The celery task that creates the transcode job could also continue running [while checking the job status until the job has completed or failed](http://boto3.readthedocs.io/en/latest/reference/services/elastictranscoder.html#waiters), and then update the video status based on that result.

mbertrand:
@giocalitri  @pdpinch Testing an SNS HTTP endpoint locally (option 1) can be achieved by [using ngrok](https://aws.amazon.com/blogs/developer/testing-webhooks-locally-for-amazon-sns/).  This would require creating an SNS subscription to an ngrok URL that would change every time ngrok is started (unless we get a paid plan that enables more permanent URL's).  

```
Session Status                online
Account                       Matt Bertrand (Plan: Free)
Version                       2.2.8
Region                        United States (us)
Web Interface                 http://127.0.0.1:4040
Forwarding                    http://d2f9ac8c.ngrok.io -> localhost:8089
Forwarding                    https://d2f9ac8c.ngrok.io -> localhost:8089
```

mbertrand:
Option 3 implemented in #90

PR #90, opened 2017-07-27
Improved error handling for tasks

mbertrand:
This is the same as #50, which was accidentally closed.

Related to https://github.com/mitodl/odl-video-service/issues/24 and https://github.com/mitodl/odl-video-service/issues/58

#### What's this PR do?
- New statuses for tasks to better track progress (Uploading, Transcoding, Upload Failed, Transcode Failed).
- Overrides default Transcoder.create_job class to record more detailed results info in EncodeJob objects
- Uses try-catch blocks to update statuses of videos and tasks when failures occur in celery tasks.
- Fixes progress bar on upload page.

- Upload a video file that's about 50 MB or larger.  Progress bar should begin filling in shortly.  Look at network requests and responses: responses should show amount of bytes transferred. When upload is complete, requests should stop being sent and the last response should show a status of 'Transcoding'.  Go to the video detail page, the video 'Encoding Status' should also be 'Transcoding'.  When transcoding is finished, the status should be shown as 'Complete'
- Choose a file to upload from dropbox, then rename the dropbox file after it appears on the upload list but before the progress bar starts filling in.  Eventually, after 15 minutes or so, the upload should fail after AWS gives up trying. Go to the video detail page and the 'Encoding Status' should be 'Upload failed'
- Rename a non-video dropbox file as "myvideo.mp4" and upload it.  Upload should complete successfully but transcoding should eventually fail, and the 'Encoding status' on the video detail page should be 'Transcoding failed'
- Go to ../admin/dj_elastictranscoder/encodejob/ and you should see EncodeJob objects for all videos whose transcoding tasks passed or failed, along with messages that contain the job results details.

### DIFF
--- a/cloudsync/conftest.py
+++ b/cloudsync/conftest.py
@@ -1,9 +1,6 @@
 """
 conftest for pytest in this module
 """
-import io
-import os
-
 import pytest
 import requests_mock
 import botocore.session
@@ -18,12 +15,6 @@ def reqmocker():
 
 
 @pytest.fixture
-def mock_video_url():
-    """Mocks video url"""
-    return "http://example.com/video.mp4"
-
-
-@pytest.fixture
 def mock_video_headers():
     """mocks video headers"""
     disp = "inline; filename=video.mp4; filename*=UTF-8''video.mp4"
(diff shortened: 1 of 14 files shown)


giocalitri:
👍

PR #102, opened 2017-07-31
Introduced mandatory settings

### LINKED ISSUE #72: Check for null/blank values for critical settings and raise an error
If settings required for upload/transcoding are blank, raise an exception on application startup.

giocalitri:
closes #72

#### What's this PR do?
checks for mandatory settings at the app startup

remove at least one of the mandatory settings from the .env and the app should not start

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,6 @@ COPY . /src
 WORKDIR /src
 RUN chown -R mitodl:mitodl /src
 
-# Gather static
-RUN ./manage.py collectstatic --noinput
 RUN apt-get clean && apt-get purge
 USER mitodl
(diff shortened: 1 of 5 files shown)


giocalitri:
@mbertrand I talked to George I removed the command from the docker file because we do not actually collect the static assets in that way.

mbertrand:
There's no styling for the admin console unless I run collectstatic manually, is that okay?  Does the deployment script handle that?  Otherwise everything looks good to me.

giocalitri:
oh.... that was unplanned.
I need another solution

giocalitri:
@mbertrand  now it should work

PR #105, opened 2017-08-01
Added channels model and removed user from Video model

### LINKED ISSUE #51: As a collection admin, I need to create a collection
Video uploaders (collection admins) need a way to collect videos that have similar admin and view permissions. For now, anyone who can log in can create a collection. 

A collection admin will be required to create a collection before they can upload video.

Collection metadata includes:
- a title (required, used for displaying a list of channels)
- a description (optional)
- owner/creator (django user) 

### acceptance criteria
- [ ] create collection model
- [ ] change video model to associate videos with collections instead of users
- [ ] associate user to collection 
- [ ] basic interface (no styling)

### future issues 
- view permissions (moira list, optional) 
- support for moira lists in admin permissions
- admin permissions (moira list, optional)
  "admin" permissions are the same as the owner's permissions

giocalitri:
closes #51 

#### What's this PR do?
Adds the channels model, rest APIs for them and changes the way the site can be navigated

Start the app, create a channel and upload videos in that channels

This is not a backwards compatible migration. I suggest to delete all the videos from the app **BEFORE** applying this migration, or you just need to manually associate the orphan videos to a new channel from the admin.


ui/models.py
```diff
@@ -20,6 +20,16 @@
...
 
 
+class Channel(models.Model):
+    """
+    Model for Video Channels
+    """
+    title = models.TextField()
+    description = models.TextField(null=True, blank=True)
+    slug = models.UUIDField(unique=True, null=False, blank=False, default=uuid4)
+    owners = models.ManyToManyField(settings.AUTH_USER_MODEL)
```
mbertrand: Any advantage of this approach vs having one creator/owner and using moira_lists to determine who can administer/edit/view?

giocalitri: Not really. The idea was to avoid to corner ourselves in a spot where we need to have different users that actually own the channel.
But the more I think about it, the less I am convinced.
I think I will move it to a foreignKey

giocalitri:
Ok. I think it is time to write some tests around.

ui/api_test.py
```diff
@@ -1,122 +1,93 @@
...
+
 
 pytestmark = pytest.mark.django_db
 
 
-def test_video_job_status_error(mocker, video, encodejob):  # pylint: disable=unused-argument
+def test_process_dropbox_data_happy_path(mocker):
     """
-    Verify that Video.job_status property returns the status of its encoding job
+    Tests that the process_dropbox_data in case everething is fine
```
mbertrand: Minor type `everething`

ui/forms.py
```diff
@@ -52,24 +52,38 @@ def to_python(self, values):
...
+        fields = ('title', 'description', 'owner', )
+
+
 class VideoForm(ModelForm):
     """
     Video Form
     """
     moira_lists = DynamicModelMultipleChoiceField(
-        model=MoiraList, field="name",
+        model=models.MoiraList,
```
mbertrand: Maybe `moira_lists` should be associated with collections instead of videos now?

ui/models.py
```diff
@@ -20,6 +20,36 @@
...
 
+class Collection(models.Model):
+    """
+    Model for Video Collections
+    """
+    key = models.UUIDField(unique=True, null=False, blank=False, default=uuid4)
+    title = models.TextField()
+    description = models.TextField(null=True, blank=True)
+    owner = models.ForeignKey(settings.AUTH_USER_MODEL)
+
```
mbertrand: Add `moira_lists` to `Collection` and remove from `Video`?

giocalitri:
@mbertrand I updated the models and fixed related problems with REST APIs and serializers.
I completely removed the moira lists from the UI for the time being.

ui/templates/ui/collection_list.html
```diff
@@ -0,0 +1,63 @@
...
+  {% for collection in collection_list %}
+    <li><a href="{% url "collection-detail" collection.hexkey %}">{{ collection.title }}</a></li>
+  {% empty %}
+    <li><i>Please create a collection first</i></li>
+  {% endfor %}
+</ul>
+
+<hr>
+
+<h2>Create New Collection</h2>
```
mbertrand: @giocalitri if you want to hold off on the form issue until the next PR for #118 that's fine with me.  Everything else works great.

giocalitri: that is related to the fact that the current UI does not makes ajax requests with the right content type.
Anyway I put in place a quick fix and test

mbertrand:
👍

Issue #108, opened 2017-08-01
USwitch player JS only works with specific MIT domains


mbertrand:
The USwitch player throws a JS error unless the domain from which it is loaded is localhost, mit.edu, or amps-web.amps.ms.mit.edu

The company which produces the player is working on modifications to fix this.

mbertrand:
Received the new modifications but it will also require a slight change to the React component that renders it in OVS.

mbertrand:
The change that needs to be made to the React component's `render` method:
```
        <div ref={ node => this.videoNode = node }  allowFullScreen id='omniPlayer'>
          <video id='video' playsinline webkit-playsinline></video>
        </div>
```

mbertrand:
@pdpinch is it okay for me to work on this one this week since it's related to #109?

pdpinch:
Yes

Issue #104, opened 2017-08-01
Omniplayer CSS needs modification


mbertrand:
The Omniplayer CSS file contains some global styles that are not easily overridden, for instance:

```
* {
    margin: 0;
    padding: 0;
}
```

These should be removed from the file or modified to only affect elements within the player div.

mbertrand:
This is no longer an issue now that the USwitch player being placed in an iframe.

Issue #103, opened 2017-07-31
Make OmniPlayer javascript files available to the app


mbertrand:
Since OmniPlayer is not open source it can't be included in the odl-video-service repo.  Upload all its files (JS, CSS, images) to S3 and make them available via URL, for example:

```
https://s3.amazonaws.com/odl_mit_video/omniplayer/lib/omni-external-libs.js
```

Related to #93, which depends on this being in place.

mbertrand:
Alternatively, if OmniPlayer can be transformed into a yarn-compatible JS package, put it in a private repo and use yarn and webpack to integrate it.

mbertrand:
Current solution is to host the files on S3.  Waiting to here back from OmniLive whether or not they can package the files to be yarn-compatible.

mbertrand:
OmniLive provided the new files in the standard format, not as a package.  So sticking with external url's for now

pdpinch:
Isn't this done?

mbertrand:
Yes as long as the current approach is acceptable (external script tags instead of webpack integration).

pdpinch:
I'm afraid that's going to be up to the vendor.

PR #109, opened 2017-08-01
Add USwitch player to OVS

### LINKED ISSUE #93: Support USwitch player as an alternative to Video.js
Part of Lecture Capture Epic

The Lecture Capture feature uses a different video player which has the ability to mimic the playback of 4 synchronized video streams. This allows the user to "choose" between viewing the instructor, the chalkboard or either of two other video sources. (need example link)

This player may come with license restrictions which limit us to hosting it on an mit.edu domain

### Acceptance criteria:
- [x] Update this ticket once we determine if this is called the Unity player or the Omniplayer
- [x] The choice of player should be set at the video level. 
- [x] The video player template should import the chosen player (presumably via `<script>` tag)
- [x] Support the custom Omni/Unity Player playlist format (if necessary)
- [x] We want to continue to allow channel admins to upload their own video to a channel that will play with the standard video.js

###  Questions:
- [x] What is the player actually called, in the code we have licensed?
- [x] Do we need to make changes to the VideoObject model in order to support the custom playlist format?

mbertrand:
Fixes #93

#### What's this PR do?
- Adds a boolean field to the Video model ('multiangle')
- if the video is multiangle, loads the USwitch player instead of the video.js player

- Add this line to your `.env` file: `USWITCH_URL=https://s3.amazonaws.com/odl-video-service-uswitch-dev`
- `docker-compose build`
- `docker-compose run`
- Using the admin console, set 'multiangle=True' on any video
- Modify your `/etc/hosts` so that one of the following domains (`localhost`, `mit.edu`, `amps-web.amps.ms.mit.edu`) points to your docker host, for example `192.168.99.100 amps-web.amps.ms.mit.edu`
- Load the video detail page for the modified video, using `amps-web.amps.ms.mit.edu` as the domain.  You should see the USwitch player instead of the video.js player.
- Play the video, there will be one main window and 4 smaller windows at the top showing the 4 corners of the video, labeled Camera 1-4.

There is an issue with styling on the detail page when the USwitch player is used - #104 
The player only works with certain domains: #108
This assumes the USwitch JS/CSS will be hosted on S3: #103


noisecapella:
When I click on the fullscreen icon for this video player it makes the browser full screen, not the video itself. Is there some configuration setting relevant to this?

noisecapella:
When I switch from the multivideo view it pops up a window but the URL is not found:

screenshot from 2017-08-02 10-50-59

static/scss/layout.scss
```diff
@@ -25,3 +26,14 @@
     bottom: 0;
 }
 
+// OmniPlayer styles
```
noisecapella: At some point we should add a CSS linter, but in the mean time can you make the indentation consistent below?

noisecapella:
The player seems to absorb all keystrokes. If you click on the description textbox for example and type something, no text appears

static/js/entry/uswitch.js
```diff
@@ -0,0 +1,26 @@
+/* global SETTINGS: false */
+require('react-hot-loader/patch');
+const $script = require('scriptjs');
```
noisecapella: I'm not sure we should use `$script` here. The below async imports depend on one another so that they're loaded in sequence instead of in parallel. If we include the script tags in the HTML right before the `render_bundle` for this entry we should get the same behavior and the script tags will load in parallel too

ui/views.py
```diff
@@ -85,15 +85,24 @@ class VideoDetail(DetailView):
...
+                "src": videofile.cloudfront_url,
+                "title": video.title,
+                "description": video.description,
+            }
+            if video.multiangle:
+                context['uswitchPlayerURL'] = settings.USWITCH_URL
+                default_settings["uswitchPlayerURL"] = settings.USWITCH_URL
         except VideoFile.DoesNotExist:
             videofile = None
+            default_settings['videofile'] = ''
```
noisecapella: `videofile` is an object above but it's a string here. Can you make this `None` instead?

ui/views_test.py
```diff
@@ -68,3 +72,16 @@ def test_video_detail_unencoded(
...
+
+def test_video_detail_omniplayer(admin_client, video, videofileHLS, mocker):
+    """Test video detail page when Video.multiangle is True"""
+    mocker.patch('ui.utils.get_cloudfront_signed_url', return_value=videofileHLS.cloudfront_url)
+    video.status = 'Complete'
+    video.multiangle = True
+    video.save()
+    url = reverse('video-detail', kwargs={'pk': video.id})
+    response = admin_client.get(url)
+    assert 'uswitchPlayerURL' in response.context_data
```
noisecapella: Can you assert the value in the context here?

noisecapella: We should also set a fake value for `settings.USWITCH_URL` so that we know that it is the source of the value here. By default it's an empty string

mbertrand:
Looks like the USwitch player needs to be run in an iframe to avoid problems listed above (fullscreen mode not working correctly, text input absorbed), yuck.  Made that change.  Fixed the issue with the `mosaic.html` popup window by creating a new view and template for it, where the template contains the `div` tag from the original `mosaic.html` supplied with the player, and the script from that original HTML has been moved to S3 in a new file [mosaic.js](https://github.com/mitodl/omniplayer/blob/master/mosaic.js)

odl_video/test_settings.py
```diff
@@ -0,0 +1,5 @@
+"""
+Django test settings for odl_video project.
+"""
+from odl_video.settings import *  # pylint: disable=wildcard-import,unused-wildcard-import
+USWITCH_URL = 'https://testing_odl.mit.edu'
```
noisecapella: Instead of setting the URL here can we set it only in the test where we need it set? The `settings` fixture should let you modify settings on a per test basis

noisecapella:
Yes, all the functionality issues are fixed now

noisecapella:
Can you also revert the change to `yarn.lock` since none of the packages are changing anymore?

odl_video/test_settings.py
mbertrand: Do you have an example of that? I originally tried the following in `conftest.py`:
```
def pytest_configure():
    settings.configure(USWITCH_URL='https://testing_odl.mit.edu')
```

but then all the tests failed because 'settings were already configured', in pytest.ini: `ds=odl_video.settings`

mbertrand:
I removed `scriptjs` from yarn (the only new thing I had added), `videojs-contrib-quality-levels` should have been removed earlier (not being used), not sure what triggered the other changes in `yarn.lock`

odl_video/test_settings.py
mbertrand: Made the change

PR #111, opened 2017-08-07
Watch Bucket monitoring and processing

### LINKED ISSUE #101: Lecture Capture: parse S3 watch folder for new videos
As the technical services manager, I want lecture capture videos to be ingested into the OVS system and have transcoding started automatically. 

We will write a periodic task in OVS that checks for files in the S3 watch folder that the 1Beyond machines upload video to (see #100). For any files that exist in the bucket, the script will attempt to create a corresponding video object in OVS and start the transcode process. The created video object will be matched to an existing collection, or a new collection will be created. 

### Acceptance criteria
- [x] Alert DevOps on any error (OpsGenie? Sentry?)
- [x] create a video object for each video file encountered
  - [x] if the video doesn't match any existing collection, create one, owned by the user with the e-mail specified in`LECTURE_CAPTUR_USER`
- [x] After the video object is successfully created and the video is moved from the S3 watch bucket to the S3 upload bucket, it can be removed from the S3 watch bucket. This will allow us to treat the S3 watch bucket as a queue of sorts. 
- [ ] Turn on S3 versioning in the bucket, to avoid any accidental data loss.

mbertrand:
Fixes #101

#### What's this PR do?
Creates a scheduled celery task that will check for the existence of files in an S3 bucket (assigned via settings/environment variable).  Each file found will be copied to the OVS upload bucket, Video and Videofile model objects will be created for it (with `multiangle=True`), and finally the original file will be deleted from the watch bucket.

- Set the following in your .env file:
```
VIDEO_S3_WATCH_BUCKET=odl-video-service-watch-ci
VIDEO_WATCH_BUCKET_FREQUENCY=30
USWITCH_URL=https://s3.amazonaws.com/odl-video-service-uswitch-dev/ci
LECTURE_CAPTURE_USER=<username of existing user>
```
- `docker-compose build`
- `docker-compose up`
- Upload a video file to the `odl-video-service-watch-ci` bucket with an S3 name similar to `MIT-6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4`.  Within 30 seconds, the file should start to be processed.
- Once the video has started to be transcoded, the original file should no longer exist in `VIDEO_S3_WATCH_BUCKET`
- The video file should be in the `MIT-6.046-L01-2017-Spring` collection belonging to the user specified in `LECTURE_CAPTURE_USER` - you can either create the collection yourself or it will be created automatically.
- After being transcoded, the USwitch player should load on the video detail page.  You'll need to point any .mit.edu subdomain to your docker host IP for it to work


cloudsync/tasks.py
```diff
@@ -165,23 +223,51 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+    # Copy the file to the upload bucket using a new s3 key
+    s3_client = get_s3_client()
+    copy_source = {
+        'Bucket': watch_bucket.name,
+        'Key': key
+    }
+    try:
+        s3_client.copy(copy_source, settings.VIDEO_S3_BUCKET, video_file.s3_object_key)
+    except:
+        video.delete()
```
giocalitri: this needs to be in a try/except of its own in case something goes wrong

cloudsync/tasks.py
```diff
@@ -165,23 +223,51 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+        raise
+
+    # Delete the original file from the watch bucket
+    s3_client.delete_object(Bucket=settings.VIDEO_S3_WATCH_BUCKET, Key=key)
```
giocalitri: what if this fails? is there a way to recover the state?

cloudsync/tasks_test.py
```diff
@@ -229,3 +236,106 @@ def test_stream_to_s3_no_video():
...
+    user.username = 'admin'
+    user.save()
+    with pytest.raises(ClientError) as exc:
+        monitor_watch_bucket.delay()
+    assert 'NoSuchBucket' in str(exc)
+    mocker.patch('cloudsync.tasks.process_watch_file',
+                 side_effect=ValueError('Fake ValueError'))
+    with pytest.raises(ValueError) as exc:
+        monitor_watch_bucket.delay()
+    assert 'Fake ValueError' in str(exc)
```
giocalitri: these tests are not sufficient: we need to test each of the major variation of the APIs and especially all the cases where they can fail.
we also need to assert for each mocked function, that it has been called with the expected parameters.

cloudsync/tasks_test.py
```diff
@@ -229,3 +236,106 @@ def test_stream_to_s3_no_video():
...
+                'MIT-6.046-2018-Fall',
+                'MIT-6.046',
+                'MIT-6.046-L02',
+                'MIT-STS.001-2015-Fall-L03',
+                'MIT-HST.969-2018-Spring-L01',
+                'NUS-7.S939-2019-Spring-L05',
+                'MIT-Adhoc-2-190',
+                'MIT-18.test-L01',
+                'MIT-18.test')
+    for item in zip(filenames, expected):
```
giocalitri: I think that is better if you put together the two list in tuples where each tuple is input and output. This is already difficult to read.

cloudsync/tasks_test.py
```diff
@@ -229,3 +236,106 @@ def test_stream_to_s3_no_video():
...
+    bad_names = ('MIT-6.046-lec-mit-0000-2017apr06-L01.mp4',  # Missing recording time
+                 'MIT-6.046-lec-mit-0000-1510-L01.mp4',  # Missing recording date
+                 'MIT-6.046-1510-L01.mp4',  # Missing -lec-mit-0000
+                 'Completely random name.mp4',  # No similarity to expected format
+                 'MIT-6.046-lec-mit-0000-2017apr06-L01',  # no extension
+                 '6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4')  # Missing prefix
+
+    for name in bad_names:
+        with pytest.raises(ValueError):
+            extract_title(name)
```
giocalitri: In generale for both tests: what about non standard characters? do we care?

odl_video/settings.py
```diff
@@ -480,3 +484,4 @@ def get_all_config_keys():
 ]
 # This is necessary for loading USwitch video player code from an external source
 USWITCH_URL = get_string('USWITCH_URL', '')
+LECTURE_CAPTURE_USER = get_string('LECTURE_CAPTURE_USER', 'admin')
```
giocalitri: I do not think we need a default inexistent user here that might fail.

I also would like something that verify that the user set here actually exists at the app startup (where I added the settings verifications) so that we can return a meaningful error in case it does not exist.

ui/utils.py
```diff
@@ -152,3 +152,8 @@ def get_bucket(bucket_name):
 def get_transcoder_client():
     """Get an ElasticTranscoder client object"""
     return boto3.client('elastictranscoder', settings.AWS_REGION)
+
+
+def get_s3_client():
+    """Get an S3 client object"""
+    return boto3.client('s3')
```
giocalitri: where are you using this? if the only place is in the watch API, you can move the code there, otherwise please add tests and a better docstring for this.

cloudsync/tasks.py
mbertrand: What would be the appropriate action to take in this case, besides raising it?

cloudsync/tasks.py
mbertrand: If it fails then it would likely be imported again on the next task run, leading to a dupe video in the collection.  In that case maybe it would be best to delete the video & videofile just created?

cloudsync/tasks.py
```diff
@@ -123,7 +126,63 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+        try:
+            process_watch_file(key.key)
+        except (ClientError, Exception, ) as exc:  # pylint: disable=broad-except
+            # Log any exception, raise later so other files can be processed.
+            msg = exc.response if isinstance(exc, ClientError) else str(exc)
+            log.exception("AWS error when ingesting file from watch bucket %s: %s", key.key, msg)
+            error = exc
+    if isinstance(error, Exception):
+        # Raise the last exception
+        raise error   # pylint: disable=raising-bad-type
```
mbertrand: I am not raising the error before, just logging it, so that an exception with one file won't prevent other files from being processed.  Then the last error logged (if any) will be raised once all files have been tried.

cloudsync/tasks.py
```diff
@@ -165,23 +223,51 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
-            video.update_status(VideoStatus.TRANSCODE_FAILED)
-        except ClientError as exc:
-            # Log the exception but don't raise it so other videos can be checked.
-            log.exception("AWS error when refreshing job status for video %d: %s", video.id, exc.response)
-            video.update_status(VideoStatus.TRANSCODE_FAILED)
+    watch_bucket = get_bucket(settings.VIDEO_S3_WATCH_BUCKET)
+
+    title = extract_title(key)
+    collection, _ = Collection.objects.get_or_create(title=title,
+                                                     owner=User.objects.get(username=settings.LECTURE_CAPTURE_USER))
```
mbertrand: Yes, but I think just letting the exception raise here is ok? No other actions necessary because there is nothing yet that has to be modified, deleted, etc if it does happen?

cloudsync/tasks.py
```diff
@@ -209,3 +295,21 @@ def parse_content_metadata(response):
...
+    """
+    Parses the filename for required course information
+    Args:
+        filename(str): The name of the video file, in format
+        'MIT-<course#>-<year>-<semester>-lec-mit-0000-<recording_date>-<time>-<session>.mp4'
+
+    Returns:
+        A string which should match a collection title: MIT-<course>-<session>-<year>-<semester>
+    """
+    rx = r'([A-Z]+)-([\w\.\s]+)-((\d{4})-(Fall|Spring)-)?lec-mit-0000-(\w+)-(\d+)(-([L\d\-]+))?(_\d+)?.*\.\w'
```
mbertrand: If the regex changes significantly, the function's code that determines what part of the regex match equals course, session, etc might also need to change.

cloudsync/tasks.py
pdpinch: It is, but I wouldn't mind knowing when you think this will become a problem.

cloudsync/tasks.py
pdpinch: I think we should keep in mind what Elaine's inputs look like on the front end. I know that she is using things like course number and section number, but those are by convention and aren't enforced to be in the formats you are looking for. 

I think it would be safer, easier and more flexible to just extract the two strings that Elaine inputs as parameters to the 1Beyond script, something like:

collection_prefix = "^(.*)lec-mit-0000-"
collection_suffix = "lec-mit-0000-(\w+)-(\d+)(-([L\d\-]+))" #do we really need all these parens?
collection_title = collection_prefix + collection_suffix

cloudsync/tasks_test.py
pdpinch: I think we should have a test for unicode characters in the file names.

cloudsync/tasks.py
mbertrand: I like the `collection_prefix`.
The `collection_suffix` above will be different for every video, as it includes recording date, recording time, and occasionally a room number (for adhoc files) or an iterator for the session ("L01_1", "L01_2").  So we'd end up with a different collection for each file if we used the whole thing.  Need to just extract the session from there if present, and room number for adhoc files (but I'll ask Elaine if room number is really necessary for those).

cloudsync/tasks.py
giocalitri: Sorry, but why do you want to raise an exception at this point of the code?

cloudsync/tasks.py
giocalitri: ok... I am fine with leaving as is if you put in place a check for the existence of the user at the app startup

cloudsync/tasks.py
giocalitri: maybe add some logging to say that this is a leftover of a failure caused by another failure so we know we can delete the object safely.
maybe also add the object id so we can easily find it...

cloudsync/tasks.py
giocalitri: well... maybe it is enough to log about the failure and add some informations about which video was this call associated with.

cloudsync/tasks.py
giocalitri: ok, got it

cloudsync/tasks.py
mbertrand: Just as an indication that the celery task didn't fully succeed. But since the individual errors will be logged I will go ahead and remove it.

cloudsync/api.py
```diff
@@ -74,3 +81,128 @@ def refresh_status(video, encode_job=None):
...
+    except ClientError as exc:
+        log.error('Transcode job creation failed for video %s', video.id)
+        video.update_status(VideoStatus.TRANSCODE_FAILED)
+        if hasattr(exc, 'response'):
+            transcoder.message = exc.response
+        raise
+    finally:
+        transcoder.create_job_for_object(video)
+        if video.status != VideoStatus.TRANSCODE_FAILED:
+            video.update_status(VideoStatus.TRANSCODING)
```
giocalitri: question: when can happen that `video.status == VideoStatus.TRANSCODE_FAILED`?

cloudsync/api.py
```diff
@@ -74,3 +81,128 @@ def refresh_status(video, encode_job=None):
...
+
+
+def process_watch_file(key):
```
giocalitri: nit: we might want to call `key` something like `s3key` (there are too many keys around)

cloudsync/api.py
```diff
@@ -74,3 +81,128 @@ def refresh_status(video, encode_job=None):
...
+    watch_bucket = get_bucket(settings.VIDEO_S3_WATCH_BUCKET)
+
+    title = extract_title(key)
+    collection, _ = Collection.objects.get_or_create(title=title,
+                                                     owner=User.objects.get(username=settings.LECTURE_CAPTURE_USER))
+
+    # Create the necessary models
+    with transaction.atomic():
+        video = Video.objects.create(
+            source_url='https://{}/{}/{}'.format(settings.AWS_S3_DOMAIN, settings.VIDEO_S3_WATCH_BUCKET, key),
```
giocalitri: this makes me think that we are going to introduce a "retry" functionality and we cannot really retry in this case.
Should we add something in the model to specify that this is a S3 source instead of dropbox?
Maybe let's open an issue about this.

cloudsync/api.py
```diff
@@ -74,3 +81,128 @@ def refresh_status(video, encode_job=None):
...
+    rx = (r'(.+)-lec-mit-0000-'  # prefix to be used as the start of the collection name
+          r'(\w+)'  # Recording date (required)
+          r'-(\d+)'  # Recording time (required)
+          r'(-([L\d\-]+))?'  # Session or room number (optional)
+          r'.*\.\w')  # Rest of filename including extension (required)
+    matches = re.search(rx, filename)
+    if not matches or len(matches.groups()) != 5:
+        raise VideoFilenameError('No matches found for filename %s with regex %s', filename, rx)
+    prefix, _, _, _, session = matches.groups()
+    return '-'.join([i for i in (prefix, session) if i])
```
giocalitri: nit: please do not use single character variables.

giocalitri: Also what does it mean that `if i` is there? can we have empty strings?

cloudsync/api_test.py
```diff
@@ -1,25 +1,45 @@
...
+@pytest.fixture()
+def video():
+    """Fixture to create a video"""
+    return VideoFactory()
+
+
+@pytest.fixture()
+def videofile(video):
+    """Fixture to create a videofile"""
+    return VideoFileFactory(video=video)
```
giocalitri: I do not think this is a good approach with the fixtures: the `VideoFileFactory` takes care of creating a video if not passed; unless I misunderstood what you are trying to do, there is not need for passing the video fixture as parameter here.

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+                 ('MIT-6.046-lec-mit-0000-2017apr06-0404-L02.mp4', 'MIT-6.046-L02'),
+                 ('MIT-STS.001-2015-Fall-lec-mit-0000-2015dec21-2222-L03_2.mp4', 'MIT-STS.001-2015-Fall-L03'),
+                 ('MIT-HST.969-2018-Spring-lec-mit-0000-2018mar09-0001-L01.mp4', 'MIT-HST.969-2018-Spring-L01'),
+                 ('NUS-7.S939-2019-Spring-lec-mit-0000-2019apr02-1210-L05.mp4', 'NUS-7.S939-2019-Spring-L05'),
+                 ('MIT-Adhoc-lec-mit-0000-2017aug10-0954-2-190.mp4', 'MIT-Adhoc-2-190'),
+                 ('MIT-18.test-lec-mit-0000-2017aug10-0956-L01.mp4', 'MIT-18.test-L01'),
+                 ('MIT-18.test-lec-mit-0000-2017aug10-0956.mp4', 'MIT-18.test'),
+                 ('MIT-Simons Lecture Series-lec-mit-0000-2017aug10-0956-2-10.mp4', 'MIT-Simons Lecture Series-2-10'),
+                 ('/&*3:<>俺正和-lec-mit-0000-2017aug10-0956.mp4', '/&*3:<>俺正和'))
+    for item in filenames:
```
giocalitri: nit, please unpack the tuple with meaningful variable names, so something like
`for input, output in filenames:`

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    bad_names = ('MIT-6.046-1510-L01.mp4',  # Missing -lec-mit-0000
+                 'Completely random name.mp4',  # No similarity to expected format
+                 'MIT-6.046-lec-mit-0000-2017apr06-L01')  # no extension
+
+    for name in bad_names:
+        with pytest.raises(VideoFilenameError):
+            extract_title(name)
+
+
+@pytest.mark.django_db
```
giocalitri: please set this globally at the module level using
`pytestmark = pytest.mark.django_db`
and you can remove the decorator from the single tests.

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+@mock_s3
+@override_settings(LECTURE_CAPTURE_USER='admin')
+def test_watch_nouser():
+    """
+    Test that an exception is correctly handled
```
giocalitri: exception caused by not having an user?

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    """
+    s3 = boto3.resource('s3')
+    s3c = boto3.client('s3')
+    upload = 'MIT-6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), upload)
+    with pytest.raises(User.DoesNotExist) as exc:
+        process_watch_file(upload)
+        assert 'User matching query does not exist' in str(exc)
```
giocalitri: maybe we can remove this assert given that we already check `User.DoesNotExist`.
in general, unless they are custom exceptions, it is not worth to check for exception string messages.

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+@override_settings(LECTURE_CAPTURE_USER='admin')
+def test_watch_nouser():
+    """
+    Test that an exception is correctly handled
+    """
+    s3 = boto3.resource('s3')
+    s3c = boto3.client('s3')
+    upload = 'MIT-6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
```
giocalitri: do you think it is worth to create a fixture to handle this s3 setup? you seem to do the same in few tests

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    with pytest.raises(Video.DoesNotExist):
+        Video.objects.get(title=upload)
+
+
+@pytest.mark.django_db
+@mock_s3
+@override_settings(LECTURE_CAPTURE_USER='admin')
+def test_watch_s3_error():
+    """Test that an AWS S3 ClientError is correctly handled"""
+    user = UserFactory(username='admin')  # pylint: disable=unused-variable
```
giocalitri: nit: you can remove `user =` if you are not going to use the variable

cloudsync/api.py
```diff
@@ -1,19 +1,26 @@
...
 import logging
+import re
 
+import boto3
+from botocore.exceptions import ClientError
 from django.conf import settings
+from django.contrib.auth.models import User
+from django.db import transaction
 
+from cloudsync.utils import VideoFilenameError, VideoTranscoder
```
giocalitri: please move the `VideoFilenameError` to an `exceptions` module

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    s3 = boto3.resource('s3')
+    s3c = boto3.client('s3')
+    upload = 'MIT-6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), upload)
+    with transaction.atomic():
```
giocalitri: why do you care about the transaction here?

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    """Test that a bad filename raises a VideoFilenameError"""
+    user = UserFactory(username='admin')  # pylint: disable=unused-variable
+    s3 = boto3.resource('s3')
+    s3c = boto3.client('s3')
+    upload = 'Bad filename.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), upload)
+    with transaction.atomic():
```
giocalitri: also here

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+        with pytest.raises(VideoFilenameError) as exc:
+            process_watch_file(upload)
+            assert 'No matches found for filename' in str(exc)
+    with pytest.raises(Video.DoesNotExist):
+        Video.objects.get(title=upload)
```
giocalitri: nit: you can also simply `assert not Video.objects.filter(title=upload).exists()` and not dealing with the exception

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    user = UserFactory(username='admin')  # pylint: disable=unused-variable
+    s3 = boto3.resource('s3')
+    s3c = boto3.client('s3')
+    upload = 'MIT-6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), upload)
+    process_watch_file(upload)
+    mock_encoder.assert_called_once()
```
giocalitri: please assert also with which parameters it has been called

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    """
+    Test that video status is updated properly after a transcode job is successfully created
+    """
+    mocker.patch.multiple('cloudsync.tasks.settings',
+                          ET_PRESET_IDS=('1351620000001-000061', '1351620000001-000040', '1351620000001-000020'),
+                          AWS_REGION='us-east-1', ET_PIPELINE_ID='foo')
+    mock_encoder = mocker.patch('cloudsync.api.VideoTranscoder.encode')
+    transcode_video(video, videofile)  # pylint: disable=no-value-for-parameter
+    mock_encoder.assert_called_once()
+    assert len(video.encode_jobs.all()) == 1
```
giocalitri: nit: you could use `video.encode_jobs.count()` instead of `len`

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    assert Video.objects.get(id=video.id).status == VideoStatus.TRANSCODING
```
giocalitri: please filter on the status and id here so if the test fails purest would give better log

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    """
+    job_result = {'Job': {'Id': '1498220566931-qtmtcu', 'Status': 'Error'}, 'Error': {'Code': 200, 'Message': 'FAIL'}}
+    mocker.patch.multiple('cloudsync.tasks.settings',
+                          ET_PRESET_IDS=('1351620000001-000061', '1351620000001-000040', '1351620000001-000020'),
+                          AWS_REGION='us-east-1', ET_PIPELINE_ID='foo')
+    mock_encoder = mocker.patch('cloudsync.api.VideoTranscoder.encode',
+                                side_effect=ClientError(error_response=job_result, operation_name='ReadJob'))
+    with pytest.raises(ClientError):
+        transcode_video(video, videofile)
+    mock_encoder.assert_called_once()
```
giocalitri: please assert with what it has been called

cloudsync/api_test.py
```diff
@@ -130,3 +150,148 @@ def test_process_transcode_results(mocker):
...
+    assert len(video.encode_jobs.all()) == 1
+    assert Video.objects.get(id=video.id).status == VideoStatus.TRANSCODE_FAILED
```
giocalitri: same comments as before

cloudsync/tasks_test.py
```diff
@@ -229,3 +234,59 @@ def test_stream_to_s3_no_video():
     """Test DoesNotExistError"""
     with pytest.raises(Video.DoesNotExist):
         stream_to_s3(999999)  # pylint: disable=no-value-for-parameter
+
+
+@pytest.mark.django_db
```
giocalitri: please move this at the module level

cloudsync/tasks_test.py
```diff
@@ -229,3 +234,59 @@ def test_stream_to_s3_no_video():
...
+               'Bad Name.mp4',
+               'MIT-6.046-lec-mit-0000-2017apr06-0404.mp4')
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    for filename in uploads:
+        bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), filename)
+    monitor_watch_bucket.delay()
+    assert mock_encoder.call_count == 2
+    assert not Video.objects.filter(title=uploads[1])
+    assert Video.objects.get(title=uploads[0])
+    assert Video.objects.get(title=uploads[2])
```
giocalitri: I think you can simplify these 2 tests because you already tested the entire logic in the api, so here you can just mock the api and assert it has been called correctly.
if you need the api to raise, you can configure the mocked api to do so as well.

cloudsync/tasks.py
```diff
@@ -165,23 +223,51 @@ def transcode_from_s3(self, video_id):  # pylint: disable=unused-argument
...
+    # Create the necessary models
+    with transaction.atomic():
+        video = Video.objects.create(
+            source_url='https://{}/{}/{}'.format(settings.AWS_S3_DOMAIN, settings.VIDEO_S3_WATCH_BUCKET, key),
+            collection=collection,
+            title=key,
+            multiangle=True  # Assume all videos in watch bucket are multi-angle
+        )
+        video_file = VideoFile.objects.create(
+            s3_object_key=video.get_s3_key(),
```
giocalitri: please open an issue for this

cloudsync/api.py
```diff
@@ -74,3 +81,128 @@ def refresh_status(video, encode_job=None):
...
+        settings.AWS_ACCESS_KEY_ID,
+        settings.AWS_SECRET_ACCESS_KEY
+    )
+    try:
+        transcoder.encode(video_input, outputs, Playlists=playlists)
+    except ClientError as exc:
+        log.error('Transcode job creation failed for video %s', video.id)
+        video.update_status(VideoStatus.TRANSCODE_FAILED)
+        if hasattr(exc, 'response'):
+            transcoder.message = exc.response
```
mbertrand: The message is passed on the the EncodeJob object and saved there in `transcoder.create_job_for_object(video)`, the transcoder is not a model-based class.

giocalitri: got it, thanks

cloudsync/api.py
mbertrand: A few lines up from here, if a ClientError occurs when creating the transcode job.  Also later, when transcode job statuses are checked by the periodic celery task.

cloudsync/api.py
mbertrand: I think setting the source_url to an S3 link provides a way to differentiate them from videos downloaded from dropbox (which will all have dropbox domain links).  But it requires parsing the value and maybe it would be better to have another model field that's one of a few options 'S3', 'Dropbox', and maybe 'Direct Upload' if we ever introduce that.

cloudsync/api.py
mbertrand: Sometimes there is no session so `i` would be `None` in those cases.

cloudsync/api_test.py
mbertrand: Yes, not having a user with the `LECTURE_CAPTURE_USER` username

cloudsync/api_test.py
mbertrand: Maybe, but sometimes I need to create two buckets, sometimes just one

cloudsync/tasks_test.py
mbertrand: Here, I wanted to test that if multiple files are in the watch bucket and one of them can't be processed, the other ones still are, since the task will log exceptions but not raise them.

cloudsync/api.py
giocalitri: ohhhhhh sorry... now I understand the logic... this is in case the failure has not already happened and everything is fine
sorry for the confusion

cloudsync/api.py
giocalitri: yeah... that's why I think we need a specific issue because then we need to implement different behaviors according to it

cloudsync/api.py
giocalitri: got it

cloudsync/api_test.py
giocalitri: please add this to the docstring

cloudsync/api_test.py
giocalitri: ok, you can leave as is then

cloudsync/tasks.py
mbertrand: https://github.com/mitodl/odl-video-service/issues/134

cloudsync/tasks_test.py
giocalitri: ok, I understand, but, if you can, it is better to avoid re-testing the entire logic of the backend because it is really difficult to understand the difference.

cloudsync/api_test.py
mbertrand: Otherwise this happens: https://stackoverflow.com/questions/21458387/transactionmanagementerror-you-cant-execute-queries-until-the-end-of-the-atom

cloudsync/api_test.py
mbertrand: Here is an example of what it is called with, is it worth mocking this?

`({'Key': '183a1166b8b046148c8bdfa579970d24/video.mp4'}, [{'Key': 'transcoded/183a1166b8b046148c8bdfa579970d24/video_1351620000001-000061', 'PresetId': '1351620000001-000061', 'SegmentDuration': '10.0', 'ThumbnailPattern': 'thumbnails/183a1166b8b046148c8bdfa579970d24/video_thumbnail_{count}'}, {'Key': 'transcoded/183a1166b8b046148c8bdfa579970d24/video_1351620000001-000040', 'PresetId': '1351620000001-000040', 'SegmentDuration': '10.0'}, {'Key': 'transcoded/183a1166b8b046148c8bdfa579970d24/video_1351620000001-000020', 'PresetId': '1351620000001-000020', 'SegmentDuration': '10.0'}], Playlists=[{'Format': 'HLSv3', 'Name': 'transcoded/183a1166b8b046148c8bdfa579970d24/video__index', 'OutputKeys': ['transcoded/183a1166b8b046148c8bdfa579970d24/video_1351620000001-000061', 'transcoded/183a1166b8b046148c8bdfa579970d24/video_1351620000001-000040', 'transcoded/183a1166b8b046148c8bdfa579970d24/video_1351620000001-000020']}])`

giocalitri: yes: if for some reason we change the code and it gets called in a different way we can easily know, I think.
do you disagree?

cloudsync/api_test.py
giocalitri: ok, got it

cloudsync/api_test.py
```diff
@@ -1,25 +1,45 @@
...
 import pytest
+from botocore.exceptions import ClientError
 from django.conf import settings
+from django.contrib.auth.models import User
+from django.db import transaction
+from django.test import override_settings
 from moto import mock_s3
 
-from cloudsync.api import process_transcode_results, refresh_status
+from cloudsync.api import process_transcode_results, refresh_status, extract_title, process_watch_file, transcode_video
```
giocalitri: nit: there are enough to put them on multiple lines or do `from cloudsync import api`

cloudsync/api_test.py
```diff
@@ -130,3 +150,172 @@ def test_process_transcode_results(mocker):
...
+                 ('MIT-6.046-lec-mit-0000-2017apr06-0404-L02.mp4', 'MIT-6.046-L02'),
+                 ('MIT-STS.001-2015-Fall-lec-mit-0000-2015dec21-2222-L03_2.mp4', 'MIT-STS.001-2015-Fall-L03'),
+                 ('MIT-HST.969-2018-Spring-lec-mit-0000-2018mar09-0001-L01.mp4', 'MIT-HST.969-2018-Spring-L01'),
+                 ('NUS-7.S939-2019-Spring-lec-mit-0000-2019apr02-1210-L05.mp4', 'NUS-7.S939-2019-Spring-L05'),
+                 ('MIT-Adhoc-lec-mit-0000-2017aug10-0954-2-190.mp4', 'MIT-Adhoc-2-190'),
+                 ('MIT-18.test-lec-mit-0000-2017aug10-0956-L01.mp4', 'MIT-18.test-L01'),
+                 ('MIT-18.test-lec-mit-0000-2017aug10-0956.mp4', 'MIT-18.test'),
+                 ('MIT-Simons Lecture Series-lec-mit-0000-2017aug10-0956-2-10.mp4', 'MIT-Simons Lecture Series-2-10'),
+                 ('/&*3:<>俺正和-lec-mit-0000-2017aug10-0956.mp4', '/&*3:<>俺正和'))
+    for (filename, title) in filenames:
```
giocalitri: nit: you do not need the `(` and `)` here

cloudsync/api_test.py
```diff
@@ -130,3 +150,172 @@ def test_process_transcode_results(mocker):
...
+    s3 = boto3.resource('s3')
+    s3c = boto3.client('s3')
+    upload = 'MIT-6.046-2017-Spring-lec-mit-0000-2017apr06-0404-L01.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), upload)
+    with transaction.atomic():
+        with pytest.raises(ClientError) as exc:
+            process_watch_file(upload)
+            assert 'NoSuchBucket' in str(exc)
```
giocalitri: Do you think it is necessary to assert this? I would prefer that an upgrade to their library changing error messages does not break our tests

cloudsync/api_test.py
```diff
@@ -130,3 +150,172 @@ def test_process_transcode_results(mocker):
...
+    s3c = boto3.client('s3')
+    upload = 'Bad filename.mp4'
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_WATCH_BUCKET)
+    s3c.create_bucket(Bucket=settings.VIDEO_S3_BUCKET)
+    bucket = s3.Bucket(settings.VIDEO_S3_WATCH_BUCKET)
+    bucket.upload_fileobj(io.BytesIO(os.urandom(6250000)), upload)
+    with transaction.atomic():
+        with pytest.raises(VideoFilenameError) as exc:
+            process_watch_file(upload)
+            assert 'No matches found for filename' in str(exc)
```
giocalitri: same here for the string in the exception message

cloudsync/api_test.py
```diff
@@ -130,3 +150,172 @@ def test_process_transcode_results(mocker):
...
+          'PresetId': '1351620000001-000061', 'SegmentDuration': '10.0',
+          'ThumbnailPattern': 'thumbnails/' + new_video.hexkey + '/video_thumbnail_{count}'},
+         {'Key': 'transcoded/' + new_video.hexkey + '/video_1351620000001-000040', 'PresetId': '1351620000001-000040',
+          'SegmentDuration': '10.0'},
+         {'Key': 'transcoded/' + new_video.hexkey + '/video_1351620000001-000020',
+          'PresetId': '1351620000001-000020', 'SegmentDuration': '10.0'}], Playlists=[
+              {'Format': 'HLSv3', 'Name': 'transcoded/' + new_video.hexkey + '/video__index',
+               'OutputKeys': ['transcoded/' + new_video.hexkey + '/video_1351620000001-000061',
+                              'transcoded/' + new_video.hexkey + '/video_1351620000001-000040',
+                              'transcoded/' + new_video.hexkey + '/video_1351620000001-000020']}])
```
giocalitri: nit: please format this json better

cloudsync/api_test.py
```diff
@@ -130,3 +150,172 @@ def test_process_transcode_results(mocker):
...
+    mock_encoder.assert_called_once_with(
+        {'Key': videofile.s3_object_key},
+        [{'Key': 'transcoded/' + new_video.hexkey + '/video_1351620000001-000040',
+          'PresetId': '1351620000001-000040', 'SegmentDuration': '10.0',
+          'ThumbnailPattern': 'thumbnails/' + new_video.hexkey + '/video_thumbnail_{count}'},
+         {'Key': 'transcoded/' + new_video.hexkey + '/video_1351620000001-000020',
+          'PresetId': '1351620000001-000020', 'SegmentDuration': '10.0'}], Playlists=[
+              {'Format': 'HLSv3', 'Name': 'transcoded/' + new_video.hexkey + '/video__index',
+               'OutputKeys': ['transcoded/' + new_video.hexkey + '/video_1351620000001-000040',
+                              'transcoded/' + new_video.hexkey + '/video_1351620000001-000020']}])
```
giocalitri: nit please format this JSON better

cloudsync/api_test.py
```diff
@@ -130,3 +150,172 @@ def test_process_transcode_results(mocker):
...
+                                side_effect=ClientError(error_response=job_result, operation_name='ReadJob'))
+    with pytest.raises(ClientError):
+        transcode_video(new_video, videofile)
+    mock_encoder.assert_called_once_with(
+        {'Key': videofile.s3_object_key},
+        [{'Key': 'transcoded/' + new_video.hexkey + '/video_1351620000001-000020',
+          'PresetId': '1351620000001-000020', 'SegmentDuration': '10.0',
+          'ThumbnailPattern': 'thumbnails/' + new_video.hexkey + '/video_thumbnail_{count}'}], Playlists=[
+              {'Format': 'HLSv3', 'Name': 'transcoded/' + new_video.hexkey + '/video__index',
+               'OutputKeys': ['transcoded/' + new_video.hexkey + '/video_1351620000001-000020']}])
```
giocalitri: nit please format this  JSON better

cloudsync/tasks_test.py
giocalitri: question: No change to these test?

mbertrand: IMO this test isn't very useful without the asserts at the end because they verify that valid videos are processed even if a bad video is in the bucket.  That particular logic is not tested anywhere else.  But if you prefer I can remove them.

cloudsync/api.py
```diff
@@ -1,19 +1,27 @@
...
 
+from cloudsync.exceptions import VideoFilenameError
+from cloudsync.utils import VideoTranscoder
 from ui.constants import VideoStatus
 from ui.encodings import EncodingNames
 from ui.models import (
     VideoFile,
     VideoThumbnail,
-)
+    Collection, Video)
```
mbertrand: OK

cloudsync/tasks_test.py
giocalitri: I understand: leave as is

cloudsync/api.py
```diff
@@ -74,3 +82,128 @@ def refresh_status(video, encode_job=None):
...
+        raise
+    finally:
+        transcoder.create_job_for_object(video)
+        if video.status != VideoStatus.TRANSCODE_FAILED:
+            video.update_status(VideoStatus.TRANSCODING)
+
+
+def process_watch_file(s3key):
+    """
+    Copy the file from the watch bucket to the upload bucket, create model objects, and transcode
```
mbertrand: OK

cloudsync/api_test.py
mbertrand: OK

cloudsync/api_test.py
mbertrand: OK

cloudsync/api_test.py
mbertrand: OK I'll remove it

cloudsync/api_test.py
mbertrand: OK

cloudsync/api_test.py
mbertrand: OK

cloudsync/api_test.py
mbertrand: OK

cloudsync/api_test.py
mbertrand: OK

cloudsync/api.py
mbertrand: same as above

PR #114, opened 2017-08-08
allow video object to have no moira lists

### LINKED ISSUE #48: Video detail page should not make moira lists field required
Otherwise, if moria lists should be required then we should change the upload page to make it required there

pdpinch:

fixes #48 

#### What's this PR do?

Allows a video object to have null moira_lists

Go to the admin end edit a video object. Without this change, you'll get a validation error unless there's a moira list set. After this change, you should get no validation error. 

According to the product description, moira lists should be optional.

### DIFF
--- a/ui/models.py
+++ b/ui/models.py
@@ -51,7 +51,7 @@ class Video(models.Model):
     title = models.CharField(max_length=250, blank=True)
     description = models.TextField(blank=True)
     source_url = models.URLField()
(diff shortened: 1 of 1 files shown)


giocalitri:
I think this is covered by #105 that temporarily removes the moira lists

giocalitri:
I think we can close this unless there is something missing in the current state of the code.

Issue #100, opened 2017-07-31
Lecture Capture: automatically upload from capture machines to S3


pdpinch:
As the technical services manager, I want lecture capture videos to be automatically uploaded from the 1Beyond capture machines into a S3 bucket where they can be processed by OVS. 

We will write a script that will be run periodically (probably daily, after other 1Beyonds tasks are done) on the 1Beyond systems to upload videos to a S3 bucket that OVS will watch and process (see #101) 

### acceptance criteria
- [x]  Notify DevOps on any failure to upload (OpsGenie? Sentry?)
- [ ] Script needs to be configurable by the technical services manager so that necessary info is embedded in the file name, namely:
  - [ ] Collection name = course number + section number
  - [ ] datetime
  - [ ] room number? (useful because it can be used to identify the 1Beyond machine involved, if there is a failure)
- [ ] on success, remove local copies of video files

### questions
- [ ] How many lectures can we store on the 1Beyond before storage becomes a problem?

mbertrand:
`on success, remove local copies of video files`: success == uploaded to S3? Or uploaded to S3 and imported/transcoded by OVS?

MITTechTV:
File name metadata also need course section number (ie, L01, L02, etc) for proper authentication with Moira lists in Stellar as some courses do multiple sections and restrict access to particular section number.

pdpinch:
Thanks Elaine. Let me suggest that we combine course number and section number into the collection name. 

I've updated the description to include this change.

shaidar:
#### File Upload
This is the main piece and based on initial assessment, we will be using [aws s3 sync](http://docs.aws.amazon.com/cli/latest/reference/s3/sync.html) to sync files between the local folder (U-Switch_Records_Done) and the S3 bucket. Haven't decided yet on whether to run this on a scheduled basis or to watch the folder for changes and trigger the script, but if we're sticking with the way things are currently setup, it'll be a scheduled task.

#### Notifications & Logging
- Notifications on failure will be sent to a DevOps slack channel to investigate.
- Log script actions locally on the machine to a log file.

#### Dependencies
- Python 3.6.2
- AWS CLI

#### Challenges
- Secrets management
- Possible manual work to install dependencies on lecture capture machines

#### Testing
- Will setup a VM and verify initial assumptions
- Run script in parallel on one of the hosts to debug and test
- Deploy dependencies and script on remaining hosts

shaidar:
[...]

PR #123, opened 2017-08-09
Automatically upload videos from specified folder to s3 bucket

shaidar:
Fixes [Issue#100](https://github.com/mitodl/odl-video-service/issues/100)

#### What's this PR do?
Script reads config values from a settings ini file and uses the AWS cli to sync a local folder (videos done) to a specified s3 bucket. A local log file is maintained on the individual host and on success or failure of the s3 sync a message is posted to a slack channel with some details.

Ideally this would be tested on a Windows host since that's where it will mainly run on, but technically, it does not matter. Modify ini file to contain your own settings and ensure that the following dependencies are installed:
- logbook
- requests

Scipt requires the settings file as an input file and so to run:
`python s3_sync.py -i s3_sync_settings.ini`

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/s3_sync/s3_sync.py
+++ b/s3_sync/s3_sync.py
@@ -25,7 +25,7 @@ except ImportError as err:
 # Instantiate argparse to get settings_file as argument
 parser = argparse.ArgumentParser(description='.')
 parser.add_argument('-i', dest="settings_file", required=True,
-                    help='ini file containing configs', metavar='FILE')
+                    help='path to ini file containing configs', metavar='FILE')
 args = parser.parse_args()
 settings_file = args.settings_file
 
@@ -43,23 +43,6 @@ logger = Logger(__name__)
 
 # Get Computer name
 computer_name = os.environ['COMPUTERNAME']
-
-
-def verify_settings_file_exists(settings_file):
-    """
-    Verify that settings file exists.
-
-    Args:
-      settings_file (str): Path for the local settings file.
-
-    Returns:
-      If file exists, return None, otherwise exit.
-    """
-    if os.path.exists(settings_file):
-        return
-    else:
-        logger.error("Settings file not found:", settings_file)
-        sys.exit("[-] Settings file not found")
 
 
 def set_environment_variables():
@@ -82,9 +65,7 @@ def verify_local_folder_exists(local_video_records_done_folder):
     Returns:
       If folder exists return None, and if not, logs error and exit.
     """
-    if os.path.exists(local_video_records_done_folder):
-        return
-    else:
+    if not os.path.exists(local_video_records_done_folder):
         logger.error("Local Video Records Done folder not found")
         sys.exit("[-] Local Video Records Done folder not found")
 
@@ -100,9 +81,7 @@ def verify_aws_cli_installed(aws_cli_binary):
       If file exists, return None, else log error and exit.
 
     """
-    if os.path.exists(aws_cli_binary):
-        return
-    else:
+    if not os.path.exists(aws_cli_binary):
         logger.error("Could not find AWS CLI executable")
         sys.exit("[-] Could not find AWS CLI executable")
 
@@ -119,12 +98,12 @@ def verify_s3_bucket_exists(s3_bucket_name):
         objects in bucket otherwise error and exit on any issues trying
         to list objects in bucket.
     """
+    ls_s3_bucket_cmd = 'aws s3 ls {}'.format(s3_bucket_name)
     try:
-        ls_s3_bucket_cmd = 'aws s3 ls {}'.format(s3_bucket_name)
         subprocess.run(ls_s3_bucket_cmd, check=True,
                        stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     except subprocess.SubprocessError:
-        logger.exception("Failed to list specified s3 bucket")
+        logger.exception("Failed to list specified s3 bucket: {}", s3_bucket_name)
         sys.exit("[-] Failed to list specified s3 bucket")
 
 
@@ -136,10 +115,12 @@ def notify_slack_channel(slack_message):
       slack_message (str): message to send to slack
     """
     try:
-        requests.post(os.environ.get('slack_webhook_url'),
-                      json={"text": slack_message,
-                            "username": config['Slack']['bot_username'],
-                            "icon_emoji": config['Slack']['bot_emoji'], })
+        requests.post(
+            os.environ.get('slack_webhook_url'),
+            json={
+                "text": slack_message,
+                "username": config['Slack']['bot_username'],
+                "icon_emoji": config['Slack']['bot_emoji'], })
     except (requests.exceptions.RequestException, NameError) as err:
         logger.warn("Failed to notify slack channel with following error: {}", err)
 
@@ -153,18 +134,18 @@ def sync_local_to_s3(local_video_records_done_folder, s3_bucket_name):
         files ready to be copied to S3.
       s3_bucket_name (str): s3 bucket name
     """
+    s3_sync_cmd = 'aws s3 sync {} "s3://"{}'.format(local_video_records_done_folder, s3_bucket_name)
     try:
-        s3_sync_cmd = 'aws s3 sync {} "s3://"{}'.format(local_video_records_done_folder, s3_bucket_name)
         cmd_output = subprocess.run(s3_sync_cmd, check=True,
                                     stdout=subprocess.PIPE,
                                     stderr=subprocess.PIPE)
-        logger.info("S3 sync successfully ran: {}", cmd_output)
-        notify_slack_channel(f"Sync succeeded on: *{computer_name}* \n `str({cmd_output})`")
-        logger.info("Syncing complete")
     except subprocess.SubprocessError as err:
         logger.exception("Failed to sync local files to s3 bucket")
         notify_slack_channel(f"Sync failed: *{computer_name}* \n `{err}`")
         sys.exit("[-] Failed to sync local files to s3 bucket")
+    logger.info("S3 sync successfully ran: {}", cmd_output)
+    notify_slack_channel(f"Sync succeeded on: *{computer_name}* \n `str({cmd_output})`")
+    logger.info("Syncing complete")
 
 
 def main():
--- a/s3_sync/s3_sync.py
+++ b/s3_sync/s3_sync.py
@@ -9,18 +9,18 @@ Use:
 python s3_sync.py -i <settings_file.ini>
 
 """
+import argparse
+import os
+import subprocess
+import sys
+from configparser import ConfigParser, ExtendedInterpolation
+
 try:
-    import argparse
-    import json
-    import logging
-    import os
     import requests
-    import subprocess
-    import sys
-    from configparser import ConfigParser, ExtendedInterpolation
     from logbook import Logger, RotatingFileHandler
 except ImportError as err:
-    sys.exit(err)
+    print("Failed to import module: ", err)
+    sys.exit("Make sure to pip install requests and logbook")
 
 # Instantiate argparse to get settings_file as argument
 parser = argparse.ArgumentParser(description='.')
@@ -30,14 +30,14 @@ args = parser.parse_args()
 settings_file = args.settings_file
 
 # Read settings_file
-config = ConfigParser(interpolation = ExtendedInterpolation())
+config = ConfigParser(interpolation=ExtendedInterpolation())
 config.read(settings_file)
 
 # Configure logbook logging
 logger = RotatingFileHandler(config['Logs']['logfile'],
-                             max_size = int(config['Logs']['max_size']),
-                             backup_count = int(config['Logs']['backup_count']),
-                             level = int(config['Logs']['level']))
+                             max_size=int(config['Logs']['max_size']),
+                             backup_count=int(config['Logs']['backup_count']),
+                             level=int(config['Logs']['level']))
 logger.push_application()
 logger = Logger(__name__)
 
@@ -154,7 +154,7 @@ def sync_local_to_s3(local_video_records_done_folder, s3_bucket_name):
       s3_bucket_name (str): s3 bucket name
     """
     try:
-        s3_sync_cmd = 'aws s3 sync {local_video_records_done_folder} "s3://"{s3_bucket_name}'.format(local_video_records_done_folder=local_video_records_done_folder, s3_bucket_name=s3_bucket_name)
+        s3_sync_cmd = 'aws s3 sync {} "s3://"{}'.format(local_video_records_done_folder, s3_bucket_name)
         cmd_output = subprocess.run(s3_sync_cmd, check=True,
                                     stdout=subprocess.PIPE,
                                     stderr=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


s3_sync/s3_sync.py
```diff
@@ -0,0 +1,158 @@
...
+
+	Returns:
+	  list: if connection established and bucket found, return list of
+	  	objects in bucket otherwise error and exit on any issues trying
+	  	to list objects in bucket.
+	"""
+	try:
+		ls_s3_bucket_cmd = 'aws s3 ls {}'.format(s3_bucket_name)
+		subprocess.run(ls_s3_bucket_cmd, shell=True, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+		return
```
blarghmatey: This return isn't necessary as the function will automatically exit (with an implicit `return None`) if there is no exception thrown.

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,158 @@
...
+	Send notification to Slack Channel
+
+	Args:
+	  slack_message (str): message to send to slack
+	"""
+	try:
+		resp = requests.post(os.environ.get('slack_webhook_url'), json={"text": slack_message,
+			"username": config['Slack']['bot_username'],
+			"icon_emoji": config['Slack']['bot_emoji'],})
+		return
```
blarghmatey: Same. Also, if you aren't going to use the value from `resp` then you should avoid assigning it to avoid confusion.

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,158 @@
...
+	   	ready to be copied to S3.
+	  s3_bucket_name (str): s3 bucket name
+	"""
+	try:
+		s3_sync_cmd = 'aws s3 sync {local_video_records_done_folder} "s3://"{s3_bucket_name}'.format(
+			local_video_records_done_folder=local_video_records_done_folder, s3_bucket_name=s3_bucket_name)
+		cmd_output = subprocess.run(s3_sync_cmd, shell=True, check=True, stdout=subprocess.PIPE,
+			stderr=subprocess.PIPE)
+		logger.info("S3 sync successfully ran: {}", cmd_output)
+		notify_slack_channel("Sync succeeded on: " + "*" + computer_name + "*" + "\n" + "`" + str(cmd_output) + "`")
```
blarghmatey: Use the `str.format` function instead of the `+` operator.

blarghmatey: If we can enforce a Python 3.6 runtime then you can use fstrings (https://docs.python.org/3/reference/lexical_analysis.html#f-strings)

s3_sync/s3_sync.py
```diff
@@ -33,126 +34,156 @@
...
-	if os.path.exists(local_video_records_done_folder):
-		return
-	else:
-		logger.exception("Local Video Records Done folder not found")
-		sys.exit("[-] Local Video Records Done folder not found")
+    """
+    if os.path.exists(local_video_records_done_folder):
+        return
+    else:
+        logger.exception("Local Video Records Done folder not found")
```
blarghmatey: This should be `logger.error`. `exception` is only really relevant when an exception has been raised and there is a traceback to be captured.

s3_sync/s3_sync.py
```diff
@@ -33,126 +34,156 @@
...
+        return
+    else:
+        try:
+            r = requests.get(config.get('AWS','CLI', fallback='https://s3.amazonaws.com/aws-cli/AWSCLI64.msi'), stream=True)
+            with open('aws_cli.msi', 'wb') as fd:
+                for chunk in r.iter_content(chunk_size=128):
+                    fd.write(chunk)
+            subprocess.run('msiexec /i aws_cli.msi /qn', check=True,
+                           stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        except:
```
blarghmatey: You should almost never use a bare except clause because it can cause maintenance difficulties when an error is captured without you meaning to do so. If you have an understanding of the types of exceptions that may be raised in this context then you should capture those and let the others through.

shaidar:
One of the disadvantages of adding a case to install AWS CLI in case it's not found is the need to run the script as admin on the host since it would need those escalated rights to install the msi.

s3_sync/s3_sync.py
```diff
@@ -33,126 +34,156 @@
...
+    """
+    try:
+        s3_sync_cmd = 'aws s3 sync {local_video_records_done_folder} "s3://"{s3_bucket_name}'.format(local_video_records_done_folder=local_video_records_done_folder, s3_bucket_name=s3_bucket_name)
+        cmd_output = subprocess.run(s3_sync_cmd, check=True,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
+        logger.info("S3 sync successfully ran: {}", cmd_output)
+        notify_slack_channel(f"Sync succeeded on: *{computer_name}* \n `str({cmd_output})`")
+        logger.info("Syncing complete")
+        return
```
blarghmatey: Extraneous return statement

s3_sync/s3_sync.py
```diff
@@ -33,126 +34,156 @@
...
+      slack_message (str): message to send to slack
+    """
+    try:
+        requests.post(os.environ.get('slack_webhook_url'),
+                      json={"text": slack_message,
+                            "username": config['Slack']['bot_username'],
+                            "icon_emoji": config['Slack']['bot_emoji'], })
+    except (requests.exceptions.RequestException, NameError) as err:
+        logger.warn("Failed to notify slack channel with following error: {}", err)
+        return
```
blarghmatey: Extraneous return statement

s3_sync/s3_sync.py
shaidar: I thought if you do logger.error it would not include a stack trace of the error in the log file or is that not accurate?

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,184 @@
...
+    Send notification to Slack Channel
+
+    Args:
+      slack_message (str): message to send to slack
+    """
+    try:
+        requests.post(os.environ.get('slack_webhook_url'),
+                      json={"text": slack_message,
+                            "username": config['Slack']['bot_username'],
+                            "icon_emoji": config['Slack']['bot_emoji'], })
```
giocalitri: nit: try to use this indenting approach
```
requests.post(
    os.environ.get('slack_webhook_url'),
    json={
        "text": slack_message,
        "username": config['Slack']['bot_username'],
        "icon_emoji": config['Slack']['bot_emoji'], 
    }
)
```

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,184 @@
...
+      s3_bucket_name (str): s3 bucket name
+    """
+    try:
+        s3_sync_cmd = 'aws s3 sync {} "s3://"{}'.format(local_video_records_done_folder, s3_bucket_name)
+        cmd_output = subprocess.run(s3_sync_cmd, check=True,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
+        logger.info("S3 sync successfully ran: {}", cmd_output)
+        notify_slack_channel(f"Sync succeeded on: *{computer_name}* \n `str({cmd_output})`")
+        logger.info("Syncing complete")
```
giocalitri: please keep only the line that might fail in the try/except statement

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,184 @@
...
+    Args:
+      local_video_records_done_folder (str): local folder containing video
+        files ready to be copied to S3.
+      s3_bucket_name (str): s3 bucket name
+    """
+    try:
+        s3_sync_cmd = 'aws s3 sync {} "s3://"{}'.format(local_video_records_done_folder, s3_bucket_name)
+        cmd_output = subprocess.run(s3_sync_cmd, check=True,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE)
```
giocalitri: question: cannot this fail?

giocalitri:
Also, you still have pylint errors

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Returns:
+      list: if connection established and bucket found, return list of
+        objects in bucket otherwise error and exit on any issues trying
+        to list objects in bucket.
+    """
+    try:
+        ls_s3_bucket_cmd = 'aws s3 ls {}'.format(s3_bucket_name)
+        subprocess.run(ls_s3_bucket_cmd, check=True,
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
shaidar: Ideally we would use boto, but at this point it does not seem to have that functionality:
https://github.com/boto/boto3/issues/358

giocalitri: sorry: comment from before we discussed the issue on Friday

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+# Configure logbook logging
+logger = RotatingFileHandler(config['Logs']['logfile'],
+                             max_size = int(config['Logs']['max_size']),
+                             backup_count = int(config['Logs']['backup_count']),
+                             level = int(config['Logs']['level']))
+logger.push_application()
+logger = Logger(__name__)
+
+# Get Computer name
+computer_name = os.environ['COMPUTERNAME']
```
shaidar: This is always set as far as I know on a Windows box.

s3_sync/s3_sync.py
shaidar: It can fail and that's why I put it in a try/except block. Is there anything else I need to do?

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+    import requests
+    import subprocess
+    import sys
+    from configparser import ConfigParser, ExtendedInterpolation
+    from logbook import Logger, RotatingFileHandler
+except ImportError as err:
+    sys.exit(err)
+
+# Instantiate argparse to get settings_file as argument
+parser = argparse.ArgumentParser(description='.')
```
shaidar: Think it's a bit of an over explanation and feels like something you'd write in a CS class.

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+parser.add_argument('-i', dest="settings_file", required=True,
+                    help='ini file containing configs', metavar='FILE')
```
shaidar: Changed

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+args = parser.parse_args()
+settings_file = args.settings_file
+
+# Read settings_file
+config = ConfigParser(interpolation = ExtendedInterpolation())
+config.read(settings_file)
```
shaidar: If it doesn't exist, it would fail to read. Don't currently have any checks on whether it's malformed

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+                             backup_count = int(config['Logs']['backup_count']),
+                             level = int(config['Logs']['level']))
+logger.push_application()
+logger = Logger(__name__)
+
+# Get Computer name
+computer_name = os.environ['COMPUTERNAME']
+
+
+def verify_settings_file_exists(settings_file):
```
shaidar: Removed function

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Verify that settings file exists.
+
+    Args:
+      settings_file (str): Path for the local settings file.
+
+    Returns:
+      If file exists, return None, otherwise exit.
```
shaidar: Removed function

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    if os.path.exists(settings_file):
+        return
+    else:
+        logger.error("Settings file not found:", settings_file)
+        sys.exit("[-] Settings file not found")
```
shaidar: Changed

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+        records done folder.
+
+    Returns:
+      If folder exists return None, and if not, logs error and exit.
+    """
+    if os.path.exists(local_video_records_done_folder):
+        return
+    else:
+        logger.error("Local Video Records Done folder not found")
+        sys.exit("[-] Local Video Records Done folder not found")
```
shaidar: Changed

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Returns:
+      If file exists, return None, else log error and exit.
+
+    """
+    if os.path.exists(aws_cli_binary):
+        return
+    else:
+        logger.error("Could not find AWS CLI executable")
+        sys.exit("[-] Could not find AWS CLI executable")
```
shaidar: Changed

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+    Args:
+      s3_bucket_name (str): The s3 bucket name
+
+    Returns:
+      list: if connection established and bucket found, return list of
+        objects in bucket otherwise error and exit on any issues trying
+        to list objects in bucket.
+    """
+    try:
+        ls_s3_bucket_cmd = 'aws s3 ls {}'.format(s3_bucket_name)
```
shaidar: Moved

s3_sync/s3_sync.py
```diff
@@ -0,0 +1,179 @@
...
+        subprocess.run(ls_s3_bucket_cmd, check=True,
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except subprocess.SubprocessError:
+        logger.exception("Failed to list specified s3 bucket")
+        sys.exit("[-] Failed to list specified s3 bucket")
```
shaidar: Added bucket name

s3_sync/s3_sync.py
shaidar: Changed

s3_sync/s3_sync.py
shaidar: Changed

s3_sync/s3_sync.py
giocalitri: do we need any?

s3_sync/s3_sync.py
giocalitri: sorry, it was not clear before: now with the new code it is better

s3_sync/s3_sync.py
giocalitri: ok

s3_sync/s3_sync.py
giocalitri: ok

s3_sync/s3_sync.py
giocalitri: ok

s3_sync/s3_sync.py
giocalitri: ok

s3_sync/s3_sync.py
giocalitri: ok

s3_sync/s3_sync.py
giocalitri: ok

s3_sync/s3_sync.py
shaidar: Put filed reading in try/except block

giocalitri:
If we do not plan to write tests I think this is good enough to go:

pdpinch:
@shaidar can you open an issue for writing unit tests? I'm OK with putting them off, but I would like them eventually. 

Also, you're missing the "fixes" keyword in your PR description.

PR #147, opened 2017-08-16
Store x509 certificate/key as environment variables

### LINKED ISSUE #142: Handle x509 certificates using environment variables
In deployment environments like Heroku it's inconvenient to store certificates as a file on disk.  Yet the `requests` python package used in libraries such as mit-moira and pylmod expects certificates to be passed as filepaths; in particular, `requests.session.cert()`.

Our approach in other apps has been to put the certificate into a environment variable, write the variable to a file on startup, and then pass the file to a library like mit-moira.  In this case, it is recommended to write to a temporary file on startup, that will be overwritten each time the app is restarted.

mbertrand:
Fixes #142 
Dependent on https://github.com/mitodl/mit-moira/pull/3

#### What's this PR do?
Gets the contents of an x509 certificate and private key as environmental variables, and saves them to files on startup.  Uncomments `ui.utils.get_moira_client` functionality.

- In your .env file, set `MIT_WS_CERTIFICATE` to the contents of your x509 certificate file and `MIT_WS_PRIVATE_KEY` to the contents of your x509 private key (I can provide a sample).
- `docker-compose build; docker-compose run web python manage.py shell`
- ```
  from ui.models import MoiraList
  odl_list = MoiraList(name='odl-engineering')
  odl_list.members()
  ```
- The above should return a list of all users in the odl-engineering list, including Anna's email, with no duplicates (assuming https://github.com/mitodl/mit-moira/pull/3 has been merged)


ui/utils_test.py
```diff
@@ -0,0 +1,54 @@
...
+def test_get_moira_client_nocert(mocker):
+    """Test that the correct error message is returned if no x509 certificate & key files exist"""
+    utils.write_x509_files()
+    os.remove(utils.MIT_WS_CERTIFICATE_FILE)
+    mock_moira = mocker.patch('ui.utils.Moira')
+    with pytest.raises(RuntimeError) as err:
+        utils.get_moira_client()
+        assert utils.MIT_WS_CERTIFICATE_FILE in str(err)
+        assert utils.MIT_WS_PRIVATE_KEY_FILE not in str(err)
+    assert not mock_moira.called
```
gsidebo: I think we should change the approach here.

1. If we can't avoid involving the file system for these tests, we should specify new file paths so that files that exist in the default path aren't altered during the test run. 
1. We should just create empty files at those paths as necessary instead of invoking `write_x509_files`. `get_moira_client` just depends on those files existing. Invoking `write_x509_files` to satisfy that requirement isn't as explicit about what `get_moira_client` expects, and it also unnecessarily extends the scope of this test
1. We can/probably should parametrize the tests where one or both files don't exist. All 3 of them specify some file that shouldn't exist, then check that the missing file path appears in the exception message. I don't think it's necessary to check for an existing file path to be absent from the error message
1. `mock_moira` is created in every test case. We should create a fixture for that patch

ui/apps.py
```diff
@@ -21,3 +21,7 @@ def ready(self):
         if missing_settings:
             raise ImproperlyConfigured(
                 'The following settings are missing: {}'.format(', '.join(missing_settings)))
+
+        # write the x509 certification & key to files
+        from ui.utils import write_x509_files
+        write_x509_files()
```
mbertrand: I talked with Giovanni about checking for existing files first before writing them; decided instead it might be better to just overwrite them on startup in case the x509 keys/certs change.

ui/models.py
```diff
@@ -29,7 +29,7 @@ class MoiraList(models.Model):
     def members(self):
         """Members"""
         moira = utils.get_moira_client()
-        return moira.list_members(self.name)
+        return set(moira.list_members(self.name, type=''))
```
mbertrand: Sure

ui/utils.py
```diff
@@ -13,43 +14,37 @@
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import padding
 from botocore.signers import CloudFrontSigner
-# from mit_moira import Moira
+from mit_moira import Moira
+
+
+# x509 filenames
+MIT_WS_CERTIFICATE_FILE = os.path.join(settings.BASE_DIR, settings.STATIC_ROOT, 'mit_x509.cert')
+MIT_WS_PRIVATE_KEY_FILE = os.path.join(settings.BASE_DIR, settings.STATIC_ROOT, 'mit_x509.key')
```
mbertrand: Agreed.

ui/utils.py
```diff
@@ -13,43 +14,37 @@
...
+    if not cert_file_exists:
+        msg = "Missing required secret: {cert}".format(
+            cert=MIT_WS_CERTIFICATE_FILE,
+        )
+        raise RuntimeError(msg)
+    if not key_file_exists:
+        msg = "Missing required secret: {key}".format(
+            key=MIT_WS_PRIVATE_KEY_FILE,
+        )
+        raise RuntimeError(msg)
```
mbertrand: Thanks, will do

ui/utils_test.py
mbertrand: Will do

ui/apps.py
gsidebo: I'm asking why we are counting on the app to write these files at all. Why can't we just expect the two files to exist with the correct key/cert? If the keys/certs change, you just change the file in place, and we don't have to deal with the awkwardness of storing those large secret values with `\n` sprinkled throughout in the `.env` file

ui/models_test.py
```diff
@@ -45,6 +45,12 @@ def videofile():
     return VideoFileFactory()
 
 
+@pytest.fixture
+def moiralist():
+    """Fixture to create a moira list"""
+    return MoiraListFactory()
```
gsidebo: OK, I'll leave to your discretion. Generally I use fixtures for test data when there's a 'tree' of specific hierarchical data that I need, and/or there's a model object I need with some really specific properties. That's just a broad guiding principle, though. If you think we're getting use out of this fixture-izing, I'm for it 👍

ui/apps.py
mbertrand: It has something to do with deploying to heroku. Originally these keys were expected to already exist as files along with some others (like the cloudfront key).  But it wasn't possible to deploy these files to heroku so they were all changed to environment variables.  Most could be read directly without the need for a file to be written, but the x509 keys don't fall in that category because the mit-moira library uses `zeep` which in turn relies on `requests` which [only reads certificates from a file](http://docs.python-requests.org/en/master/user/advanced/#client-side-certificates).

mbertrand: https://github.com/mitodl/odl-video-service/issues/78

gsidebo: Aaaaah I see. Thanks for clarifying! 👍

ui/utils_test.py
```diff
@@ -0,0 +1,58 @@
...
+    with pytest.raises(RuntimeError) as err:
+        utils.get_moira_client()
+        if not key_exists:
+            assert settings.MIT_WS_PRIVATE_KEY_FILE in str(err)
+        if not cert_exists:
+            assert settings.MIT_WS_CERTIFICATE_FILE in str(err)
+    if not key_exists or not cert_exists:
+        assert not mock_moira.called
+    else:
+        assert mock_moira.called_once_with(settings.MIT_WS_CERTIFICATE_FILE, settings.MIT_WS_PRIVATE_KEY_FILE)
```
gsidebo: In general I think it's a good idea to avoid helper functions in tests that use a lot of boolean flags, make a lot of assertions, and try to cover a bunch of different scenarios. It tends to obscure the purpose of the tests. When I mentioned parametrization, I meant something like this:

```python
@pytest.mark.parametrize("key_file, cert_file", [
    (NamedTemporaryFile(), None),
    (None, NamedTemporaryFile()),
    (None, None),
])
def test_get_moira_client_missing_secrets(mock_moira, settings, key_file, cert_file):
    settings.MIT_WS_PRIVATE_KEY_FILE = 'bad/file/path' if not key_file else key_file.name
    settings.MIT_WS_CERTIFICATE_FILE = 'bad/file/path' if not cert_file else cert_file.name
    with pytest.raises(RuntimeError) as err:
        utils.get_moira_client()
        assert not mock_moira.called
        if key_file is None:
            assert settings.MIT_WS_PRIVATE_KEY_FILE in str(err)
        if cert_file is None:
            assert settings.MIT_WS_CERTIFICATE_FILE in str(err)


def test_get_moira_client_success(mock_moira, settings):
    """Test that a client is returned from get_moira_client"""
    tempfile1, tempfile2 = (NamedTemporaryFile(), NamedTemporaryFile())
    settings.MIT_WS_PRIVATE_KEY_FILE = tempfile1.name
    settings.MIT_WS_CERTIFICATE_FILE = tempfile2.name
    utils.get_moira_client()
    assert mock_moira.called_once_with(settings.MIT_WS_CERTIFICATE_FILE, settings.MIT_WS_PRIVATE_KEY_FILE)
```

mbertrand: OK

gsidebo:
Code looks great! I'll test functionality again shortly

odl_video/settings.py
```diff
@@ -475,6 +475,15 @@ def get_all_config_keys():
 USWITCH_URL = get_string('USWITCH_URL', '')
 LECTURE_CAPTURE_USER = get_string('LECTURE_CAPTURE_USER', '')
 
+# x509 certificate and private key content as environment variables - required
+MIT_WS_CERTIFICATE = get_key('MIT_WS_CERTIFICATE', '')
+MIT_WS_PRIVATE_KEY = get_key('MIT_WS_PRIVATE_KEY', '')
```
gsidebo: It looks like these two variables are assigned at lines 405 and 406 as well. Can you get rid of those?

Issue #135, opened 2017-08-14
Front-end for video detail page


pdpinch:
As discussed, this will be an independent page/template, not a single-page-app view. 

Mock-up at https://projects.invisionapp.com/boards/4C3980T7H3NM9/

PR #158, opened 2017-08-21
Permissions with moira lists

### LINKED ISSUE #52: As a channel admin, I would like to view and edit a collection that I have permission to upload to
The channel admin should be able to view the list of videos in any channel they have admin permissions for, and upload additional videos to it. 

~~Depends on mitodl/mit-moira#1~~

Depends on #142

### LINKED ISSUE #53: As a viewer, I would like to be able to view a private video that I have permission to view
When viewing videos, access to the video stream should be controlled by moira permissions applied to the video channel. 

Depends on mitodl/mit-moira#1 

We need a design for what happens when the user doesn't have permission.

### LINKED ISSUE #146: Support for moira lists that contain e-mail addresses
Some moira lists contain e-mail addresses as well as kerberos user IDs. It would be nice for these to work the same with view and admin permissions as kerberos ids, but the mit-moira library doesn't currently return the e-mail addresses. 

See https://github.com/mitodl/mit-moira/issues/4

### LINKED ISSUE #148: Modify Collection model: rename moira_lists to view_lists
part of #53

Collections need to manage two sets of permissions, who can view and who can admin the collection. To manage those separately, we will rename the moira_lists fields to view_lists.

### LINKED ISSUE #149: add admin_lists to Collection model
part of #52 

Collections need to manage two sets of permissions, who can view and who can admin the collection. To manage those separately, we will add an admin_lists field.

### LINKED ISSUE #151: create permissions class for collection admins
part of #52

### LINKED ISSUE #152: As a user I should be able to see a list of all collections I have admin access for
... not just the ones that I own.

part of #52 

Currently, the queryset that we use to get a list of collections that can be administrated filters on collection owners. It should instead filter on moira list membership. 

It might help to use the moira web service to get a list of all moira lists that the current user is a member of.

### LINKED ISSUE #153: Superusers should be able to use all REST APIs
part of #52 and #53

mbertrand:
Fixes #52 
Fixes #53 
Fixes #146
Fixes #148 
Fixes #149 
Fixes #151 
Fixes #152 
Fixes #153

#### What's this PR do?
Adds additional permission classes and functions to control who can view or edit collections and videos.

- Log in as the superuser
- Create 3 collections and add 1 or more videos to each one
- Go to the admin console
- For collection 1, add 'odl-engineering' to 'VIEW MOIRA LISTS'
- For collection 2, add 'odl-engineering' to 'ADMIN MOIRA LISTS'
- Create a new user who is not a superuser, and fill in the email address field with your MIT email address
- In a private browser window, log in as the new user. 
- You should see only collection 2 in your list of collections.
- For each link you visit, try the corresponding view in the API and make sure the permissions work roughly the same (what collections/videos you can view or edit don't change).
- Go to collection 2's detail page for collection 2, you should be able to see it.  There's no edit form here right now but on the API page you should be able to make edits.
- Go to collection 2's video detail page, you should be able to see & edit it.
- Try to visit the collection 1 detail page (manually enter the URL), you should get a 403 Forbidden error (or 'Not found' on the API page) 
- Try to view the video detail page for collection 1's video, you should be able to view it.  but if you try to edit it the changes shouldn't be saved. 
- Try to visit the collection detail and video detail pages for collection 3.  You should be denied access to both. 
- Create a new collection. Now you should see the new collection and collection 2 in your list.  Add a video to the new collection.
- Go back to the other window where you're logged in as a superuser. Refresh the collections page.  You should see collections 1-3 and the new collection you created as the other user.  Make sure you can view and edit that other user's collection and its video too.


ui/views_test.py
```diff
@@ -143,14 +145,17 @@ def test_mosaic_view(logged_in_client, settings):  # pylint: disable=redefined-o
...
 
 
-def test_upload_dropbox_videos_authentication(logged_in_apiclient):
+def test_upload_dropbox_videos_authentication(mock_moira_client, logged_in_apiclient):
     """
-    Tests that only authenticated users can call UploadVideosFromDropbox
+    Tests that only authenticated users with collection admin permissions can call UploadVideosFromDropbox
     """
     client, user = logged_in_apiclient
     client.logout()
```
gsidebo: I know this isn't part of this PR, but why are we using a logged-in API client, then immediately logging out? Why not just use the not-logged-in `apiclient`?

ui/views_test.py
```diff
@@ -297,13 +304,17 @@ def test_collection_viewset_create(logged_in_apiclient):
...
+        'admin_lists': []
     }
     assert client.post(url, input_data, format='json').status_code == status.HTTP_403_FORBIDDEN
 
     # or if does not specify the owner id
     input_data = {
-        'title': 'foo title'
+        'title': 'foo title',
+        'view_lists': [],
+        'admin_lists': []
```
gsidebo: I'm not really sure about the value of testing that a POST request with a missing id will be forbidden. I would recommend we remove it entirely unless you think it has real value

ui/views_test.py
```diff
@@ -330,19 +343,23 @@ def test_collection_viewset_create_as_superuser(logged_in_apiclient):
...
         'title': 'foo title'
     }
     assert client.post(url, input_data, format='json').status_code == status.HTTP_201_CREATED
 
     # if does not specify the owner id it gets a different error
     input_data = {
-        'title': 'foo title'
+        'title': 'foo title',
+        'view_lists': [],
+        'admin_lists': []
```
gsidebo: OK now I'm confused. Why is there a different error code in this case vs. the test case above?

ui/permissions.py
```diff
@@ -1,46 +1,182 @@
...
+    Otherwise use the entire email address as the username.
+
+    Args:
+        user (django.contrib.auth.User): the Django user to return a Moira user for.
+
+    Returns:
+        A namedtuple containing username and type
+
+    """
+    MoiraUser = namedtuple('MoiraUser', 'username type')
```
mbertrand: Will do

ui/permissions.py
```diff
@@ -1,46 +1,182 @@
...
+
+    Args:
+        user (django.contrib.auth.User): the Django user.
+
+    Returns:
+        A list of collections the user has admin access to.
+
+    """
+    return Collection.objects.filter(
+        Q(admin_lists__in=MoiraList.objects.filter(name__in=user_moira_lists(user))) | Q(owner=user))
```
mbertrand: OK

ui/permissions.py
```diff
@@ -1,46 +1,182 @@
...
+        if not lists:
+            lists = obj.admin_lists.all()
+            if not lists:
+                return False
+        common_lists = list(set(user_moira_lists(request.user)).intersection([mlist.name for mlist in lists]))
+        if common_lists:
+            return True
+        return has_admin_permission(obj, request)
+    else:
+        return has_admin_permission(obj, request)
```
mbertrand: OK

ui/permissions.py
```diff
@@ -1,46 +1,182 @@
...
+    """
+    if request.user == obj.owner or request.user.is_superuser:
+        return True
+    lists = obj.admin_lists.all()
+    if not lists:
+        return False
+    common_lists = list(set(user_moira_lists(request.user)).intersection([mlist.name for mlist in lists]))
+    if common_lists:
+        return True
+    return False
```
mbertrand: OK

ui/views_test.py
```diff
@@ -370,10 +391,118 @@ def test_collection_viewset_detail_as_superuser(logged_in_apiclient):
...
+    client, _ = logged_in_apiclient
+    video = VideoFactory()
+    collection = video.collection
+    moira_list = factories.MoiraListFactory()
+    collection.admin_lists = [moira_list]
+    collection.save()
+    mock_moira_client.return_value.user_lists.return_value = ['other_list']
+    url = reverse('video-detail', kwargs={'video_key': video.hexkey})
+    result = client.get(url)
+    assert result.status_code == status.HTTP_403_FORBIDDEN
```
gsidebo: _NOTE: Sorry, I wrote a comment during the last round of reviews that didn't save for some reason. Repeating it here_

I know it's sort of the pattern in the file, but in these new test cases there's a lot of repetition of the sort that pytest is built to address. One thing we can/should do is fixture-ize commonly-needed scenarios. For example, a lot of these cases need a collection that a user will have admin or view permissions for. Instead of repeating that a bunch of times and obscuring the intent of the test, we can turn this idea into fixtures doing something like this:

```python
from types import SimpleNamespace

@pytest.fixture
def collection_data():
    video = VideoFactory()
    return SimpleNamespace(
        collection=video.collection,
        video=video
    )

@pytest.fixture
def admin_col_data(collection_data, mock_moira_client):
    moira_list = factories.MoiraListFactory()
    collection.admin_lists = [moira_list]
    collection.save()
    mock_moira_client.return_value.user_lists.return_value = [moira_list.name]
    return collection_data

@pytest.fixture
def viewable_col_data(collection_data, mock_moira_client):
    moira_list = factories.MoiraListFactory()
    collection.view_lists = [moira_list]
    collection.save()
    mock_moira_client.return_value.user_lists.return_value = [moira_list.name]
    return collection_data
    
def test_collection_detail_view_permission(logged_in_apiclient, viewable_col_data):
    url = reverse('video-detail', kwargs={'video_key': viewable_col_data.video.hexkey})
    result = client.get(url)
    assert result.status_code == status.HTTP_200_OK
    assert json.loads(result.context_data['js_settings_json'])['editable'] is False
    
def test_collection_detail_admin_permission(logged_in_apiclient, admin_col_data):
    url = reverse('video-detail', kwargs={'video_key': admin_col_data.video.hexkey})
    result = client.get(url)
    assert result.status_code == status.HTTP_200_OK
    assert json.loads(result.context_data['js_settings_json'])['editable'] is True

# etc, etc...
```
In that example, any test case that should involve a scenario where the user has admin permissions for the given collection would just need to add `admin_col_data` to the list of arguments.

[This test file in micromasters](https://github.com/mitodl/micromasters/blob/master/financialaid/views_pytest_test.py) has some of these patterns if you're interested. Using `SimpleNamespace` instead of multiple return values is particularly helpful and readable.

ui/permissions.py
```diff
@@ -68,4 +204,7 @@ def has_permission(self, request, view):
             uuid.UUID(collection_key)
         except ValueError:
             raise ValidationError('wrong UUID format for {}'.format(collection_key))
-        return Collection.objects.filter(owner=request.user, key=collection_key).exists()
+        collection = Collection.objects.filter(key=collection_key)
+        if collection:
+            return has_admin_permission(collection.first(), request)
+        return False
```
mbertrand: The if statement above it will prevent this block from being called if there are no collections (equivalent to False):
```
matches = Collection.objects.filter(title='hfhhf')
if matches:
    print('got a collection')
else:
    print('Nothing there')
```
`> Nothing there`

ui/serializers.py
```diff
@@ -128,6 +139,8 @@ class Meta:
             'title',
             'description',
             'owner',
+            'view_lists',
+            'admin_lists'
```
mbertrand: nope, will fix

ui/views.py
```diff
@@ -69,7 +70,7 @@ class CollectionList(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context['collection_list'] = Collection.for_owner(self.request.user)
+        context['collection_list'] = CollectionViewSet(request=self.request).get_queryset()
```
mbertrand: OK

ui/views_test.py
mbertrand: Since it wasn't added as part of the PR I'd prefer to leave this test for now and wait for input from Giovanni.  I think this is to just make sure that each condition in the CollectionViewSet's if statement is being tested:
`if not request.user.is_superuser and (owner_id is None or int(owner_id) != request.user.id)`

ui/views_test.py
mbertrand: Previous: a user doesn't have permission to create a collection belonging to someone else or nobody at all - so the permission issue raises an exception first.  This one: a superuser can create a collection belonging to anyone, so no permission error is raised, which eventually leads to the next error because owner_id is required in the database.

ui/views_test.py
mbertrand: will do

ui/permissions.py
gsidebo: The check for `collection is not None` also prevents `has_admin_permission` from being run.

```python
import time
def expensive_function():
    time.sleep(5)
    return True
collections = None
# expensive_function doesn't run b/c collections is None
collections is not None and expensive_function()
```

mbertrand: `has_admin_permission` won't be run if the queryset result is empty:
```
matches = Collection.objects.filter(title='hfhhf') . # Returns empty <QuerySet []>
if matches:
    print('got a queryset containing 1+ collections, will run has_admin_permission() and collection.first()')
    has_admin_permission(collection.first(), request)
else:
    print('No collections in queryset, not running has_admin_permission() or collection.first()')
```
`> No collections in queryset, not running has_admin_permission() or collection.first()`

ui/views_test.py
mbertrand: Maybe to test that logout works properly? Not sure.

ui/permissions.py
gsidebo: Right. I'm saying that we can write the same thing more succinctly (no `if` clause needed) with no extra queries/logic involved. Unless I'm mistaken, these two bits of code are functionally identical with no computational difference:

```python
collection = Collection.objects.filter(key=collection_key).first()
return collection is not None and has_admin_permission(collection, request)
```

```python
collection = Collection.objects.filter(key=collection_key)
if collection:
    return has_admin_permission(collection.first(), request)
return False
```

ui/views_test.py
```diff
@@ -370,10 +403,104 @@ def test_collection_viewset_detail_as_superuser(logged_in_apiclient):
...
+    """
+    Tests that a user cannot view a video if user is not a member of collection's lists
+    """
+    client, _ = logged_in_apiclient
+    user_list_data.collection.view_lists = [user_list_data.moira_list]
+    user_list_data.collection.save()
+    mock_moira_client.return_value.user_lists.return_value = ['other_list']
+    url = reverse('video-detail', kwargs={'video_key': user_list_data.video.hexkey})
+    result = client.get(url)
+    assert result.status_code == status.HTTP_403_FORBIDDEN
```
gsidebo: What do you think of adding fixtures for `admin_user_list_data` and `viewable_user_list_data` (or some similar names)? These would capture the concept of collections that the user has admin or view permissions for. For example, lines 501-503 are just boilerplate to ensure that a user has view permissions for a collection, and it's repeated a few times in other test cases in this file (and quite a lot more in `ui/permissions_test.py`). We can capture that in a fixture:

```python
def viewable_user_list_data(user_list_data, mock_moira_client):
    user_list_data.collection.view_lists = [user_list_data.moira_list]
    user_list_data.collection.save()
    mock_moira_client.return_value.user_lists.return_value = ['other_list']
    return user_list_data
```

This test case could then replace `user_list_data` in the args with `viewable_user_list_data` (same deal for collections that we want the user to have admin permissions for).

ui/permissions_test.py
```diff
@@ -32,28 +59,95 @@ def collection():
...
+
+
+@pytest.fixture
+def request_data_su():
+    """
+    Fixture for tests requiring a superuser, request, and view
+    """
+    user = factories.UserFactory()
+    user.is_superuser = True
+    user.save()
```
gsidebo: Pretty sure we can just do `user = factories.UserFactory(is_superuser=True)` without the 2 lines below it

ui/permissions_test.py
```diff
@@ -32,28 +59,95 @@ def collection():
...
     """
-    other_user = factories.UserFactory()
-    request = Mock(
-        user=other_user
-    )
-    view = Mock(kwargs={'user': other_user.username})
-    assert collection_owner_permission.has_object_permission(request, view, collection) is False
+    assert collection_owner_permission.has_object_permission(request_data.request,
+                                                             request_data.view,
+                                                             collection) is False
```
gsidebo: I think we need to be a little more explicit about this. There's nothing about the fixture names or the actual test logic that clearly indicates that the collection has an owner that is different from the request user. I think we should create a new collection in the body of this test case (`collection = Collection(owner__user__username='other_user')` or something) and then test `has_object_permission`

ui/views_test.py
gsidebo: OK. If we remove this, we can do it in a separate PR

ui/templates/ui/collection_list.html
```diff
@@ -40,7 +40,10 @@
       title: form.querySelector("[name=title]").value,
       description: form.querySelector("[name=description]").value,
       owner: form.querySelector("[name=owner]").value,
-    }
+      view_lists: [],
+      admin_lists: []
+    };
+    console.log(data);
```
gsidebo: console.log

ui/permissions.py
```diff
@@ -13,34 +14,98 @@
...
+        return True
+    if request.method in SAFE_METHODS:
+        lists = obj.view_lists.all()
+        if not lists:
+            lists = obj.admin_lists.all()
+            if not lists:
+                return False
+        if has_common_lists(request.user, [mlist.name for mlist in lists]):
+            return True
+        return has_admin_permission(obj, request)
```
mbertrand: will do

ui/permissions.py
```diff
@@ -13,34 +14,98 @@
...
+        True or False
+
+    """
+    if request.user == obj.owner or request.user.is_superuser:
+        return True
+    lists = obj.admin_lists.all()
+    if not lists:
+        return False
+    if has_common_lists(request.user, [mlist.name for mlist in lists]):
+        return True
```
mbertrand: ok

ui/permissions.py
```diff
@@ -13,34 +14,98 @@
...
+        return has_admin_permission(obj, request)
+
+
+class IsAdminOrHasViewPermissionsForVideo(IsAuthenticated):
+    """Permission to view a video, based on its collection"""
+    def has_object_permission(self, request, view, obj):
+        return has_view_permission(obj.collection, request)
+
+
+class IsAdminOrHasAdminPermissionsForVideo(IsAuthenticated):
```
mbertrand: ok

gsidebo:
General comment: please try to avoid really large changes in the middle of a review cycle that aren't directly related to the PR feedback (for example, the js changes, the large amount of changes to `ui/permissions_test.py`, etc.). It makes the review process more challenging. If you realize some features/test/etc. are missing in the PR, you can change the label to 'Work in Progress' and let the reviewer know when you're starting and when you wrap up. In terms of refactoring unrelated within a PR, I guess the general guideline is just to try and keep the reviewer's work load light within reason. We discussed this a few times in sprint retros before you joined the team so basically this is just me filling you in 😄

ui/views_test.py
mbertrand: ok

ui/permissions_test.py
mbertrand: ok

ui/permissions_test.py
mbertrand: In this case it seems the fixture just made the test more obtuse, how about just adding this additional assert just above it to make it clear the users are different:
`assert request_data.user != collection.owner`

mbertrand:
Sorry, some of the major changes especially with the JS was due to having to rebase with master after the merge of https://github.com/mitodl/odl-video-service/issues/135 in order to avoid conflicts

ui/templates/ui/collection_list.html
mbertrand: ok

ui/permissions.py
mbertrand: The more succinct way also seems less readable/intuitive to me, also won't work quite right as is (collection is an empty list, not literally `None`, so would have to do something like `len(collection) > 0`.

gsidebo:
Understood. No worries!

PR #159, opened 2017-08-22
Added function to move uploaded files to another local folder

shaidar:
Related to closed [Issue#100](https://github.com/mitodl/odl-video-service/issues/100)

#### What's this PR do?
- Once files have been synced to S3, move them to another local folder on disk
- Added synced folder entry to settings file
- Used the `s3api head-bucket` call instead of `s3 ls` to verify that the bucket exists

I manually tested this on my local Win10 VM and all appears to be functional.

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/s3_sync/s3_sync.py
+++ b/s3_sync/s3_sync.py
@@ -11,6 +11,7 @@ python s3_sync.py -i <settings_file.ini>
 """
 import argparse
 import os
+import re
 import subprocess
 import sys
 from configparser import ConfigParser, ExtendedInterpolation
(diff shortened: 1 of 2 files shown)


s3_sync/s3_sync.py
```diff
@@ -151,6 +151,26 @@ def sync_local_to_s3(local_video_records_done_folder, s3_bucket_name):
     logger.info("Syncing complete")
 
 
+def move_files_to_synced_folder(local_video_records_done_folder, local_video_records_synced_folder):
```
blarghmatey: This should probably operate on a list of files that were successfully synced, otherwise we run the risk of moving a video file out of the source directory when it hasn't yet been synced to S3. This could happen because of a sync failure (partial or total), or because the file was written to disk in the time between when the sync started and when it completes.

shaidar: Added requested functionality.

blarghmatey:
:+1:

Issue #160, opened 2017-08-24
A collection will show up multiple times in list if creator is in one of the moira lists it's assigned to to.


mbertrand:
Easily fixable by adding `.distinct()` to the end of the filter queries.

Issue #5, opened 2017-06-13
Limit video upload to django staff


mbertrand:
At this time, we want to limit access to uploading videos (and creating collections) to only MITx affiliates. The easiest way to do this to limit access to users with the django staff role. Elaine or Peter can take responsibility for adding users as necessary. 

At some point in the future, we may want to use a moira list to maintain these access controls.

Ferdi:
Yes, if they pass touchstone login , they can upload videos . Right @pdpinch  ?

pdpinch:
We're going to need some kind of authorization, because anyone can get a TouchStone collaborator account. 

Let's talk and then I'll update the issue.

mbertrand:
Related question: If a video is restricted via moira, should a user who isn't in the appropriate moira list still be able to visit the video's detail page and read the title, description etc, but just not be able to view the video?

pdpinch:
I updated the description to reflect the feedback we got from the Product Council.

mbertrand:
@pdpinch non-staff users who are admins for a collection (via moira_list) should still be able to upload to that collection, right?  It's just creating collections that should be restricted to staff users (or superusers)?

Ferdi:
@pdpinch @mbertrand I thought we'd do this via a special moira list ?

pdpinch:
I assumed that using staff was faster to code. If that's not the case, we should go with a moira list because that is the eventual goal (as mentioned in the description).

mbertrand:
Using staff was easy to code.  Using a moira_list would be slightly more effort but not too bad, would need another setting to specify which moira list to use.

PR #161, opened 2017-08-24
Only staff users should be able to create collections

mbertrand:
Fixes https://github.com/mitodl/odl-video-service/issues/5
Fixes https://github.com/mitodl/odl-video-service/issues/160

#### What's this PR do?
- Prevents a user from creating a collection (front-end and API) if user is not staff or superuser
- Prevents a collection from showing up more than once in the list when a user is both the  owner and a member of one of its admin/view moira lists.
- Also adds missing moira lists fields to front-end form

- Create a new user (not staff and not superuser)
- After logging in as that user, you should not see the collection creation form on the collection lists page
- If you try to post a new collection using the API, you should get an access denied message
- As admin, change the user to a staff user.
- Repeat the above: you should be able to see & use the collection creation form on the lists page, and be able to post a new collection using the API (but only if the owner id == the user id).
- Repeat the above as a superuser.  You should be able to see & use the collection creation form on the lists page, and be able to post a new collection using the API (with any owner id as long as it's an existing user and not null).

### DIFF
--- a/ui/permissions.py
+++ b/ui/permissions.py
@@ -30,6 +30,13 @@ def has_common_lists(user, list_names):
     return len(set(user_moira_lists(user)).intersection(list_names)) > 0
 
 
+def is_staff_or_superuser(user):
+    """
+    Determine if a user is either a staff or super user
+    """
+    return user and (user.is_superuser or user.is_staff)
+
+
 def has_view_permission(obj, request):
     """
     Determine if a user can view a collection or its videos based
@@ -85,8 +92,19 @@ class IsAdminOrReadOnly(BasePermission):
 
 class HasCollectionPermissions(IsAuthenticated):
     """
-    Permission to view or edit collection, currently both limited to users with admin access
+    Permission to view, edit, or create collections
+    View/edit currently both limited to users with admin access
+    Creation currently limited to staff or superusers
     """
+    def has_permission(self, request, view):
+        if request.method == 'POST':
+            if not is_staff_or_superuser(request.user):
+                return False
+            owner = str(request.data.get('owner'))
+            if request.user.is_staff and owner and owner.isdigit():
+                return int(owner) == request.user.id
+        return True
+
     def has_object_permission(self, request, view, obj):
         return has_admin_permission(obj, request)
 
--- a/ui/views.py
+++ b/ui/views.py
@@ -69,7 +69,8 @@ class CollectionList(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['collection_list'] = Collection.objects.all_admin(self.request.user)
-        context['form'] = forms.CollectionForm(initial={'owner': self.request.user.id})
+        if ui_permissions.is_staff_or_superuser(self.request.user):
+            context['form'] = forms.CollectionForm(initial={'owner': self.request.user.id})
         context["js_settings_json"] = json.dumps(default_js_settings(self.request))
         return context
 
@@ -85,7 +86,6 @@ class CollectionDetail(TemplateView):
         if not ui_permissions.has_admin_permission(collection, self.request):
             raise PermissionDenied
         video_list = Video.objects.filter(collection=collection)
-
         default_settings = default_js_settings(self.request)
         context["collection"] = collection
         context["video_list"] = video_list
@@ -205,7 +205,7 @@ class CollectionViewSet(viewsets.ModelViewSet):
     )
     permission_classes = (
         permissions.IsAuthenticated,
-        ui_permissions.HasCollectionPermissions,
+        ui_permissions.HasCollectionPermissions
     )
 
     def get_queryset(self):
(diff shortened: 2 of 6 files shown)


ui/views_test.py
```diff
@@ -72,6 +72,21 @@ def user_admin_list_data():
...
+    client, user = logged_in_apiclient
+    url = reverse('models-api:collection-list')
+
+    input_data = {
+        'owner': user.id,
+        'title': 'foo title',
+        'view_lists': [],
+        'admin_lists': []
+    }
+    return SimpleNamespace(client=client, user=user, url=url, data=input_data)
```
gsidebo: The client isn't really _data_. `post_data.client` is awkward to read and reason about, and I think it's important to not use fixtures to store all kinds of unrelated data. I would move that out of the return value namespace and have each test case just specify `logged_in_apiclient` in the args list. This is made a little bit confusing due to the multiple return values approach that `logged_in_apiclient` uses, but that can be fixed in a later PR. I would also move `url` out since it isn't data, and instead make that a class level variable (you can group these tests into a class since they're all testing the same view class). What do you think about something like this?:

```python
class CollectionListTests:
    url = reverse('models-api:collection-list')
    
    @pytest.fixture
    def input_data(self, logged_in_apiclient):
        _, user = logged_in_apiclient
        return {
            'owner': user.id,
            'title': 'foo title',
            'view_lists': [],
            'admin_lists': []
        }
    
    def test_collection_viewset_create_as_normal_user(logged_in_apiclient, input_data):
        client, user = logged_in_apiclient
        result = client.post(self.url, input_data, format='json')
        # ...
```

Let me know if you have any disagreements about the pattern in general

ui/permissions.py
```diff
@@ -30,6 +30,13 @@ def has_common_lists(user, list_names):
     return len(set(user_moira_lists(user)).intersection(list_names)) > 0
 
 
+def has_staff_permission(user):
```
mbertrand: OK

ui/permissions.py
```diff
@@ -83,6 +90,15 @@ def has_permission(self, request, view):
         return request.user and request.user.is_superuser
 
 
+class IsAdminOrStaff(BasePermission):
+    """IsAdminOrStaff permission to create new collections"""
+    def has_permission(self, request, view):
+        if request.method == 'POST':
+            if not has_staff_permission(request.user):
+                return False
+        return True
```
mbertrand: great idea, though the above will need a little tweaking since superusers can create collections for other users but staff can't.

ui/views_test.py
mbertrand: How about a compromise with no test class, but just the input_data as a fixture (still called `post_data` because eventually we might need `put_data` and `patch_data`), and url specified in each test for clarity:
```
@pytest.fixture
def post_data(logged_in_apiclient):
    """Fixture for testing collection creation using valid post data"""
    _, user = logged_in_apiclient

    input_data = {
        'owner': user.id,
        'title': 'foo title',
        'view_lists': [],
        'admin_lists': []
    }
    return input_data

def test_collection_viewset_create_as_normal_user(post_data, logged_in_apiclient):
    url = reverse('models-api:collection-list')
    result = client.post(url, post_data, format='json')
    assert result.status_code == status.HTTP_403_FORBIDDEN
```

ui/views.py
```diff
@@ -69,7 +69,8 @@ class CollectionList(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['collection_list'] = Collection.objects.all_admin(self.request.user)
-        context['form'] = forms.CollectionForm(initial={'owner': self.request.user.id})
+        context['form'] = (forms.CollectionForm(initial={'owner': self.request.user.id})
+                           if ui_permissions.is_staff_or_superuser(self.request.user) else None)
```
gsidebo: We can just leave this key/value out of `context` if the permission doesn't exist, i.e.:

```python
if ui_permissions.is_staff_or_superuser(self.request.user):
    context['form'] = forms.CollectionForm(initial={'owner': self.request.user.id})
```

The template doesn't break if the key is missing

mbertrand: will do, thanks for the review!

mbertrand:
@gsidebo looks like that one last change was enough to [cause codecov tests to fail](https://codecov.io/gh/mitodl/odl-video-service/compare/bbad3e261c420b9027e47979ae69c55d0450e1ea...ab941e2853e90d5736b8bb7d5fc01a2199662d82/diff), so I'll go ahead and create another unit test for the existence of context['form'] for admins vs non-admins.  Let me know if you'd like to look it over again afterward.

pdpinch:
You should treat codecov as advice, which you can take or leave. It doesn't need to pass before you can merge.

mbertrand:
Thanks @pdpinch, I already created the tests, they were really simple and fixed the codecov issue.  I will go ahead and merge soon unless there are any objections.

PR #143, opened 2017-08-15
Use material components for video detail page

noisecapella:
#### What are the relevant tickets?
Fixes #135 
Fixes #136 

WIP

PR #162, opened 2017-08-24
style changes to video details page

roberthouse54:
#143 
#### What's this PR do?
Styles the page using css, tweaking the material components 

See that the page works on desktop and mobile, and that nothing is broken or looks wrong.

I think we should not use the "CSS only" forms in this library. We should either use the standard animated material forms with floating labels, or we should implement our own forms. 

Also, there is a large margin underneath the video title, because there is a space for the video description field, even if one has not been entered. Ideally we should hide this element if there is no entered description.


aliceriot:
not sure if we need to fix this now or not, but if you do mobile in portrait mode the video seems to be smaller than it should be:

strange_little_thing

aliceriot:
otherwise the styling looks good!

package.json
```diff
@@ -15,6 +15,7 @@
     "@material/menu": "^0.4.3",
     "@material/textfield": "^0.3.3",
     "@material/toolbar": "^0.4.4",
+    "@material/typography": "^0.3.0",
```
aliceriot: for this change, when you ran `yarn add @material/typography` it should have also changed the `yarn.lock` file, we need to get those changes too to ensure we always install the same version.

static/scss/generalStyles.scss
```diff
@@ -0,0 +1,55 @@
...
+  color: $linkcolor;
+  text-decoration: none;
+}
+
+.fontgray {
+  color: $fontgray;
+}
+
+button {
+
```
aliceriot: can we remove this blank line?

static/scss/header.scss
```diff
@@ -1,4 +1,5 @@
 @import '_variables';
+@import 'breakpoints';
```
aliceriot: We should be able to just import `'breakpoints'` in the main file (`layout.scss`) and use it in any file we import afterwards (so we only have to have it imported in one place). We should be able to do the same thing with `_variables`.

pdpinch:
We can deal with the portrait mode in a follow-up issue -- just be sure to mention which device/browser you're using.

roberthouse54:
I fixed the portrait mode thing already

static/scss/generalStyles.scss
roberthouse54: All set

static/scss/header.scss
roberthouse54: Fixed

package.json
roberthouse54: alice said "actually, you know what, you don't have to worry about that"

static/scss/videoPlayer.scss
```diff
@@ -28,6 +26,11 @@
   position: relative !important;
   width: inherit !important;
   height: inherit !important;
+  max-height: calc(100vh - 200px);
+
+  @include breakpoint(phone) {
+    max-height: calc(100vh - 60px);
```
aliceriot: This doesn't quite seem to fix it for me - I think what we need to do is apply the `max-height` only when the video player is not fullscreen. I think something like this would do it:

```scss
.video-js {
  &:not(.vjs-fullscreen) {
    .vjs-tech {
      max-height: calc(100vh - 200px);
      
      @include breakpoint(phone) {
        max-height: calc(100vh - 60px);
      }
    }
  }
}
```

so that we don't set a `max-height` when the thing is fullscreened, but we do otherwise.

roberthouse54:
OK, I think this is now fixed for full-screen mode as well as portrait mode.

aliceriot:
yeah it's fixed!

PR #174, opened 2017-08-31
Created React pages for collection list and detail

### LINKED ISSUE #137: Front-end for collection detail page
As discussed, this will be independent from the video view page (#135) and the video embed page (#136), i.e.an independent page/template. This should save users who are only viewing one video (or just an embedded video) from having to download the entire react app. 

Mock-ups at https://projects.invisionapp.com/d/main#/projects/boards/5469149

gsidebo:
Closes #137 

#### What's this PR do?
Adds React pages for collection list and detail

Log in with a user that has a few collections, and at least one collection with videos. Click on the collection and video links to make sure you are taken to the correct pages.

`static/js/containers/CollectionDetailPage.js` and `static/js/containers/CollectionListPage.js`

- `CollectionListPage.js` is a placeholder of sorts. We're not exactly sure yet about what we're going to do with the collection list page, and we might get rid of it. In the meantime, it made more sense to create a barebones list page in React than it did to leave the old static page. This is also why there are no `CollectionListPage.js` tests (yet)
- I started working on pulling out some of the common UI code (dialog boxes, common page structure, etc.) into separate files, but the diff was quickly ballooning up. I saved those changes to a different branch and opened a new issue to complete that work (https://github.com/mitodl/odl-video-service/issues/175). **For now, the sandwich menu and the share/edit links in the collections pages are not functional.**
- There's an issue with the react router not successfully changing the selected collection when the URL changes. I'll ask @aliceriot about this, but in the meantime you just need to reload the collection detail page if it's not showing the correct one.

ss 2017-08-31 at 08 29 32


ui/views_test.py
```diff
@@ -439,52 +441,6 @@ def test_collection_viewset_detail_as_superuser(logged_in_apiclient):
     assert result.status_code == status.HTTP_204_NO_CONTENT
 
 
-def test_collection_detail_permission_as_superuser(logged_in_apiclient):
```
mbertrand: It might be helpful to keep these tests. I think they should still work if
`url = reverse('collection-detail', kwargs={'collection_key': collection.hexkey})` 
is replaced with 
`url = '/collections/{}'.format(collection.hexkey)`

ui/migrations/0010_video_default_sort.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ui', '0009_moira_lists'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='video',
+            options={'ordering': ['created_at']},
```
mbertrand: Would descending order be better (more recent first)?  Maybe a question for Peter or Ferdi.

ui/urls.py
```diff
@@ -11,9 +11,11 @@
     url(r'^$', views.Index.as_view(), name='index'),
     url(r'^register/$', views.register, name='register'),
     url(r'^login/$', views.ui_login, name='login'),
-    url(r'^collections/$', views.CollectionList.as_view(), name='collection-list'),
-    url(r'^collections/(?P<collection_key>[0-9a-f]+)/$', views.CollectionDetail.as_view(), name='collection-detail'),
-    url(r'^collections/(?P<collection_key>[0-9a-f]+)/upload/$', views.Upload.as_view(), name='upload'),
+
+    url(r'^collections/', views.CollectionReactView.as_view(), name='collection-react-view'),
```
mbertrand: Is it standard when using a React frontend use the same view for both an individual object and a list of objects? This confused me at first, until I saw the React code that determines what to show based on the URL.  It seems less intuitive to me but it works so I'm okay with it.

gsidebo:
@mbertrand 
Re: collection detail:
- Sandwich menu: I created an issue to create common UI code that can be used between all pages in the app (#175). Right now, functionality that should be common among several pages is tightly coupled to `VideoDetailPage`. I think it will be less work to merge this without the sandwich menu/edit & share links/etc. and refactor later than to copy/paste the same code in a bunch of places and remove it later. I started work on the refactor as part of this PR, but stashed it when I considered what it was doing to the diff size. Let me know what you think.
- Load More Videos: I'm willing to bet that we don't need that yet and won't need it for a while. Unless we're expecting to have ~100+ videos in each collection, I really don't think we need to page the API results.
- Thumbnail size: Waiting to hear back from Alice, and I'll definitely standardize the card size. Good catch.

Re: collection form - agreed. I'll do that.

ui/urls.py
gsidebo: Using React, it's pretty standard to have an entire web application using the same view! If you look at Micromasters (browser-accessible URLs [here](https://github.com/mitodl/micromasters/blob/master/ui/urls.py)), there is just one view (as far as the backend is concerned) for both the learners page and the dashboard page, but those two views represent almost the entire UI for the app.

ui/views_test.py
gsidebo: The tests for the collection detail API are testing the same things, and that is what the react view uses to fetch collection detail data. It would be redundant to keep this around (unless I'm missing something)

ui/migrations/0010_video_default_sort.py
Ferdi: right, more recent videos come first

Ferdi:
I think we can ignore "load more" for now. Just open another issue

ui/views_test.py
mbertrand: okay

gsidebo:
@mbertrand Let me know if the dimensions are off for the card. I didn't get to test with the updated thumbnail size

mbertrand:
@gsidebo the card dimensions seem fine, the thumbnails are displaying a bit smaller than they might need to be (max 415x200 instead of ~600x338) but the cards are consistent and look good.  The branch needs to be rebased with master though, there are some conflicts.  Also there is one JS lint error in CollectionsApp.js

mbertrand:
@gsidebo After talking with Rob it seems like the upload page needs to be changed: there should be an 'Add Videos from Dropbox' button on the collection detail page (I thought that's what 'Load More Videos' was but turns out that's just paging), and the progress bar should appear in a new card as shown in https://projects.invisionapp.com/boards/863980TEFVYP7#/5469149/167848254 - not sure if this should fall under #175 or not.

pdpinch:
Let's open separate issues for these changes. I don't think we need them for tomorrow for Lecture Capture. 

Are we going to be able merge and release this today?

mbertrand:
@pdpinch I modified #183 and added #184 for these changes.

gsidebo:
Trying to upload videos with thumbnails so I can post a screenshot for Rob to look at, but experiencing some trouble with uploading

pdpinch:
We're running out of time. Let's merge this and we can open additional issues for any styling fix-up.

gsidebo:
@roberthouse54 Does this look OK height/width wise?

ss 2017-09-05 at 14 24 31

mbertrand:
👍

PR #191, opened 2017-09-05
OVS style changes

roberthouse54:
#174 

#### What's this PR do?
Fixes some of the styling on the collection details page and the collections list page

See that the styles look correct and that the css is valid


mbertrand:
👍

PR #195, opened 2017-09-07
Added collection edit dialog form

### LINKED ISSUE #190: Collection settings on Collection detail page
As a collection owner, I need to be able to change the settings for a collection after it's created. This is especially important for collections that are automatically created for lecture capture videos. 

- [ ] add gear icon to collection detail page (see https://projects.invisionapp.com/boards/863980TEFVYP7#/5469149/167848254)
- [ ] On clicking the gear icon, the user should get a menu similar in function to https://video.odl.mit.edu/collection_form/ but design should look like https://projects.invisionapp.com/boards/YD3980TTNKC32/

gsidebo:
Closes #190

#### What's this PR do?
Adds the collection edit form to the collection detail page (collection creation can use this form as well with some tweaking)

Open a collection detail page, click the gear icon, change some things around, and save.df

Probably `CollectionForm.js` 

- I added the show/hide dialog functionality as a higher-order component since I already did some of that work while working on the collection detail page.
- It seems like our dialog component doesn't have any internal scrolling capability, and the containing page doesn't show a scrollbar when the dialog is too big. That could be a deal-breaker for a dialog this big.
- I didn't do any special error handling due to time constraint & lack of design

ss 2017-09-07 at 02 00 56


static/js/components/dialogs/hoc.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+import React from 'react';
+import R from 'ramda';
+
+import * as commonUiActions from '../../actions/commonUI';
```
mbertrand: Minor nitpick, feel free to ignore if you prefer: The filenames for commonUI and collectionsUI have a capitalized 'I' while videosUi has a lowercase 'i'

static/js/containers/CollectionDetailPage.js
```diff
@@ -21,6 +26,7 @@ class CollectionDetailPage extends React.Component {
     collectionKey: string,
     editable: boolean,
```
mbertrand: I noticed that the gear icon is always enabled and will show the edit dialog even if the user does not have edit permission, but a 'You do not have permission' page will load after the form is submitted.  Ideally the gear icon would be disabled instead (@pdpinch maybe this should wait for a future PR?). 

The `CollectionReactView` would need to be modified to something like this on order to return the appropriate `editable` value if a detail page is being requested:
```python
    def get_context_data(self, **kwargs):  # pylint: disable=arguments-differ
        keymatch = re.search(r'/([0-9a-f]{32})/$', self.request.path)
        context["js_settings_json"] = json.dumps({
            **default_js_settings(self.request),
            'editable': keymatch and ui_permissions.has_admin_permission(
                Collection.objects.get(key=keymatch.group(1)), self.request)
        })
        return context
```
However I couldn't quite figure out how to pass that value through the router to the `CollectionDetailPage`

pdpinch: If it can't be done today, we should leave this for another issue. 

Ideally, the gear icon shouldn't display at all for someone without admin permission.

mbertrand: If `editable` is passed to the `CollectionDetailPage` props, the gear icon could be selectively displayed like this:

```jsx
    return <div className="centered-content">
      <header>
        ......
        {editable &&
        <div className="tools">
          <a onClick={R.partial(showDialog, [DIALOGS.NEW_COLLECTION])}>
            <i className="material-icons">settings</i>
          </a>
        </div>
        }
      </header>
      { this.renderCollectionVideos(videos) }
    </div>;
```

gsidebo: Good catch, Matt. Returning the permission as part of the API call might also work. I'll tinker with it, but if it looks too heavy I'll create a separate issue

static/js/components/dialogs/hoc.js
gsidebo: I went back and forth mentally about whether I prefer `*Ui` or `*UI`, and I think I agree with you that `*Ui` is preferable. I'll change them

mbertrand:
@gsidebo The functionality works great now! Some of the tests are failing though, once they're passing again I'll approve the PR.

mbertrand:
Just one lint issue and then it should be good to go:
```
/src/static/js/components/dialogs/hoc_test.js
  10:59  error  Missing semicolon  semi
✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.
```

gsidebo:
@mbertrand Weird... local linter didn't pick that up. Anyway, it's fixed

mbertrand:
Looks great! 👍

Issue #165, opened 2017-08-25
Wrong contents in sandwich menu


pdpinch:
Design at https://projects.invisionapp.com/boards/Q33980TUMV29A/

### acceptance criteria:

- [ ] name of logged in user
- [ ] logout link (i.e. https://video-ci.odl.mit.edu/logout/)
- [ ] List of collections the user can admin

### out of scope (for now)
- [ ] create collection button
- [ ] list of collections user can view

PR #197, opened 2017-09-07
Styling the side-drawer

roberthouse54:
#165 

#### What's this PR do?
This sets the styling for the sidedrawer. This PR also includes making the "OVS Video Services" page header a link to the root directory. The side drawer should look as follows:

image

See that the style matches the above, and that the css looks correct.


gsidebo:
@roberthouse54 If you haven't already, can you make sure the drawer contents look good as the window size shrinks vertically?

roberthouse54:
I just rebased. Should be ok now.

PR #199, opened 2017-09-08
Changed default mailgun_url to video.odl.mit.edu domain

shaidar:
No relevant tickets as this is just changing a fallback value in case it's not specified.

#### What's this PR do?
Changed default mailgun_url to video.odl.mit.edu domain instead of micromasters as this service has its own mailgun domain.

Check that the right value for `MAILGUN_URL` is set

(Optional)

(Optional)

(Optional)

(Optional)


blarghmatey:
:+1:

PR #222, opened 2017-09-11
Enable google analytics for video & collection pages

### LINKED ISSUE #139: Google analytics: Page view stats

### LINKED ISSUE #304: Google Analytics: capture video events
As a collection admin, I would like to know how many viewers have viewed my video and how they interacted with it. 

There appears to be a number of Google Analytics plugins for VideoJS. We should choose one that supports a reasonably similar set of events to what edX collects, in the interests of supporting similar analysis tools. It would also be useful to find one that is extensible so we can track multiangle playback events.

### LINKED ISSUE #305: Google Analytics: Capture multiangle interactions
As a collection admin who has multiangle videos in their collection, I'd like to be able to track multiangle events, such as when the user changes the camera angle. 

We could use this data to determine the utility of the multiangle format. Another potential use for this data would be to "crowdsource" an edit decision list, which could then be used to automate the process of cutting the mutliangle video down to a single video steam.

mbertrand:
Fixes #139
Fixes #304 
Fixes #305

#### What's this PR do?
- Adds google analytics tracking  (pageviews, video player events) to `/collections` and `/videos` pages

- Create a Google analytics account, use something like `http://ovs.mit.edu:8089` for the website URL
- Update your `/etc/hosts` file so that the above domain points to your docker IP.
- Set `GA_TRACKING_ID=<your account id>` in your `.env` file
- Go to the `REAL-TIME/Overview` tab of your Google Analytics account.
- In another tab, start visiting pages in OVS using 'http://ovs.mit.edu:8089" as the URL.  For each page wait a couple seconds after loading, and it should appear in the GA console under `Top Active Pages`
- On a video detail page, open up the network tab of the developer console.  Play, pause, and FF the video, turn subcaptions on/off, go in/out of fullscreen.  For each of these, a 'collect' POST request should be sent to Google Analytics.

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -1,12 +1,15 @@
 // @flow
 /* global videojs: true */
-import React from 'react';
+import React from "react";
 
 import { makeVideoSubtitleUrl } from "../lib/urls";
 import { getHLSEncodedUrl, videojs } from "../lib/video";
 import type { Video, VideoSubtitle } from "../flow/videoTypes";
 import { FULLSCREEN_API } from "../util/fullscreen_api";
 import { CANVASES } from "../constants";
+import { sendGAEvent } from "../util/google_analytics";
+
+const gaEvents = ["play", "pause", "seeked", "timeupdate", "fullscreen off", "fullscreen on", "ended"];
 
 const makeConfigForVideo = (video: Video): Object => ({
   autoplay: false,
@@ -19,14 +22,14 @@ const makeConfigForVideo = (video: Video): Object => ({
   playbackRates: [0.50, 0.75, 1.0, 1.25, 1.5, 2.0, 4.0],
   sources: [{
     src: getHLSEncodedUrl(video),
-    type: 'application/x-mpegURL',
+    type: "application/x-mpegURL",
   }]
 });
 
 const drawCanvasImage = function(canvas, videoNode, shiftX, shiftY) {
   const x = shiftX ? Math.floor(videoNode.videoWidth / 2) : 0;
   const y = shiftY ? Math.floor(videoNode.videoHeight / 2) : 0;
-  const context = canvas.getContext('2d');
+  const context = canvas.getContext("2d");
   context.drawImage(
     videoNode,
     x, y, Math.floor(videoNode.videoWidth / 2), Math.floor(videoNode.videoHeight / 2),
@@ -51,6 +54,7 @@ export default class VideoPlayer extends React.Component {
   videoNode: HTMLVideoElement;
   videoContainer: HTMLDivElement;
   cameras: HTMLDivElement;
+  percentTracked: Array<number>;
 
   updateSubtitles() {
     const { video } = this.props;
@@ -76,6 +80,18 @@ export default class VideoPlayer extends React.Component {
             srcLang: subtitle.language,
             label: subtitle.language_name,
           }, true);
+        }
+        // Add listeners to each track
+        let player = this.player;
+        for (let idx = 0; idx < this.player.textTracks().length; idx++) {
+          if (!trackUrls.includes(tracks[idx].src)) {
+            tracks[idx].addEventListener("modechange", function() {
+              sendGAEvent(
+                "video",
+                "modechange",
+                `Subtitles for ${this.label} ${this.mode}`, player.currentTime());
+            });
+          }
         }
       });
     }
@@ -103,11 +119,11 @@ export default class VideoPlayer extends React.Component {
     }
   }
 
-  cropVideo() {
+  cropVideo = () => {
     const { selectedCorner } = this.props;
     const shiftX = CANVASES[selectedCorner].shiftX;
     const shiftY = CANVASES[selectedCorner].shiftY;
-    const transformProps = ['transform', 'WebkitTransform', 'MozTransform', 'msTransform', 'OTransform'];
+    const transformProps = ["transform", "WebkitTransform", "MozTransform", "msTransform", "OTransform"];
 
     const prop = transformProps.find(property => this.player.el_.style[property] !== undefined) || transformProps[0];
     const aspectRatio = (this.player.videoWidth() / this.player.videoHeight());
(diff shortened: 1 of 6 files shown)


aliceriot:
works as advertised! reading code.

static/js/components/VideoPlayer.js
```diff
@@ -129,38 +145,66 @@ export default class VideoPlayer extends React.Component {
...
       document[FULLSCREEN_API.exitFullscreen]();
     } else {
       // $FlowFixMe videoContainer.parentElement is not going to be null
       this.videoContainer.parentElement[FULLSCREEN_API.requestFullscreen]();
     }
+    this.player.el_.dispatchEvent(new Event(`fullscreen ${fullscreen ? "off" : "on"}`));
+  }
+
+  analyzeEvent(action: string, label: string) {
+    let percentTracked = this.percentTracked;
```
aliceriot: is this `percentTracked` thingy just for tracking which values we've already tracked?

static/js/components/VideoPlayer.js
```diff
@@ -129,38 +145,66 @@ export default class VideoPlayer extends React.Component {
...
+    this.player.on(action, function() {
+      if (action === "timeupdate") {
+        // Track percentage played in increments of 10
+        const currentTime = Math.round(this.currentTime());
+        const duration = Math.round(this.duration());
+        let percentPlayed = Math.round(currentTime / duration * 100);
+        _.range(10,100,10).forEach((percent:number) => {
+          if (percentPlayed >= percent && !percentTracked.includes(percent)) {
```
aliceriot: idk if I'm reading this right, but it seems like what we want to do is find the closest multiple of 10 percentage value that is `<=` to our current percentage, right?

if that's right I think we can avoid iterating over this `_.range` call by doing this:

```js
const percentRoundTen = percentPlayed - (percentPlayed % 10)
```

static/js/components/VideoPlayer.js
```diff
@@ -129,38 +145,66 @@ export default class VideoPlayer extends React.Component {
...
+      } else {
+        sendGAEvent("video", `${action}`, label, Math.round(this.currentTime()));
+      }
+    }, false);
   }
 
   componentDidMount() {
     const { video } = this.props;
     const cropVideo = this.cropVideo.bind(this);
+    const analyzeEvent = this.analyzeEvent.bind(this);
```
aliceriot: if you declare `this.analyzeEvent` as an arrow function I think you can avoid doing this `.bind(this)` fiddling here

static/js/components/VideoPlayer.js
```diff
@@ -129,38 +145,66 @@ export default class VideoPlayer extends React.Component {
...
       // $FlowFixMe
       document[FULLSCREEN_API.exitFullscreen]();
     } else {
       // $FlowFixMe videoContainer.parentElement is not going to be null
       this.videoContainer.parentElement[FULLSCREEN_API.requestFullscreen]();
     }
+    this.player.el_.dispatchEvent(new Event(`fullscreen ${fullscreen ? "off" : "on"}`));
+  }
+
+  analyzeEvent(action: string, label: string) {
```
aliceriot: this function is kinda mis-named to me, since what we're doing here is not analyzing an event but setting up an event handler which will do that when the event fires. Can we call it something like `createEventHandler` instead?

static/js/components/VideoPlayer.js
mbertrand: Yes, like 10%, 20%, etc.  So it won't resend an event for 10% tracked at 20% and 30% tracked etc.

static/js/components/VideoPlayer.js
mbertrand: Thanks, will do

static/js/components/VideoPlayer.js
mbertrand: sure

static/js/components/VideoPlayer.js
aliceriot: So does that mean if I tracked to 12%, then to 22%, and then back to 13%, we'll only see two events? It seems like maybe we should record three events in that case.

mbertrand: @pdpinch do you have any preference on this?

pdpinch: I think I'm going to need more context.

mbertrand: @pdpinch:
- User watches > 10% of the video, GA sends notification that user has watched 10%.
- User fast-forwards to 23% of the video, GA sends notification that user has watched 20%.
- User rewinds video back to 13%.  GA does not send notification that user has watched 10% again.  
- User keeps watching past 20%. GA does not send another notification that user has watched 20%.
- User keeps watching to 31%, and GA sends notification that user has watched 30% of the video.

mbertrand: @pdpinch @aliceriot my opinion is that percentPlayed for any 10% increment should be recorded only once per page view, the first time it's reached. @blarghmatey talked to me about previous experiences with recording this, it can be a complex issue depending on exactly what info you want to get out of it - for example differentiating between watching the first 50% of a video vs watching 40% then FF'ing to 50%.  Maybe this should be fleshed out some more in a separate issue and future PR.

mbertrand: After discussing with @pdpinch, decision for now is to stick with current functionality since it generally matches what is done with the videojs-ga plugin.  @noisecapella since Alice is on vacation would you mind finishing up the review?

static/js/components/VideoPlayer.js
```diff
@@ -78,6 +82,17 @@ export default class VideoPlayer extends React.Component {
           }, true);
         }
       });
+      // Add listeners to each track
+      const player = this.player;
+      for (let track of tracks.tracks_) {
+        track.addEventListener("modechange", function() {
+          sendGAEvent(
```
noisecapella: It looks like this line is uncovered, can you write a test for it?

mbertrand: This part of the code is all stubbed out - would need an actual real videojs player object being available.

static/js/components/VideoPlayer_test.js
```diff
@@ -40,18 +41,23 @@ describe('VideoPlayer', (props = {}) => {
...
+    gaStub = sandbox.stub(libGA, 'sendGAEvent');
     playerStub = {
       el_: {
-        style: {}
+        style: {},
+        dispatchEvent: sandbox.stub()
       },
       tracks: [],
+      on: sandbox.stub(),
+      currentTime: function() {return 30.5;},
```
noisecapella: Can you change this function and `duration` to be arrow functions: `() => 30.5` for example? (The other functions were updated in a separate PR)

static/js/util/google_analytics.js
```diff
@@ -0,0 +1,38 @@
...
+
+const formatGAEvent = R.compose(removeNilValue, makeGAEvent);
+
+export const sendGAEvent = (
+  category: string,
+  action: string,
+  label: string,
+  value?: number
+) => {
+  ga.event(formatGAEvent(category, action, label, value));
```
noisecapella: Can you add a `Math.round(value)` here and also a check that `value` is a `number`?

mbertrand: Updated (`Math.round` at `makeGAEvent`; new `isValidNumber` function).

static/js/components/VideoPlayer_test.js
mbertrand: OK

PR #229, opened 2017-09-12
Sort collections by reverse creation time

### LINKED ISSUE #211: Add collection ordering
Collections are ordered by their UUID keys right now, which results in random ordering as far as the end user is concerned. We should probably add a field to the collection table to store the creation date and order the collections based on that. We should consider using the `TimestampedModel` approach that Micromasters uses for model definition

### acceptance criteria
- [ ] Make collection a `TimestampedModel`
- [ ] add a sort to the collection listing to order by creation date descending

mbertrand:
Fixes #211 

#### What's this PR do?
Adds a `created_at` field to the Collection model and sets its default sort order to created_at descending.

- `docker-compose run web python manage.py migrate`
- create a few new collections
- Your collection list should show the new collections in the reverse order that you created them.


ui/models.py
```diff
@@ -97,9 +97,13 @@ class Collection(models.Model):
     owner = models.ForeignKey(settings.AUTH_USER_MODEL)
     view_lists = models.ManyToManyField(MoiraList, blank=True, related_name='view_lists')
     admin_lists = models.ManyToManyField(MoiraList, blank=True, related_name='admin_lists')
+    created_at = models.DateTimeField(auto_now_add=True, blank=True, null=True)
```
gsidebo: Should this be defined with `blank=True, null=True`? `Video.created_at` is not defined with those properties, and it seems like the same should be true here. When I migrated, the `created_at` values for my existing collections were all filled in automatically

mbertrand: I can definitely get rid of `blank=True`.  If I don't set `null=True`, the migrate command will insist upon a manually entered default value for the field since there are already existing collections.  I thought it would be best to leave them null instead.

gsidebo: Ah, right. Let's do this in 2 migrations then. We've done similar migrations in micromasters. Check out these 3 files for reference: (1) https://github.com/mitodl/micromasters/blob/master/exams/migrations/0011_add_date_grades_available.py, (2) https://github.com/mitodl/micromasters/blob/master/exams/migrations/0012_populate_date_grades_available.py, (3) https://github.com/mitodl/micromasters/blob/master/exams/migrations/0013_require_date_grades_available.py

The first adds the field and makes it nullable, the second fills in all existing records with a value for that field, then the third sets the field to non-null. I'm almost positive that steps 2 and 3 can be combined into a single migration. Basically you would set the field to non-null in the model, create a new migration, then edit that new migration file manually to include the code to fill in the date. The `operations` for the migration would look something like this:

```python
    operations = [
        migrations.RunPython(populate_collection_created, migrations.RunPython.noop),
        migrations.AlterField(
            model_name='collection',
            name='created_at',
            field=models.DateTimeField(),
        ),
    ]
```

Let me know what you think

PR #240, opened 2017-09-13
Add dialog for creating a new collection

### LINKED ISSUE #227: Improve Create Collection UI
This combines issue #206 and #212 

#190 added a dialog form to edit a collection. This same form can/should be leveraged to create a new collection.

### acceptance criteria

- [ ] Add "Create Collection" button to sandwich menu, for users with permission to create collections
- [ ] When clicked, open the dialog form established in #190 
- [ ] Add a "Create a Collection" button to the collection list page (/collections/)
- [ ] When the new collection is saved, redirect the user to the collection detail page

noisecapella:
Fixes #227

#### What's this PR do?
Refactors the collection edit dialog to also be used in creating a new collection, and removes the collection form

 - Create a new collection using the button in the drawer. The collection should be created successfully and it should go on the beginning of the collection list. Make sure to add items to the moira list.
 - Create a new collection using the 'Create Collection' button at `/collections/`. You should see the same user experience.
 - Edit one of the collections and verify that this produces the expected change.


static/js/components/material/Drawer.js
```diff
@@ -20,22 +23,36 @@ type DrawerProps = {
...
+    if (this.createCollectionButton) {
+      this.createCollectionButton.addEventListener('click', (event: MouseEvent) => {
+        event.preventDefault();
+
+        onDrawerClose();
+        dispatch(collectionUiActions.setIsNew(true));
+        dispatch(collectionUiActions.initCollectionForm({}));
+        dispatch(commonUiActions.showDialog(DIALOGS.NEW_COLLECTION));
+      }, false);
+    }
```
gsidebo: One thing to note from the outset: the collection list page will soon be getting the Create Collection button. If it's not difficult to move it to its own file, we should do that here

noisecapella: Sure, I'll see about moving the logic to a library function

static/js/components/dialogs/CollectionFormDialog.js
```diff
@@ -109,25 +76,30 @@ class CollectionFormDialog extends React.Component {
...
-      (collection) => {
-        hideDialog();
-        this.initializeFormWithCollection(collection);
-      }
-    );
+
+    if (collectionUi.isNew) {
+      await dispatch(actions.collectionsList.post(payload));
+      hideDialog();
+      dispatch(uiActions.initCollectionForm(makeInitializedForm()));
```
gsidebo: I think it's better to just create a new action `CLEAR_COLLECTION_FORM` which would set the form to the initial state.

static/js/actions/collectionUi.js
```diff
@@ -26,3 +33,14 @@ export const setAdminLists = createAction(SET_ADMIN_LISTS);
 
 export const SET_SELECTED_VIDEO_KEY = qualifiedName('SET_SELECTED_VIDEO_KEY');
 export const setSelectedVideoKey = createAction(SET_SELECTED_VIDEO_KEY);
+
+export const SET_IS_NEW = qualifiedName('SET_IS_NEW');
+export const setIsNew = createAction(SET_IS_NEW);
+
+export const startNewCollectionDialog = (collection: ?Collection = null) =>
```
gsidebo: One nice feature we had before this PR was that the forms state was maintained when you opened/closed the form without submitting it or selecting a different collection. That wasn't specifically asked for, and it's kind of an edge case that someone would need it with this relatively small form, but it's worth noting we do lose that with this approach. Either way, I would argue that launching the new collection dialog doesn't need to do the form init at all. We can just clear it after submission. Maybe it's better to create 2 different actions here: `showNewCollectionDialog` and `showEditCollectionDialog`. `showNewCollectionDialog` would avoid that `initCollectionForm` call entirely

static/js/components/dialogs/CollectionFormDialog_test.js
```diff
@@ -54,55 +69,111 @@ describe('CollectionFormDialog', () => {
...
-    await listenForActions([INIT_COLLECTION_FORM], () => {
-      wrapper = renderComponent();
-    }).then(() => {
-      assert.notDeepEqual(
-        wrapper.find('CollectionFormDialog').prop('collectionUi').collectionForm,
-        uiState.collectionForm
-      );
+
+  for (const isNew of [true, false]) {
+    describe(`isNew is ${String(isNew)}`, () => {
```
gsidebo: For readability, please change to something like `with isNew=${String(isNew)}`

static/js/components/material/Drawer.js
```diff
@@ -20,22 +21,34 @@ type DrawerProps = {
...
+        event.preventDefault();
         onDrawerClose();
       }, false);
     }
+
+    // this will be undefined if SETTINGS.editable is false
+    if (this.createCollectionButton) {
+      this.createCollectionButton.addEventListener('click', (event: MouseEvent) => {
+        event.preventDefault();
+
```
gsidebo: Style nit: I don't think we need the blank line here

static/js/containers/CollectionDetailPage.js
```diff
@@ -94,9 +94,20 @@ class CollectionDetailPage extends React.Component {
       : null
   );
 
-  renderBody(collection: Collection) {
-    const { showDialog } = this.props;
+  showEditCollectionDialog = (e: MouseEvent) => {
+    const { dispatch, collection } = this.props;
+
+    e.preventDefault();
+
```
gsidebo: Style nit: I'd rather not have the blank line here

static/js/containers/CollectionListPage.js
```diff
@@ -65,13 +54,20 @@ class CollectionListPage extends React.Component {
...
+  };
+
   render() {
     const { commonUi } = this.props;
     const formLink = SETTINGS.editable
       ? (
-        <a className="button-link create-collection-button" href="/collection_form">
-          <i className="material-icons">add</i>
-          Create a Collection
+        <a href="#" className="button-link create-collection-button" onClick={this.openNewCollectionDialog}>
```
gsidebo: Just curious: what's the reasoning behind the `href` addition? Seems like we need `preventDefault` only because of that.

static/js/reducers/collectionUi.js
```diff
@@ -9,38 +9,55 @@ import {
...
+  editCollectionForm: INITIAL_COLLECTION_FORM_STATE,
+  isNew: true,
   selectedVideoKey: null
 };
 
+const getFormKey = (isNew: boolean): string => isNew ? "newCollectionForm" : "editCollectionForm";
+
+export const getCollectionForm = (
+  state: CollectionUiState
+): CollectionFormState => state[getFormKey(state.isNew)];
```
gsidebo: I mentioned this above but I think we can find a better place for this function

static/js/actions/collectionUi.js
noisecapella: Sure

noisecapella:
Comments should be addressed

noisecapella:
Actually I forgot to cover redirecting the user to the collection page after creating the collection. I'll add this now

noisecapella:
Ok, this should be in place now

static/js/components/dialogs/CollectionFormDialog_test.js
```diff
@@ -6,21 +6,39 @@ import { assert } from 'chai';
 import { Provider } from 'react-redux';
 import configureTestStore from 'redux-asserts';
 
-import CollectionFormDialog from './CollectionFormDialog';
+import CollectionFormDialog  from './CollectionFormDialog';
```
gsidebo: Guessing this was unintentional

PR #247, opened 2017-09-18
A few more style tweaks to OVS

roberthouse54:
#### What's this PR do?
This fixes some of the styles for OVS, specifically:
- styling for the dropbox upload button
- styling for the Create Collection button in the side drawer.
-  makes the videos in the grid aligned left as opposed to centered as currently.
- other various tweaks to margins, color, etc.
- adds a drop shadow to modal footer bar, so it is more clear if the modal has scrolling.

See that it looks ok, and the css is valid.


pdpinch:
Question, does "makes the videos in the grid aligned left as opposed to centered as currently" mean that we'll get horizontally aligned columns?

roberthouse54:
It means that if a row only has one video in it, it will move to the left side of the page, as opposed to being centered on the page. (I wasn't clear what "horizontally aligned columns" meant.)

roberthouse54:
NOTE: I modified this issue a bit. I will deal with the Safari bug in a separate PR.

static/scss/_variables.scss
```diff
@@ -1,14 +1,14 @@
...
 $icon-hover: #30a2fd;
 $fontgray: #666;
 $card-bg-color: #fff;
 $card-shadow: #c7c7c7;
-$tool-icon-color: #007de1;
-$tool-icon-color-hover: #30a2fd;
-$dropbox-blue: #007ee5;
+$tool-icon-color: #ababab;
+$tool-icon-color-hover: #007de1;
+$blue-button: #007ee5;
```
gsidebo: We can leave this, but just FYI this is the hex value that the dropbox branding guide recommended, hence the "dropbox-blue" class name

static/scss/collection.scss
```diff
@@ -158,6 +171,14 @@ $collection-tools-line-height: 36px;
...
           vertical-align: middle;
         }
+
+        i.material-icons {
+          color: $tool-icon-color;
+
+          &:hover {
+            color: $tool-icon-color-hover;
+          }
+        }
```
gsidebo: Can you combine these definitions with the definitions for `i` above? It's only there to apply styles to the icon. Result would be something like:

```scss
        i.material-icons {
          line-height: $collection-tools-line-height;
          vertical-align: middle;
          color: $tool-icon-color;

          &:hover {
            color: $tool-icon-color-hover;
          }
        }
```

PR #248, opened 2017-09-18
Added max height and inner scroll bar to dialog body

### LINKED ISSUE #205: Enable scrollbars for dialogs
Our dialog components do not have internal content-level scrollbars, and they don't create window-level scrollbars when the dialog is too tall for the browser window. This can potentially cause some significant usability problems, especially for the collection create/edit form on the collection detail page which occupies a lot of vertical space. We should either make the dialog window shrink-able and enable internal content-level scrollbars, or figure out how to get window-level scrollbars to show up.

gsidebo:
Fixes #205 

#### What's this PR do?
Adds max height and inner scroll bar to dialog body

Open a dialog and resize the window vertically. The header and footer should display as normal, and the body content section should collapse and become scrollable.

OSX defaults to hiding the internal scroll bar, which can be a UX issue (users might not know that the section is scrollable). I did a lot of tinkering to see what would get them remain visible in, but couldn't quite figure it out. We can dig a little deeper if we think it's a dealbreaker

ss 2017-09-18 at 14 35 23
ss 2017-09-18 at 14 47 34
ss 2017-09-18 at 14 47 47

### DIFF
--- a/static/scss/generalStyles.scss
+++ b/static/scss/generalStyles.scss
@@ -16,10 +16,6 @@ body {
   line-height: 1.4rem;
 }
 
-.mdc-dialog__body {
-  font-size: 1rem;
-}
-
 p.channelLink {
   margin-bottom: 0px;
 }
@@ -63,6 +59,21 @@ button {
     font-size: 18px;
     font-weight: 400;
     margin: 10px 0 7px 10px;
+  }
+}
+
+/*---------------------------------------
+   SECTION:  DIALOG STYLES
+-----------------------------------------*/
+
+.mdc-dialog {
+  .mdc-dialog__surface {
+    max-height: 90%;
+  }
+
+  .mdc-dialog__body {
+    font-size: 1rem;
+    overflow-y: scroll;
   }
 }
 


aliceriot:
unfortunately scroll bars are an OS thing, so generally we can't control their display in the browser. the only way to really do that is to use one of the JS-only scrollbar reimplmentations, and IMO that sort of solution is just covered in nope.

aliceriot:
functionality is great! reading code.

roberthouse54:
My one concern here is that the user may not realize there is more to the modal if part of the content is hidden, and the buttons are visible. I guess I am wondering if adding a drop shadow to the button bar might help a bit with that.. visually indicating that the content can scroll behind the footer? Like this...

image

gsidebo:
@roberthouse54 that's a great idea. I can create a new issue for that. Wouldn't be too difficult to implement. Do you think shadow on the footer + header or just footer?

roberthouse54:
Just the footer. I can just add this code in my current PR. So don't worry about adding an issue for it.

PR #254, opened 2017-09-19
Set logout redirect

### LINKED ISSUE #208: Redirect to login page after logging out
Logging out directs the user to `/logout` where they are shown a simple message. The user needs to manually enter a URL to log back in. We should redirect the user directly to the login page after they log out.

gsidebo:
Closes #208

#### What's this PR do?
Sets logout redirect

Logout via the drawer, make sure you are sent directly to the login page

I don't _think_ this will complicate anything with the touchstone login flow, but we may want to get some other opinions on that. The fact that we have a route defined for `/login`, which shouldn't be used with touchstone, makes me think it won't be problematic

### DIFF
--- a/ui/urls.py
+++ b/ui/urls.py
@@ -1,5 +1,7 @@
 """urls for ui"""
+from django.conf import settings
 from django.conf.urls import url, include
+from django.contrib.auth.views import logout as django_logout_view
 from rest_framework import routers
 from ui import views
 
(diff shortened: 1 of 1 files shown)


ui/urls.py
```diff
@@ -11,6 +12,7 @@
     url(r'^$', views.Index.as_view(), name='index'),
     url(r'^register/$', views.register, name='register'),
     url(r'^login/$', views.ui_login, name='login'),
+    url(r'^logout/$', django_logout_view, {'next_page': '/login/'}),
```
noisecapella: If `/login/` shouldn't be used with touchstone I think `next_page` should be something else, maybe just `/`?

PR #265, opened 2017-09-21
Fixed bug with unusable scrollbars after saving a dialog form

### LINKED ISSUE #251: Dialog form saving prevents window scroll bars
The issue that was occurring in #230 is still occurring when a form within the dialog is saved. This can be observed with the collection edit form and the video edit form (on both the collection & video detail pages). If you simply close the dialog without saving, it behaves as expected, but saving the form triggers the bug

gsidebo:
Fixes #251

#### What's this PR do?
Fixes bug with unusable scrollbars after saving a dialog form

Open a dialog with a form (edit collection, edit video), make a change, save, then try to scroll on the page after the dialog goes away.

### DIFF
--- a/static/js/components/material/Dialog.js
+++ b/static/js/components/material/Dialog.js
@@ -6,8 +6,9 @@ import Button from './Button';
 
 type DialogProps = {
   open: boolean,
-  onAccept: () => void,
-  onCancel: () => void,
+  onAccept?: () => void,
+  onCancel?: () => void,
+  hideDialog: () => void,
   children: React.Children,
   title: string,
   cancelText: string,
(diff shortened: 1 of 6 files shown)


static/js/components/material/Dialog.js
```diff
@@ -35,13 +38,19 @@ export default class Dialog extends React.Component {
...
     if (!this.dialogRoot || !this.dialogRoot.dataset) return;
 
     this.dialog = new MDCDialog(this.dialogRoot);
     this.dialog.listen('MDCDialog:accept', onAccept);
+    // Hack: If the redux action to hide the dialog is executed as part of the dialog accept
+    // handler, the dialog object doesn't tear down properly, resulting in browser scrollbars
+    // remaining inaccessible. This is fixed by attaching a second click handler to hide the
+    // dialog after 'onAccept' completes.
+    // $FlowFixMe: Flow thinks this.dialog might be null
+    this.dialog.listen('MDCDialog:accept', hideDialog || onCancel);
```
mbertrand: If onCancel is customized and does not call hideDialog  itself, the dialog won't open again after onCancel is called.  For example, I added this to the ShareDialog:

```javascript
  customCancel = (): void => {
    console.log('CANCELLED');
  };
...
    return <Dialog
      title="Share this Video"
      onCancel={this.customCancel}
```
and when the `Close` button is pressed `customCancel` is called, the dialog still closes, but can't be opened again.  This can be fixed by changing `customCancel` to call `hideDialog()`;
```javascript
  customCancel = (): void => {
    const { hideDialog } = this.props;
    console.log('CANCELLLED');
    hideDialog();
  };
```
but I think maybe the more foolproof way is to add `hideDialog` as a second click handler to the MDC `cancel` event too:
```javascript
    if (hideDialog) {
      // $FlowFixMe: Flow thinks this.dialog might be null
      this.dialog.listen('MDCDialog:cancel', hideDialog);
    }
    // $FlowFixMe: Flow thinks this.dialog might be null
    this.dialog.listen('MDCDialog:cancel', onCancel);
```

gsidebo:
@mbertrand I changed the logic for `Dialog` and the props expected. I think what we learned here is that the function to set the state in React to hide the dialog needs to be separate from the accept and cancel handlers. `onCancel` and `onAccept` are now optional props, and it's assumed that those handlers don't execute the action to hide the dialog in React

mbertrand:
👍

Issue #154, opened 2017-08-17
Send emails to all moira list admins and not only the collection owner


giocalitri:
it will be enough to modify this function:
https://github.com/mitodl/odl-video-service/blob/master/mail/tasks.py#L17-L34

Since the moira lists that we use are also mailing lists, we can just append `@mit.edu` to the list name and send an e-mail to that address.

pdpinch:
When building the mailing list, we should, if possible, union all the moira lists so as to avoid sending duplicate emails.

mbertrand:
List membership returns kerberos id's for most MIT people and not email addresses.  The code can assume that the email is `<kerberos_id>@mit.edu`, but that may not always be correct. 
There is a Moira API call to get user information but it doesn't return email address.

pdpinch:
When is <kerberos_id>@mit.edu not correct?

mbertrand:
I think it was mentioned awhile back that some people might have subdomain emails like someone@sloan.mit.edu

pdpinch:
I updated the description based on our Slack discussion.

pdpinch:
AFAIK, this is working in production.

PR #266, opened 2017-09-22
Send emails to admin moira lists

mbertrand:
#154 

#### What's this PR do?
Sends emails to the admin lists for a collection ('odl-engineering' -> 'odl-engineering@mit.edu', and also to the collection owner if not already in those lists.

NOTE: This might be difficult to test unless you already have existing moira lists with only yourself as a member, since it seems to take awhile for the list emails to work.

- Add valid values for `MAILGUN_KEY` and `MAILGUN_URL` to your .env file.
- Create a moira list with yourself as a member
- As a django user who does not have your MIT email address, but a personal email address, create/modify a collection with the above list as an admin list.
- Upload a short video
- Wait for an email sent to both your email addresses - should get 1 each.
- Repeat the above, but as a django user with your MIT email address.  You should get 1 email sent to your MIT address.

### DIFF
--- a/mail/tasks.py
+++ b/mail/tasks.py
@@ -9,7 +9,7 @@ from celery import shared_task
 from mail import api
 from mail.models import NotificationEmail
 from ui.constants import VideoStatus
-
+from ui.utils import has_common_lists
 
 log = logging.getLogger(__name__)
 
@@ -18,19 +18,17 @@ def _get_recipients_for_video(video):
     """
     Returns a list of recipients that will receive notifications for a video.
 
-    NOTE: this needs to be updated to include all the admin moira lists associated to the collection
-    where the video is.
-
     Args:
         video (ui.models.Video): a video object
 
     Returns:
         list: a list of strings representing emails
     """
-    recipients_list = []
-    owner_email = video.collection.owner.email
-    if owner_email:
-        recipients_list.append(owner_email)
+    admin_lists = video.collection.admin_lists.values_list('name', flat=True)
+    recipients_list = ['{}@mit.edu'.format(mlist) for mlist in admin_lists]
+    owner = video.collection.owner
+    if owner.email and not has_common_lists(owner, admin_lists):
+        recipients_list.append(owner.email)
     return recipients_list
 
 
--- a/mail/tasks_test.py
+++ b/mail/tasks_test.py
@@ -10,7 +10,7 @@ from django.urls import reverse
 from mail import tasks
 from mail.models import NotificationEmail
 from ui.constants import VideoStatus
-from ui.factories import VideoFactory
+from ui.factories import VideoFactory, MoiraListFactory
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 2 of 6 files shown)


mail/tasks.py
```diff
@@ -18,19 +18,17 @@ def _get_recipients_for_video(video):
...
         video (ui.models.Video): a video object
 
     Returns:
         list: a list of strings representing emails
     """
-    recipients_list = []
-    owner_email = video.collection.owner.email
-    if owner_email:
-        recipients_list.append(owner_email)
+    admin_lists = [mlist.name for mlist in video.collection.admin_lists.all()]
```
gsidebo: I think you can do this: `video.collection.admin_lists.values_list('name', flat=True)`

mbertrand: thanks

mail/tasks.py
```diff
@@ -50,6 +48,7 @@ def send_notification_email(video):
     Args:
         video (ui.models.Video): a video object
     """
+    print("SEND EMAIL!!!!!")
```
gsidebo: `print` statement :)

mbertrand:
```python
print('OOPS!')
```

mail/tasks_test.py
```diff
@@ -10,20 +10,32 @@
...
+    mocker.patch('mail.tasks.has_common_lists', return_value=False)
+
+
+def test_get_recipients_for_video(mocker):
     """
     Tests the _get_recipients_for_video api
     """
-    video = VideoFactory()
-    assert tasks._get_recipients_for_video(video) == [video.collection.owner.email]
+    lists = [MoiraListFactory(), MoiraListFactory()]
```
gsidebo: `lists = MoiraListFactory.create_batch(2)` ([doc](http://factoryboy.readthedocs.io/en/latest/reference.html#factory.create_batch))

mail/tasks_test.py
```diff
@@ -10,20 +10,32 @@
...
+    video = VideoFactory(collection=CollectionFactory(admin_lists=lists))
```
gsidebo: This _might_ work...

`video = VideoFactory(collection__admin_lists=lists)`

mail/tasks_test.py
```diff
@@ -10,20 +10,32 @@
...
+    list_emails = ['{}@mit.edu'.format(mlist.name) for mlist in video.collection.admin_lists.all()]
+    assert tasks._get_recipients_for_video(video) == list_emails + [video.collection.owner.email]
```
gsidebo: This assertion seems to depend on `has_common_lists` returning `False`. Even though you've set the default return value to `False`, I think we should explicitly set the return value before this assertion. It's not clear from within the test case that the assertion depends on that `False` value

mbertrand: OK

mail/tasks_test.py
mbertrand: OK

mail/tasks_test.py
mbertrand: Works!

gsidebo:
I've never seen any kind of admin interface for moira lists (and, obviously, haven't created new moira lists). Need instructions for that
=

mbertrand:
Go to http://web.mit.edu/moira/, and you should be able to create lists from there:
- Click 'Create a New List' and follow prompts:
  - Type of List: Moira
  - Name of List: fill in
  - Is this list a mailing list? Checked (yes)

PR #276, opened 2017-09-25
Smaller responsive video thumbnails on Collection Details page

roberthouse54:
#### What's this PR do?
This PR makes the thumbnails on the collection details page smaller, and also removes the action buttons on the video card. The video card thumbnails are different sizes depending on the browser width at 5 different screen size ranges (media queries).

Part of the goal is also to remove the "Actions" (ie. Edit and Share) from the Video Card. This PR simply hides the Actions div with display: none. A follow up issue will deal with removing the actual code for the actions and fixing all the tests, javascript, etc.

See that everything works correctly and the css is valid.

### DIFF
--- a/static/scss/collection.scss
+++ b/static/scss/collection.scss
@@ -1,4 +1,4 @@
-$video-card-width: 19%;
+$video-card-width: 15.6%;
 
 .collection-list-content,
 .collection-detail-content {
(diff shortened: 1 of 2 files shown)


roberthouse54:
This isn't quite ready for review.

static/scss/collection.scss
```diff
@@ -233,25 +246,33 @@ $video-card-width: 19%;
...
 
         img {
-          width: 100%;
+          height: 100%;
+          position: absolute;
+          margin: auto;
+          top: 0;
+          left: 0;
+          bottom: 0;
+          right: 0;
```
gsidebo: I think we can just set `top` and `left` here. I'm also not sure if `margin: auto;` is doing anything

roberthouse54: When I am testing this and I remove these things it seems to break the layout. I am testing this with thumbnails of different aspect ratios. Some are widescreen (16:9), some are standard (4:3), others are vertical, and some are extremely wide. If I remove the margin, and left and right attributes, the 4:3 videos are no longer centered. Did you test this with different aspect ratios?

PR #295, opened 2017-10-04
Add validation for MAILGUN_URL and MAILGUN_KEY

noisecapella:
Related to https://github.com/mitodl/micromasters/issues/3599

#### What's this PR do?
Adds validation for `MAILGUN_URL` and `MAILGUN_KEY` and removes the default value which will need to be provided via environment variable instead

Make sure your `.env` includes the same value as the value used in the default in `settings.py`

### DIFF
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -289,8 +289,12 @@ EMAIL_USE_TLS = get_bool('ODL_VIDEO_EMAIL_TLS', False)
 EMAIL_SUPPORT = get_string('ODL_VIDEO_SUPPORT_EMAIL', 'support@example.com')
 DEFAULT_FROM_EMAIL = get_string('ODL_VIDEO_FROM_EMAIL', 'webmaster@localhost')
 
-MAILGUN_URL = get_string('MAILGUN_URL', 'https://api.mailgun.net/v3/video.odl.mit.edu')
+MAILGUN_URL = get_string('MAILGUN_URL', None)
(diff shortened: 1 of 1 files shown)


odl_video/settings.py
```diff
@@ -289,8 +289,12 @@ def get_all_config_keys():
 EMAIL_SUPPORT = get_string('ODL_VIDEO_SUPPORT_EMAIL', 'support@example.com')
 DEFAULT_FROM_EMAIL = get_string('ODL_VIDEO_FROM_EMAIL', 'webmaster@localhost')
 
-MAILGUN_URL = get_string('MAILGUN_URL', 'https://api.mailgun.net/v3/video.odl.mit.edu')
+MAILGUN_URL = get_string('MAILGUN_URL', None)
+if not MAILGUN_URL:
```
mbertrand: I think these checks might not be necessary because `MAILGUN_URL` and `MAILGUN_KEY` are included in the `MANDATORY_SETTINGS` at the bottom of settings, and are checked in [apps.py](https://github.com/mitodl/odl-video-service/blob/4ad7018b930221c67b1b8e69e5a0d7986f2cef16/ui/apps.py#L18)

noisecapella: Ah, you're right

noisecapella: Actually we should still remove the default MAILGUN_URL here and move it into .env, but I'll open another PR for that

PR #297, opened 2017-10-04
Multiangle VideoJS

### LINKED ISSUE #283: multiangle playback in VideoJS

mbertrand:
Closes #283

#### What's this PR do?
Enables multiangle playback with Video.js and removes the USwitch player.

- In the admin console, set a video to multiangle=True (ideally one that really is multiangle, but it's not necessary) - one is available with subtitles on [dropbox](https://www.dropbox.com/sh/1dk25zf882skcrc/AAAb8xCJad7nJgsGMirAvbTra?dl=0).
- Go to the video detail page for that video.  The video player should show one corner of the video, with 4 smaller screens to the right showing each angle of the video, with the one currently playing highlighted in red.
- Turn subtitles on. They should be visible/readable.
- Click on one of the other screens.  It should become highlighted in red and the main video should switch to that screen view.  Subtitles should still be on.
- Click the full screen icon.  Both the main video screen and 4 corner screens should be viewable in fullscreen mode, and clicking on each corner screen should still function the same.
- Either click the fullscreen icon again or press the Esc key.  The page should return to it's previous state.
- Resize the browser window.  The main video window and 4 corner screens should adapt in size.
- Try the above in different browsers.
- Load a video detail page for a video that is not multi-angle.  It should display normally (no corner screens, entire video width/height visible).


noisecapella:
I'm seeing this warning:

screenshot from 2017-10-06 11-38-57

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+
+  cropVideo() {
+    const { selectedCorner } = this.props;
+    if (selectedCorner) {
+      const shiftX = canvases[selectedCorner].shiftX;
+      const shiftY = canvases[selectedCorner].shiftY;
+      const transformProps = ['transform', 'WebkitTransform', 'MozTransform', 'msTransform', 'OTransform'];
+      let prop = transformProps[0];
+
+      transformProps.some((property: string) => {
```
noisecapella: Can you change this to a `find` to be more explicit: `prop = transformProps.find(property => this.player.el_.style[prop] !== undefined)`?

Also the `typeof variable === 'undefined'` is not necessary anymore because since ES5 it always exists and cannot be overridden: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+        prop = property;
+        return (typeof this.player.el_.style[prop] !== 'undefined');
+      });
+
+      const aspectRatio = (this.player.videoWidth() / this.player.videoHeight());
+      let videoWidth = Math.min(
```
noisecapella: Wouldn't the video width always be the minimum of these values?

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+        parseInt(window.getComputedStyle(this.videoNode).maxHeight) * aspectRatio,
+        window.innerWidth,
+        screen.width
+      );
+      if (isNaN(videoWidth) || this.isFullscreen()) {
+        videoWidth = Math.min(window.innerWidth, screen.width);
+      }
+      const canvasWidth = Math.floor(videoWidth / 4);
+      videoWidth = Math.floor(videoWidth - (canvasWidth - (canvasWidth / aspectRatio / 3)));
```
noisecapella: What does `canvasWidth / aspectRatio` do? It seems like it would get a height value but I'm not sure what it's for in the context of this expression, since all other values are widths

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+      this.videoContainer.style.maxWidth = `${videoWidth}px`;
+      this.videoContainer.parentElement.style.width = `${videoWidth + canvasWidth}px`;
+
+      const left = Math.round(this.player.currentWidth() / (shiftX ? -2 : 2));
```
noisecapella: Can you explain this a bit? I'm not sure what the `2` values are for here, or the `scale(2)` below

static/js/components/VideoPlayer_test.js
```diff
@@ -1,49 +1,152 @@
...
+    video.multiangle = true;
+    sandbox.stub(window, 'getComputedStyle').returns({maxHeight: 600});
+    let wrapper = renderPlayer();
+    wrapper.instance().player = playerStub;
+    wrapper.instance().videoNode = nodeStub;
+    let canvas = wrapper.find('.camera-box').at(0);
+    assert.throws(() => wrapper.instance().drawCanvas(canvas, true, false), TypeError, 'getContext');
+  });
+
+  it('subtitles added to player if present', () => {
```
noisecapella: Can you add a test where the player has two existing tracks and there are two subtitles coming from `video.videosubtitle_set`, one which is the same as the existing one and one which is new? It would check that we add, remove, and keep existing subtitles appropriately

static/js/containers/VideoDetailPage.js
```diff
@@ -95,7 +96,13 @@ class VideoDetailPage extends React.Component {
     this.uploadVideoSubtitle();
   };
 
+  updateCorner = async (corner) => {
```
noisecapella: Can you add a type for `corner`?

static/js/reducers/videoUi.js
```diff
@@ -20,10 +22,11 @@ export const INITIAL_UPLOAD_SUBTITLE_FORM_STATE = {
 };
 
 export const INITIAL_UI_STATE = {
-  videoSubtitleForm: INITIAL_UPLOAD_SUBTITLE_FORM_STATE
+  videoSubtitleForm: INITIAL_UPLOAD_SUBTITLE_FORM_STATE,
+  corner: 'upperLeft'
```
noisecapella: Can you make these values (`upperLeft` and etc) into constants and put them in `constants.js`?

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+  lowerLeft: {shiftX: false, shiftY: true},
+  lowerRight: {shiftX: true, shiftY: true}
+};
 
 export default class VideoPlayer extends React.Component {
   props: {
     video: Video,
-    useIframeForUSwitch: boolean,
+    selectedCorner?: string,
+    cornerFunc?: Function
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
   };
 
-  render() {
-    const { video, useIframeForUSwitch } = this.props;
-    const embedUrl = makeEmbedUrl(video.key);
+  player: Object;
+  videoNode: Object;
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
-      } else {
-        return <USwitchPlayer video={video} />;
+  updateSubtitles() {
+    const { video } = this.props;
+    if (this.player) {
+      // Remove existing tracks for deleted subtitles
+      const tracks = this.player.textTracks();
+      let subtitleUrls = video.videosubtitle_set.map((subtitle: VideoSubtitle) => (makeVideoSubtitleUrl(subtitle)));
+      let trackUrls = [];
+      for (let idx = 0; idx < tracks.length; idx++) {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+        if (this.player && tracks[idx] && !subtitleUrls.includes(tracks[idx].src)) {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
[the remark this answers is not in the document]
mbertrand: Actually that doesn't work:
`Uncaught (in promise) TypeError: tracks[Symbol.iterator] is not a function`

static/js/components/VideoPlayer.js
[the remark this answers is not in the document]
mbertrand: Removed `this.player` check but keeping `tracks[idx]`  because `let track of tracks` throws error.

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+          this.player.removeRemoteTextTrack(tracks[idx]);
+        } else {
+          trackUrls.push(tracks[idx].src);
+        }
       }
+      // Add tracks for any new subtitles associated with the video
+      video.videosubtitle_set.map((subtitle: VideoSubtitle) => {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+            srcLang: subtitle.language,
+            label: subtitle.language_name
+          });
+        }
+      });
+    }
+  }
+
+  drawCanvas(canvas: HTMLCanvasElement, shiftX: boolean, shiftY: boolean) {
+    function drawCanvasImage(canvas, videoNode, shiftX, shiftY) {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+    canvas.height = Math.floor(this.videoNode.offsetHeight / 4) - 2;
+    if (canvas && this.videoNode) {
+      drawCanvasImage(canvas, this.videoNode, shiftX, shiftY);
+    }
+  }
+
+  configureCameras() {
+    if (this.cameras) {
+      const canvasElements = this.cameras.getElementsByTagName("canvas");
+      Object.keys(canvases).map(corner => {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,225 @@
...
+  }
+
+  isFullscreen() {
+    // $FlowFixMe
+    return document[FULLSCREEN_API.fullscreenElement];
+  }
+
+  cropVideo() {
+    const { selectedCorner } = this.props;
+    if (selectedCorner) {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/components/VideoPlayer.js
mbertrand: OK

static/js/components/VideoPlayer.js
mbertrand: This prevents black bars from showing up on one side or the other. For example without including `parseInt(window.getComputedStyle(this.videoNode).maxHeight) * aspectRatio`, the following can appear:

 
screen shot 2017-10-06 at 1 53 02 pm

static/js/components/VideoPlayer.js
mbertrand: This was the result of some trial & error in determining how to correctly size both the video and the 4 canvas boxes on the right so that they were rendered with about equal height while also taking up 100% of the available container (especially in fullscreen mode).

static/js/components/VideoPlayer.js
mbertrand: `scale(2)` is what causes the main video screen to zoom in to a corner. The 2 or -2 for `left`, `top` determines in which direction the left and top origins should be shifted to display the correct corner.

static/js/util/fullscreen_api.js
```diff
@@ -0,0 +1,70 @@
...
+    'msFullscreenEnabled',
+    'MSFullscreenChange',
+    'MSFullscreenError'
+  ]
+];
+
+const specApi = apiMap[0];
+
+
+for (let i = 0; i < apiMap.length; i++) {
```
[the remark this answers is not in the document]
mbertrand: OK

static/js/containers/VideoDetailPage.js
mbertrand: OK

static/js/reducers/videoUi.js
mbertrand: OK, I put this in `constants.py`:

```python
export const CANVASES = {
  upperLeft: {shiftX: false, shiftY: false},
  upperRight: {shiftX: true, shiftY: false},
  lowerLeft: {shiftX: false, shiftY: true},
  lowerRight: {shiftX: true, shiftY: true}
};
```

So `videoUI.js` above would be changed to `corner: CANVASES[upperLeft]`:

static/js/components/VideoPlayer_test.js
mbertrand: OK

static/js/components/VideoPlayer.js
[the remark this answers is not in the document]
noisecapella: That's strange, I'm not sure why they aren't just returning an array here. Let's leave it as is

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,216 @@
...
+          // $FlowFixMe - corner does not have to be a number
+          canvasElements[corner],
+          CANVASES[corner].shiftX,
+          CANVASES[corner].shiftY
+        );
+      });
+    }
+  }
+
+  isFullscreen() {
```
noisecapella: Can you make this a static function?

static/js/containers/VideoDetailPage.js
```diff
@@ -95,7 +96,13 @@ class VideoDetailPage extends React.Component {
     this.uploadVideoSubtitle();
   };
 
+  updateCorner = async (corner: string) => {
```
noisecapella: I don't think this needs to be an async function since `dispatch(updateVideoJsSync(corner))` isn't returning a `Promise`. Can you remove the async here and change the type signature to return `void`?

static/js/components/VideoPlayer.js
```diff
@@ -1,37 +1,216 @@
...
+const drawCanvasImage = function(canvas, videoNode, shiftX, shiftY) {
+  const x = shiftX ? Math.floor(videoNode.videoWidth / 2) : 0;
+  const y = shiftY ? Math.floor(videoNode.videoHeight / 2) : 0;
+  const context = canvas.getContext('2d');
+  context.drawImage(
+    videoNode,
+    x, y, Math.floor(videoNode.videoWidth / 2), Math.floor(videoNode.videoHeight / 2),
+    0, 0, canvas.width, canvas.height
+  );
+  setTimeout(drawCanvasImage, 20, canvas, videoNode, shiftX, shiftY);
```
noisecapella: [This](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) says that `setTimeout` can't be used this way in IE9 and below. Do you know if we support any of those browsers for this app?

static/js/components/VideoPlayer.js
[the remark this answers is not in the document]
noisecapella: It looks like `this.player` is still here

static/js/components/VideoPlayer.js
noisecapella: Can you add this as a comment in the code?

static/js/components/VideoPlayer.js
mbertrand: @pdpinch does OVS need to support IE9 and below?

static/js/components/VideoPlayer.js
mbertrand: Sure

static/js/containers/VideoDetailPage.js
mbertrand: OK

static/js/components/VideoPlayer.js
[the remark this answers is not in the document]
mbertrand: Oops

static/js/components/VideoPlayer.js
mbertrand: @noisecapella that links says a polyfill can make this work in IE, maybe `babel-polyfill` does that automatically? - https://github.com/zloirock/core-js#settimeout--setinterval

noisecapella: Yes I think you're right

PR #309, opened 2017-10-12
Raise custom 'MoiraException' on any moira-related error

### LINKED ISSUE #307: Raise exceptions when moira fails
We're getting a fair number of `OSError: write error` exceptions in Sentry. Many, but not all of these seem to be caused by Moira failures and/or timeouts. 

We should add specific error handling for Moira errors so we can track how common they are and determine if we need to handle them differently. For now I'd be satisfied to catch these Moira errors and log them with a unique message so they don't get lumped with other `OSError: write error` messages. 

https://sentry.io/mit-office-of-digital-learning/odl-video-service/issues/349208356/

```
OSError: write error

OSError: write error
```

mbertrand:
Closes #307

#### What's this PR do?
Raises a `MoiraException` if any error occurs when making calls to the Moira API

- Disconnect wifi
- Log in to OVS using a non-superuser account
- Go to any video detail URL
- You should get a 'MoiraException' the page:
```
MoiraException at /videos/b6491433151144e080e4769c823bdd88/
('Something went wrong with moira client for user %s: %s', 'mattbert@mit.edu', 
'(\'Something went wrong with moira client: %s\', 
"HTTPSConnectionPool(host=\'moiraws.mit.edu\', port=443): Max retries exceeded with url: /moiraws/services/moira?wsdl 
(Caused by NewConnectionError(\'<urllib3.connection.VerifiedHTTPSConnection object at 0x7f78205aca20>: 
Failed to establish a new connection: [Errno -2] Name or service not known\',))")')
```

### DIFF
--- a/ui/exceptions.py
+++ b/ui/exceptions.py
@@ -1,0 +1,5 @@
+""" Custom exceptions"""
+
+
+class MoiraException(Exception):
+    """Custom exception to be used when something goes wrong with Moira API calls"""
--- a/ui/models.py
+++ b/ui/models.py
@@ -16,6 +16,7 @@ from mail import tasks
 from ui import utils
 from ui.constants import VideoStatus
 from ui.encodings import EncodingNames
+from ui.exceptions import MoiraException
 from ui.tasks import delete_s3_objects
 
 TRANSCODE_PREFIX = 'transcoded'
@@ -35,7 +36,10 @@ class MoiraList(models.Model):
 
         """
         moira = utils.get_moira_client()
-        return set(moira.list_members(self.name, type=''))
+        try:
+            return set(moira.list_members(self.name, type=''))
+        except Exception as exc:
+            raise MoiraException('Something went wrong with getting moira-list members') from exc
 
     def __str__(self):
         return self.name
--- a/ui/utils.py
+++ b/ui/utils.py
@@ -10,6 +10,8 @@ from django.conf import settings
 
 from mit_moira import Moira
 
+from ui.exceptions import MoiraException
+
 log = logging.getLogger(__name__)
 
 MoiraUser = namedtuple('MoiraUser', 'username type')
@@ -31,7 +33,10 @@ def get_moira_client():
             )
     if errors:
         raise RuntimeError('\n'.join(errors))
-    return Moira(settings.MIT_WS_CERTIFICATE_FILE, settings.MIT_WS_PRIVATE_KEY_FILE)
+    try:
+        return Moira(settings.MIT_WS_CERTIFICATE_FILE, settings.MIT_WS_PRIVATE_KEY_FILE)
+    except Exception as exc:  # pylint: disable=broad-except
+        raise MoiraException('Something went wrong with creating a moira client') from exc
 
 
 def get_moira_user(user):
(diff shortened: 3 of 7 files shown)


ui/exceptions.py
```diff
@@ -0,0 +1,6 @@
+""" Custom exceptions"""
+
+
+class MoiraException(Exception):
+    """Custom exception to be used when something goes wrong with Moira API calls"""
+    pass
```
noisecapella: Can you remove the `pass` here? It's not necessary since the docstring does the same thing here

ui/models.py
```diff
@@ -35,7 +36,10 @@ def members(self):
 
         """
         moira = utils.get_moira_client()
-        return set(moira.list_members(self.name, type=''))
+        try:
+            return set(moira.list_members(self.name, type=''))
+        except Exception as exc:
+            raise MoiraException('Something went wrong with getting moira-list members: %s', str(exc))
```
noisecapella: I don't think the `%s` syntax works here, and it's not necessary to attach the two exceptions. Can you do this instead:

    except Exception as exc:
        raise MoiraException('Something went wrong with getting moira-list members') from exc

ui/utils.py
```diff
@@ -31,7 +33,10 @@ def get_moira_client():
             )
     if errors:
         raise RuntimeError('\n'.join(errors))
-    return Moira(settings.MIT_WS_CERTIFICATE_FILE, settings.MIT_WS_PRIVATE_KEY_FILE)
+    try:
+        return Moira(settings.MIT_WS_CERTIFICATE_FILE, settings.MIT_WS_PRIVATE_KEY_FILE)
+    except Exception as exc:  # pylint: disable=broad-except
+        raise MoiraException('Something went wrong with creating a moira client: %s', str(exc))
```
noisecapella: See the exception chaining comment above

ui/utils.py
```diff
@@ -62,12 +67,11 @@ def user_moira_lists(user):
...
     """
     moira_user = get_moira_user(user)
+    moira = get_moira_client()
     try:
-        moira = get_moira_client()
         return moira.user_lists(moira_user.username, moira_user.type)
     except Exception as exc:  # pylint: disable=broad-except
-        log.exception('Something went wrong with the moira client for user %s: %s', user.email, str(exc))
-        return []
+        raise MoiraException('Something went wrong with getting moira lists for user %s: %s', user.email, str(exc))
```
noisecapella: See exception chaining comment above. To put the user email in the exception message you'll need to use `.format()` or use a format string

ui/utils_test.py
```diff
@@ -64,7 +63,7 @@ def test_user_moira_lists_error(mock_moira_client):
     """
     Test that an empty list is returned if the Moira client can't connect
```
noisecapella: Can you update this text?

ui/exceptions.py
mbertrand: OK

ui/models.py
mbertrand: OK

ui/utils.py
mbertrand: OK

ui/utils.py
mbertrand: OK

ui/utils_test.py
mbertrand: OK

PR #312, opened 2017-10-17
Remove login and registration

### LINKED ISSUE #311: Remove login and register views
We should remove the login and register views since they are for debugging use only and we can have similar functionality by logging in via Django admin and manipulating status via Django shell

noisecapella:
Fixes #311 

#### What's this PR do?
Removes login and register views

Go to `/`. It should redirect you to `/collections/` which should require you to login via Django admin if necessary.


noisecapella:
Thanks, that's the conclusion I came to too

PR #318, opened 2017-10-18
Upgrade node.js and yarn

noisecapella:
None

#### What's this PR do?
Upgrade yarn to 1.2.1 and node.js to 8.5

Nothing should break

### DIFF
--- a/Dockerfile-node
+++ b/Dockerfile-node
@@ -1,9 +1,7 @@
-FROM node:6.2
+FROM node:8.5
 LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
 
-RUN apt-get update && apt-get install libelf1
-
-RUN adduser --disabled-password --gecos "" mitodl
+RUN apt-get update && apt-get install libelf1 -y
 
 COPY package.json /src/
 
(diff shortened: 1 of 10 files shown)


static/js/global_init.js
```diff
@@ -24,6 +25,18 @@ if (!Object.entries) {
   entries.shim();
 }
 
+let sandbox;
+before(() => { // eslint-disable-line mocha/no-top-level-hooks
+  sandbox = sinon.sandbox.create();
+  sandbox.stub(HTMLCanvasElement.prototype, 'getContext').returns({
```
mbertrand: Just curious why this needs to be here in `global_init.js` and not `VideoPlayer_test.js`.  I moved this code block there to see if it would work and tests failed, but don't quite understand why.

Dockerfile-node
```diff
@@ -1,18 +1,16 @@
...
 RUN node /src/scripts/install_yarn.js
 
-RUN mkdir -p /home/mitodl/.cache/yarn
+RUN mkdir -p /home/node/.cache/yarn
 
-RUN chown mitodl:mitodl /home/mitodl/.cache/yarn
+RUN chown node:node /home/node/.cache/yarn
 
-USER mitodl
+USER node
```
mbertrand: Any particular reason for change of user?

noisecapella: Yes, in the updated Docker container there is an existing `node` user which has the 1000 uid, which is the default for non-system users. The uid must match between the users for the docker volume sharing to let the desktop user edit files from the docker user and vice versa.

static/js/global_init.js
noisecapella: I wanted to have the mock in place for both `VideoPlayer` and `VideoSubtitleCard`. Regarding `before` instead of `beforeEach`, I wanted it to handle the `setTimeout` callback which might execute between tests. It is kind of hacky but I didn't want to edit the functionality in this PR, just the tests

Issue #119, opened 2017-08-09
Tweak VideoJS CSS to prevent black bars


mbertrand:
Currently black bars may appear on the top/bottom or left/right of the video depending on the browser window width.  This can be fixed by tweaking the HTML and CSS:

in `video_detail.html`, add a wrapper around the `odl-video-player` div:
```
        <div class="wrapper">
         <div id="odl-video-player" class="video-odl-medium"></div>
        </div>
```

in `layout.scss`, remove videojs styles for everything but `video-odl-medium` and change that to:
```
.video-odl-medium{
    width:80%;
    max-width:1024px;
}
```

in `static/js/entry/video.js`,  set `fluid: true`

pdpinch:
one idea: just set the "bars" to be white, same as the page background.

mbertrand:
I tried the white/transparent background idea first but then things didn't look quite right with the menu bar at the bottom.

pdpinch:
I believe this is fixed already.

PR #339, opened 2017-10-31
This adds back black background behind video

roberthouse54:
#### What's this PR do?
The black full-width background behind videos was lost as part of a recent change. This adds it back.

See that it looks ok.

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -234,31 +234,33 @@ export default class VideoPlayer extends React.Component {
   render() {
     const { video, selectedCorner } = this.props;
     return (
-      <div className="video-odl-center">
-        <div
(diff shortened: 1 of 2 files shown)


pdpinch:
Didn't we remove the black bars on purpose, in https://github.com/mitodl/odl-video-service/issues/119 ?

roberthouse54:
I don't actually recall this. I didn't think we did. The black bars have always been a part of the design. I thought this was perhaps referencing something different. Matt told me we lost the black bars recently when he had to mess around with the sizing and layout of the multi-angle videos.

I am not clear what happened with that #119.

noisecapella:
If I resize the browser window to shrink the video the black bars remain on the sides. Is that intentional?

screenshot from 2017-10-31 15-44-08

roberthouse54:
No it is not. OK, let's put this on hold.

PR #376, opened 2017-11-28
Sync settings with cookiecutter

noisecapella:
None

#### What's this PR do?
- Uses `SECRET_KEY` environment variable instead of `DJANGO_SECRET_KEY`
 - Rename `ODL_VIDEO_SERVICE_USE_WEBPACK_DEV_SERVER` to `ODL_VIDEO_USE_WEBPACK_DEV_SERVER` for consistency with cookiecutter
 - Adds `SECURE_SSL_REDIRECT` from cookiecutter. This is necessary to force SSL use for end users
 - Move various settings around to make future comparisons with cookiecutter easier
 - Add OS X syslog settings from cookiecutter
 - Remove password validators which are unused

Nothing should be different


noisecapella:
I rebased

PR #392, opened 2017-12-20
Fix Moira-related issues

### LINKED ISSUE #370: Some emails cause Moira to throw exceptions when queried
Some non-MIT emails cause java NullPointerExceptions when querying Moira for that user's lists:

[redacted]

### LINKED ISSUE #391: Determine cause of Moira exception 'no list name found'
https://sentry.io/mit-office-of-digital-learning/odl-video-service/issues/420554208/

Dangling commas in a moira list field will cause the app to try to process a moira list with a blank name, which throws an exception.

mbertrand:
Fixes #370 
Fixes #391

#### What's this PR do?
- Ignores the exception thrown by Moira when retrieving moira lists for a user that isn't on any lists
- Prevents moira lists with blank names from being processed (dangling or multiple commas on front-end form).

- From a collection edit dialog, enter 1 or more real moira list names with extra commas at the beginning, end, and/or between the names.  Click the save button.  The dialog should close.  Open it again, and you should see the moira list names without the extra commas.
- From the command line:
   ```python
   from django.contrib.auth.models import User
   from ui.utils import user_moira_lists
   user, _ = User.objects.get_or_create(email='<listless_email>', defaults={'username': '<listless_email>'})
   user_moira_lists(user)
   ```
   This should return an empty list and not throw an exception.  A list of emails that currently throw exceptions can be found in the comments at https://sentry.io/mit-office-of-digital-learning/odl-video-service/issues/380406074/activity/

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -19,3 +19,5 @@ requests-mock
 semantic-version
 tox
 moto
+# Fixes problem with pytest on travis
+astroid==1.5.3
--- a/static/js/util/util.js
+++ b/static/js/util/util.js
@@ -18,5 +18,5 @@ export const wait = (millis: number): Promise<void> =>
 export const calculateListPermissionValue = (choice: string, listsInput: ?string): Array<string> => (
   choice !== PERM_CHOICE_LISTS || !listsInput || listsInput.trim().length === 0
     ? []
-    : R.map(R.trim, R.split(',', listsInput))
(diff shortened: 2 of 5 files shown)


test_requirements.txt
```diff
@@ -19,3 +19,4 @@ requests-mock
 semantic-version
 tox
 moto
+astroid==1.5.3
```
noisecapella: Can you add a comment explaining the reason for this?

PR #395, opened 2017-12-28
Use unique s3 key for each subtitle upload to avoid Cloudfront caching issue

### LINKED ISSUE #303: Old subtitles are shown when new subtitle file is uploaded
To reproduce:
 - Upload subtitles to a video. Verify that you see the subtitles
 - Delete the subtitles. Upload a different set of subtitles. View the subtitles of the view. They should be the new ones but the old ones show up instead.

mbertrand:
Fixes #303 

#### What's this PR do?
Gives a unique S3 key (and consequently Cloudfront URL) based on datetime for each uploaded subtitle, to avoid caching issues which cause an old subtitle file to still be used for awhile after is is replaced by another.

Upload a subtitles file and display the subtitles.  Delete it, upload a different subtitle file, and turn subtitles on again.  The new subtitles should be displayed, not the old ones.

### DIFF
--- a/static/js/components/VideoSubtitleCard.js
+++ b/static/js/components/VideoSubtitleCard.js
@@ -27,7 +27,7 @@ export default class VideoSubtitleCard extends React.Component {
             {video.videosubtitle_set.map((subtitle: VideoSubtitle, key) => (
               <div className="mdc-list-item" key={key}>
                 <span className="video-subtitle-filename">
-                  { subtitle.s3_object_key.split('/').slice(-1)}
+                  {subtitle.s3_object_key.split('/').slice(-1)[0].slice(0,10)}..._{subtitle.language}.vtt
                 </span>
                 <span className="video-subtitle-language">({subtitle.language_name})</span>
                 <span className="video-subtitle-button">
--- a/ui/models.py
+++ b/ui/models.py
@@ -210,16 +210,21 @@ class Video(models.Model):
         basename, _ = os.path.splitext(original_s3_key)
         return output_template.format(prefix=TRANSCODE_PREFIX, s3key=basename, preset=preset)
 
-    def subtitle_key(self, language='en'):
+    def subtitle_key(self, dttm, language='en'):
         """
         Returns an S3 object key to be used for a subtitle file
         Args:
             language(str): 2-letter language code
+            dttm(DateTime): a DateTime object
 
         Returns:
             str: S3 object key
         """
-        return 'subtitles/{}/subtitle_{}.vtt'.format(self.hexkey, language)
+        return 'subtitles/{key}/subtitles_{key}_{dt}_{lang}.vtt'.format(
+            key=self.hexkey,
+            dt=dttm.strftime('%Y%m%d%H%M%S'),
+            lang=language
+        )
 
     def update_status(self, status):
         """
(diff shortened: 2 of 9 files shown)


noisecapella:
Not sure if this is an issue but I can upload two subtitles at the same time. Is this expected behavior?

screenshot from 2017-12-29 14-52-38

roberthouse54:
I am not sure if this is expected behavior to be honest. I think it is possible to have multiple subtitle files a user can choose from (in a dropdown menu). The normal use case for this would be different languages. Matt can comment on what is supported, then we can figure out how to proceed.

mbertrand:
There should probably only be one subtitle allowed per language, and right now any uploaded subtitle is assumed to be English.  Will fix.

ui/models.py
```diff
@@ -219,7 +221,11 @@ def subtitle_key(self, language='en'):
         Returns:
             str: S3 object key
         """
-        return 'subtitles/{}/subtitle_{}.vtt'.format(self.hexkey, language)
+        return 'subtitles/{key}/subtitles_{key}_{dt}_{lang}.vtt'.format(
+            key=self.hexkey,
+            dt=datetime.now(tz=pytz.UTC).strftime('%Y%m%d%H%M%S'),
```
noisecapella: If the particular format doesn't matter here, can we use `datetime.isoformat()` instead? [ISO-8601](https://xkcd.com/1179/) should also have the properly where it can be sorted and it is ordered by time

ui/models.py
```diff
@@ -219,7 +221,11 @@ def subtitle_key(self, language='en'):
         Returns:
             str: S3 object key
         """
-        return 'subtitles/{}/subtitle_{}.vtt'.format(self.hexkey, language)
+        return 'subtitles/{key}/subtitles_{key}_{dt}_{lang}.vtt'.format(
```
noisecapella: Can you pass in `datetime.now(tz=pytz.UTC)` as an argument so that the function is more deterministic?

static/js/components/VideoSubtitleCard.js
```diff
@@ -27,7 +28,7 @@ export default class VideoSubtitleCard extends React.Component {
             {video.videosubtitle_set.map((subtitle: VideoSubtitle, key) => (
               <div className="mdc-list-item" key={key}>
                 <span className="video-subtitle-filename">
-                  { subtitle.s3_object_key.split('/').slice(-1)}
+                  {_.split(subtitle.s3_object_key, ('/')).slice(-1)[0].slice(0,10)}..._{subtitle.language}.vtt
```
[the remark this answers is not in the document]
mbertrand: Nothing, but in some past reviews I was asked to use the lodash function instead of a native equivalents.  I'm fine with either one.

ui/models.py
mbertrand: That format doesn't play nice with S3/Cloudfront (it will upload fine, but then can't be downloaded or displayed in video) - my guess is that the colon is the issue.  I'd prefer to keep the date part purely numeric, and include the video hex key too because it might be somewhat useful in figuring out which video it belongs to after downloading.

ui/models.py
mbertrand: ✅

mbertrand:
I changed the `<a>` tag to a `<button>` tag and modified the style it uses to be more generic (not tag-specific).

PR #404, opened 2018-01-10
Refactor video analytics collection

### LINKED ISSUE #401: Update GA events to allow vizualization
In order to present and visualize analytics data to collection admins (#402) we need to make sure we are collecting the necessary event data. 

### acceptance criteria 
- [x] fire events every 1m (60s)
- [x] `eventCategory` set to "video"
- [x] `eventAction` set to a string with the format "Ttttt" where tttt is the minute of the video when the event is fired 
- [x] `eventLabel` set to the video uuid
- [x] `customDimension1` set to the camera angle (if this works)

mbertrand:
Fixes #401 

#### What's this PR do?
- Makes all event labels == the video key
- Video playback progress is sent every 60 seconds with a `Tdddd` event (where `dddd` equals the minute with up to 3 preceding 0's).
- Records the camera angle of each event using a custom dimension (needs to be created manually in Google Analytics).
- Adds a new .env variable / django setting `GA_DIMENSION_CAMERA` that should equal the custom dimension (`"dimension1"` unless other custom dimensions already exist).
- Changes the camera angle names from `upperLeft`, `lowerRight` etc to `camera1` through `camera4`

- Set up a google analytics account if necessary
- Create a custom dimension for camera angle in GA:
  
screen shot 2018-01-10 at 1 15 18 pm

- Set `GA_TRACKING_ID` in your `.env` file to the appropriate value
- Set `GA_DIMENSION_CAMERA` in your `.env` file to `"dimension1"`
- Open a detail page for a multiangle video with the network developer tab open and start playing the video.  You should see `collect` requests sent to Google Analytics whenever you change the camera angle, play, pause, stop, etc and every time an additional minute of video is played.  The data sent with each request should include:
    - `label`: `<video.key>`
    - `cd1`: `camera<number>`
    - `ea`: `<name of event>`
- After a few minutes, log in to Google Analytics and see if the data is present (the custom dimension should be available as a `Secondary dimension`)
- Repeat some of the above with a non-multiangle video.  Everything should be about the same except that the `cd1` parameter shouldn't be sent.


ui/views_test.py
```diff
@@ -105,6 +105,7 @@ def test_video_detail(logged_in_client, settings):
     assert js_settings_json == {
         'editable': True,
         "gaTrackingID": settings.GA_TRACKING_ID,
+        "ga_dimension_camera": settings.GA_DIMENSION_CAMERA,
```
noisecapella: Can you set `settings.GA_TRACKING_ID` and `settings.GA_DIMENSION_CAMERA` at the top of the test so that it doesn't use the default of `""`?

static/js/util/google_analytics.js
```diff
@@ -37,3 +37,11 @@ export const initGA = () => {
     ga.initialize(SETTINGS.gaTrackingID, { debug: debug })
   }
 }
+
+export const setCustomDimension = (dimension: string, value: string) => {
+  if (dimension) {
+    const dimensionObject = {}
```
noisecapella: In ES6 you can specify keys like this: `const dimensionObject = { [dimension]: value }`

static/js/components/VideoPlayer.js
```diff
@@ -109,8 +110,8 @@ export default class VideoPlayer extends React.Component<*, void> {
             tracks[idx].addEventListener("modechange", function() {
               sendGAEvent(
                 "video",
-                "modechange",
-                `Subtitles for ${this.label} ${this.mode}`,
+                `Subtitles ${this.label} ${this.mode}`,
```
noisecapella: Is it intentional to swap these two arguments? What do `action` and `label` mean for google analytics?

static/js/components/VideoPlayer.js
```diff
@@ -247,20 +251,21 @@ export default class VideoPlayer extends React.Component<*, void> {
         "FullscreenToggle"
       ).prototype.handleClick = toggleFullscreen
     }
-    this.percentTracked = []
+    this.lastMinuteTracked = -1
```
noisecapella: Can you use `null` here instead to make it more explicit when the value is not set?

static/js/components/VideoPlayer.js
mbertrand: Intentional, now the label should always be equal to `video.key` for all video events including this one, and the `Subtitles...` string should be the event.

static/js/util/google_analytics.js
mbertrand: thanks!

static/js/components/VideoPlayer.js

ui/views_test.py
mbertrand: ✅

PR #429, opened 2018-02-01
TechTV import

### LINKED ISSUE #414: migration script to create user accounts
When migrating TechTV collections to OVS, we will need to create accounts for collection owners on OVS. 

If the TechTV collection owner has an @mit.edu email address, we can create an account for them. It's safe to assume that they will be able to log in through Touchstone. 

If the TechTV collection owner does not have an @mit.edu email address, we can't assume they have a Touchstone Collaboration account. For these collections, we will set the collection owner to the `LECTURE_CAPTURE_USER` (i.e. @emmello)

### LINKED ISSUE #415: migration script to create collections, videos from techTV data
### acceptance criteria
- [x] create private collection in OVS for each protected collection in TechTV
- [x] migrate TechTV collection id, name, description, permissions
   we will flatten TechTV collection hierarchy, including any necessary calculation of permissions
- [x] migrating permissions will require creating corresponding moira list objects in OVS
- [x] migrate video id, name, description, file locations (anything else?)

### questions
- [ ] Do we need to address nested moira lists in order for this to work?

### LINKED ISSUE #416: migrate caption files and convert SRT to VTT

### LINKED ISSUE #417: migrate thumbnails

### LINKED ISSUE #428: TechTV video/collection descriptions contain HTML
About 80 videos, 13 collections in TechTV contain some HTML in their descriptions, including hyperlinks.  These will currently display as plain text.  Options for dealing with this include:

- Strip out the HTML code.  
    - How to deal with hyperlinks: remove the link text and just leave the URL?
- Display descriptions as HTML but strip out any HTML tags entered by users in the edit form.

mbertrand:
- Closes #414
- Closes #415
- Closes #417 
- Closes #428

#### What's this PR do?
- Migrates TechTV collections, videos, thumbnails, MIT users to OVS.

- Install/start MySQL and import the TechTV database dump located at https://www.dropbox.com/home/ODL-Video/TechTV/SQL
   ```
    mysql -u <username> -p techtv <dump_file.sql>
    ```

- Run the django command:
  ```
  docker-compose run web python manage.py techtvimport --user <mysql_user>  --db techtv --host 10.0.2.2 --aws --collection 3033
  ```
- This should import one protected collection and all its videos.  
- After the celery tasks to copy over S3 video files are finished, the status of the imported videos should change to 'Complete' .
- Videos should be playable.  
- There will not be any captions to import.
- Thumbnails should be visible on the collection page. 
- The collection owner should be the same as in TechTV if the email address ends with `@mit.edu`, otherwise it should be the same as `settings.LECTURE_CAPTURE_USER` 
- You can run the following command to test processing all collections & videos but skip copying over the S3 files. Videos and collections will be created in OVS, but thumbnails won't display and videos won't play:
  ```
  docker-compose run web python manage.py techtvimport --user <mysql_user>  --db techtv --host 10.0.2.2
  ```


pdpinch:
This isn't going to close #416 anymore, is it?

mbertrand:
@pdpinch the code for that is in here, it's just not going to run based on the data for protected collections in last November's MySQL database backup.  I was thinking of adding some instructions here for manually entering a caption into the database for testing purposes.

noisecapella:
In general passwords and other sensitive values shouldn't be specified via command line argument, because the entire command is visible to all processes on a system, eg through `ps auxf`. If you try this with mysql you will see that it overwrites the characters in the password to prevent this leak, though this doesn't completely close the vulnerability: https://dev.mysql.com/doc/refman/5.7/en/password-security-user.html

Can you pass the password via environment variable instead?

techtv2ovs/management/commands/techtvimport.py
```diff
@@ -0,0 +1,72 @@
...
+            '--user',
+            dest='user',
+            help='MySQL user',
+            type=str
+        )
+
+        parser.add_argument(
+            '--protected',
+            dest='protected',
+            help='filter by protected status (value 0 or 1)',
```
noisecapella: Can you add a check somewhere to verify that the value is `0` or `1`?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+from techtv2ovs.constants import TTV_THUMB_BUCKET, TTV_VIDEO_BUCKET, ImportStatus
+from techtv2ovs.models import TechTVCollection, TechTVVideo
+from techtv2ovs.tasks import process_videofiles
+
+from ui.models import Collection, MoiraList, Video, VideoThumbnail
+
+# pylint: disable=attribute-defined-outside-init,broad-except,too-many-instance-attributes
+
+
+def mysql_query(query, connection):
```
noisecapella: Can we update this to support parameterized queries instead of using string formatting?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+            self.process_files(ttvvideo, s3videos)
+
+    def process_collection(self, *collection):
+        """
+        Iterate over every TechTV collection
+
+        Args:
+            collection (tuple): Tuple (row) of collection data from MySQL
+        """
+        (cid, name, description, email) = collection
```
noisecapella: Can we move these arguments into the function arguments instead of doing tuple unpacking here?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+            ttvcollection (TechTVCollection): a TechTVCollection object
+        """
+        video_query = (
+            'SELECT v.id, v.title, v.description, v.external_id, v.private, v.private_token '
+            'from videos v inner join collection_videos cv on cv.video_id = v.id where cv.collection_id = {} '
+            'and v.status = \'approved\' AND v.file_restored = 1 AND v.locked=0;'.format(ttvcollection.id)
+        )
+        videos = mysql_query(video_query, self.connection)
+        for video in videos:
+            (v_id, v_title, v_description, external_id, private, token) = video
```
noisecapella: You can do this unpacking in the for loop directly:

    for v_id, v_title, v_description, external_id, private, token in videos:

techtv2ovs/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+        run_copy (bool): Copy S3 files between buckets
+    """
+    ttv_video = TechTVVideo.objects.get(id=ttv_id)
+    ttv_video.videofile_status = ImportStatus.CREATED
+    for file in files:
+        try:
+            # Copy S3 objects to OVS buckets (if not already there)
+            src = {'Bucket': TTV_VIDEO_BUCKET, 'Key': file['Key']}
+            if 'original' in file['Key']:
+                dst = {
```
noisecapella: It looks like we are not ever using `dst` itself, just the bucket and the key inside it. Can you replace with separate variables for the both of them?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+            ttv_video.save()
+            return
+        user_list = [thumb for thumb in s3list['Contents'] if '{}/jumbo'.format(ttv_video.ttv_id) in thumb['Key']]
+        default_list = [thumb for thumb in s3list['Contents'] if 'thumbnails/jumbo' in thumb['Key']]
+        idx = 0
+        for thumb in user_list + default_list:
+            new_key = thumb['Key'].split('/')[-1].replace('jumbo', str(idx))
+            try:
+                src = {'Bucket': TTV_THUMB_BUCKET, 'Key': thumb['Key']}
+                dst = {
```
noisecapella: It looks like we are not ever using `dst` itself, just the bucket and the key inside it. Can you replace with separate variables for the both of them?

techtv2ovs/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Args:
+        filename (str): The video file name
+
+    Returns:
+        str: The encoding for the videofile.
+
+    """
+    basename = splitext(filename)[0].split('/')[-1]
+    return basename.upper() if basename == 'hd' else basename
```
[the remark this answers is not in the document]
mbertrand: The encoded video filenames are always the same ('basic.mp4', 'medium.mp4', 'hd.mp4', etc) but I can modify this just in case there's an exception to the rule somewhere.

techtv2ovs/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@shared_task(bind=True)
```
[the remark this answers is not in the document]
mbertrand: Here's some info on the '@shared_task' decorator, it's not really necessary but is used consistently with all the other OVS celery tasks: http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html#using-the-shared-task-decorator

The task is bound (`bind=true`) which is why it has `self` as the 1st argument, again not really necessary for this particular task as currently written but makes it consistent with the other OVS tasks: http://docs.celeryproject.org/en/latest/userguide/tasks.html#bound-tasks

techtv2ovs/admin.py
```diff
@@ -0,0 +1,30 @@
...
+
+from techtv2ovs.models import TechTVCollection, TechTVVideo
+
+
+class TTVVideosInline(admin.TabularInline):
+    """Inline model for video files"""
+    model = TechTVVideo
+
+
+class TechTVCollectionAdmin(admin.ModelAdmin):
```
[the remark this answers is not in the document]
mbertrand: Most likely it won't ever be necessary to edit them, but I feel it's nice to have that option in the admin console if a use case does ever come up.  How about I make most of them non-editable except statuses plus foreign keys to OVS collections/videos.

techtv2ovs/management/commands/techtvimport.py
mbertrand: OK

techtv2ovs/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+    Update the TechTVVideo and Video statuses when done (complete, error, or missing).
+
+    Args:
+        ttv_id (int): The ID of the TechTVVideo object
+        files (list): A list of S3 videofile objects
+        run_copy (bool): Copy S3 files between buckets
+    """
+    ttv_video = TechTVVideo.objects.get(id=ttv_id)
+    ttv_video.videofile_status = ImportStatus.CREATED
+    for file in files:
```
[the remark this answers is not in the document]
mbertrand: ✅

techtv2ovs/tasks.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+    connection.execute(query)
+    return connection.fetchall()
+
+
+def moiralist_name(cid, name):
+    """
+    Return a formatted moira list name based on the name of a collection
+    Based on TechTV / Ruby 'parameterize' function
+    https://github.com/TechTV/TechTV/blob/master/app/models/collection/moira_controllable.rb#L76
```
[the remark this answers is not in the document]
mbertrand: It's a private repo, link works with access, just added you as a collaborator.

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+class TechTVImporter(object):
+    """
+    Class for handling TechTV to OVS imports
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Assign class properties based on kwargs
+        """
+        self.db_user = kwargs.get('db_user', None)
```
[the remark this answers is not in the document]
mbertrand: I did originally but thought it looked messier.  I will revert back if you prefer.

techtv2ovs/utils.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Returns:
+        str: String with HTML tags/characters removed.
+    """
+    if not txt:
+        return ''
+    return normalize("NFKD", unescape(strip_tags(txt)))
+
+
+class TechTVImporter(object):
```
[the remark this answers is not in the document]
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

app.json
```diff
@@ -42,6 +42,10 @@
       "description": "The url for the Mailgun API",
       "required": true
     },
+    "MYQSL_PASSWORD": {
```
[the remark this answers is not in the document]
mbertrand: ✅

techtv2ovs/utils.py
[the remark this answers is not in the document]
noisecapella: I see it, thanks

techtv2ovs/admin.py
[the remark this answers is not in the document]
noisecapella: Sure, sounds good to me

app.json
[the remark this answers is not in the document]
noisecapella: Instead of having it be a Django setting can you use `os.getenv` in the management command instead?

techtv2ovs/tasks.py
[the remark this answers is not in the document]
noisecapella: Sounds good

techtv2ovs/utils.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        collection_query = (
+            'SELECT DISTINCT c.id, c.name, c.description, u.email FROM collections c '
+            'INNER JOIN collection_videos cv ON c.id = cv.collection_id '
+            'INNER JOIN videos v on v.id = cv.video_id '
+            'LEFT JOIN users u ON c.user_id = u.id '
+            'WHERE v.status = \'approved\' and v.file_restored = 1 AND v.locked = 0 AND c.protected = %s'
+        )
+        if self.collection:
+            collection_query += " and c.id = {}".format(self.collection)
```
noisecapella: Can you change this to use a parameter instead of string formatting?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,314 @@
...
+    """
+    Get all the available S3 videofile objects for a TechTVVideo
+
+    Args:
+        s3_prefix (str): The TechTVVideo external id to use as the S3 bucket prefix to filter by
+
+    Returns:
+        list: A list of S3 objects
+    """
+    s3client = boto3.client('s3')
```
noisecapella: It looks like this function is uncovered, can you write a test for it?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,314 @@
...
+    Replace any HTML tags with spaces, HTML-encoded characters with unicode
+
+    Args:
+        txt (str): Text to remove HTML tags from.
+
+    Returns:
+        str: String with HTML tags/characters removed.
+    """
+    if not txt:
+        return ''
```
noisecapella: Can you write a test covering this case?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def process_files(self, ttv_video, s3videos):
+        """
+        Import videofiles, thumbnails, and captions for a TechTVVideo.
+
+        Args:
+            ttv_video (TechTVVideo): The TechTVVideo object to process
+            s3videos (list): List of video file S3 objects
+        """
+        self.process_thumbs(ttv_video)
```
noisecapella: Can you write a test covering this function?

techtv2ovs/utils.py
```diff
@@ -0,0 +1,314 @@
...
+        self.collection = collection
+        self.aws = aws
+        self.protected = protected
+        self.output = output
+
+    def run(self):
+        """
+        Run the import
+        """
+        with MySQLdb.connect(
```
noisecapella: Can you write a test for this function? I'm mostly concerned about the last line, that there's an assertion that `self.etl_graph()` is the thing which gets executed

techtv2ovs/utils.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        This function builds the Bonobo ETL graph that needs to be executed.
+
+        Returns:
+             bonobo.Graph: Graph of ETL functions
+
+        """
+        graph = bonobo.Graph()
+        graph.add_chain(self.extract_collections, self.process_collection)
+        return graph
```
noisecapella: Can you write a test for this function?

mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: ✅

odl_video/settings.py
```diff
@@ -518,3 +519,7 @@ def get_all_config_keys():
         'odl_video.middleware.QueryStringFeatureFlagMiddleware',
         'odl_video.middleware.CookieFeatureFlagMiddleware',
     )
+
+
+# MySQL password, only required for importing TechTV videos.
+MYSQL_PASSWORD = get_string("MYSQL_PASSWORD", "")
```
noisecapella: Instead of having this as a setting can you use `os.getenv` since we are only using it in the management command?

PR #435, opened 2018-02-09
Handle nested moira permissions on individual video/collection pages

### LINKED ISSUE #422: Modify collection/video permissions to account for nested moira lists
Currently, permission to view a collection/video is determined by looking for overlap between the moira lists a user belongs to and the moira lists the collection/video is associated with.  This allows for fast retrieval of all the collections accessible to a user, but fails to account for nested moira lists.

Retrieving all the collections a user has access to should probably continue to use the above permissions strategy, but when on an individual collection or video detail page, the permissions check should run `moira_client.list_members()` for each associated moira list; this will return nested users.

mbertrand:
- Closes #422 

#### What's this PR do?
Modifies permission methods to check the members of each view/admin moira list associated with a collection or video, one by one, but immediately return True if a match is found for the request user.  

- Create a collection with 1+ videos, with view permissions set to the`odl-engineering` moira list (and optionally 1 or 2 more lists).
- Create a user with username/email of someone who is in the `odl-engineering-platform` moira list, and make it a staff user.
- Log in as that user.  The above collection will still not show up in the list of available collections, but if you go directly to that collection's URL, you should be allowed to view the collection and its video.
- Repeat the above but remove `odl-engineering` from view permissions and add it to admin permissions instead.  You should be able to view and edit the collection and video.
- Repeat the above but remove `odl-engineering` from the collection admin permissions and add it to the video-specific permissions instead (you will need to set `ENABLE_VIDEO_PERMISSIONS=True`).

### DIFF
--- a/ui/views_test.py
+++ b/ui/views_test.py
@@ -24,6 +24,7 @@ from ui.models import VideoSubtitle
 from ui.serializers import (
     DropboxUploadSerializer,
     VideoSerializer)
+from ui.utils import get_moira_user
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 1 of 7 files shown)


ui/views_test.py
```diff
@@ -561,8 +565,9 @@ def test_page_not_found(url, logged_in_apiclient, settings):
     }
 
 
-def test_terms_page(mock_moira_client, logged_in_client):
+def test_terms_page(mocker, logged_in_client):
     """Test terms page"""
+    mocker.patch('ui.utils.get_moira_client')
```
noisecapella: It seems a little odd that we need the moira client for the terms of service page, but I'm fine leaving this since it's out of scope of the PR

Issue #138, opened 2017-08-14
Epic: Production Deployment


pdpinch:
## "Production readiness" checklist (based off of [MIT ODL handbook](https://github.com/mitodl/handbook/blob/master/production-ready.md))

- [x] finalized DNS names
- [x] Google analytics
- [ ] document decision to use Heroku, ec2 or Heroku/ec2 hybrid
- [x] deploy current code to production system
- [ ] document production deployment process (if not automated)
- [ ] Performance Monitoring?
- [x] Production error monitoring (Sentry)
- [ ] Mail Monitoring
- [ ] ~~Load Testing (?)~~

### Required features:
- [x] Does it support configuration via environment variables?
- [x] Does it support configuration via flatfile (e.g. YAML)?
- [ ] Does it have a sane default configuration (e.g. as close to a realistic production configuration as possible)?
- [x] Does it run on parallel HTTP workers? (e.g. uWSGI, Gunicorn)?
- [ ] Does it log to standard out/error?
- [ ] Does it log to syslog?
- [ ] Can it send stack trace emails to a configurable email list?
- [x] Does it have configurable log levels (e.g. DEBUG, ERROR, WARNING, INFO)?
- [x] Does it have a test suite which measures code coverage?
- [ ] Does it have a status route (which requires a secret token) that can be used for health checking by a monitoring service?
- [ ] Does the health check validate that each service the app depends on is reachable (e.g. databases, caching, external APIs when necessary)?
- [ ] Does the health check return a response body with the availability information above in a machine-readable format in all cases?
- [ ] Does the health check return a 500 status code when any component in the health check is detected to be failing?

### Required documentation/service infrastructure:
- [x] Does it have continuous integration builds which run the full test suite?
- [x] Does it have an automated continuous delivery to a development environment?
- [ ] Does it have a documented release process?
- [ ] Does it's release process include staging and production builds, where staging builds are promoted to production?
- [ ] Does it's release process require that code changes be "signed off" as working on staging before being promoted to production?
- [ ] Does it have application/resource monitoring in production (e.g. New Relic agent, Zenoss SNMP)?
- [ ] Does it have service availability monitoring in production (e.g. periodic health checks with Zenoss or New Relic)?
- [ ] Does it have automated, periodically generated, fully restorable backups for each non-ephemeral data store in production?

pdpinch:
@shaidar @blarghmatey can you guys take a look at this and add/remove some details

[...]

PR #447, opened 2018-02-23
Updated settings and requirements to fix deployment issues

blarghmatey:
#138 

#### What's this PR do?
The file parsing available in uWSGI has a hardcoded line limit which was causing problems with managing the certificate strings via the env file. Added a block to read in the env file and set environment variables automatically as part of settings.py to work around that issue.

Removed the `-e` flags from the git dependencies so that they get installed to the `dist-packages` directory instead of the home directory of the user that runs the pip command. This was causing issues with being able to import those packages at runtime.

Added the option to set the celery broker separately from the results backend in order to allow for using RabbitMQ in place of Redis.

Ensure that the app will start properly with an env file formatted as `ENV_VAR=value` with no `export` prefix.

odl_video/settings.py and requirements.txt

### DIFF
--- a/odl_video/envs.py
+++ b/odl_video/envs.py
@@ -153,3 +153,26 @@ def get_key(name, default):
     if not isinstance(value, str):
         return value
     return value.encode().decode('unicode_escape').encode()
+
+
+def parse_env(env_file):
+    """
+    Parse the env file and set the values in the runtime environment.
+    Simplifies deployment by ensuring env values are present without
+    duplicating across multiple locations and not requiring explicit
+    sourcing.
+
+    Args:
+      env_file (str): path to a file consisting of key=value pairs
+
+    Returns:
+      None: No return value
+
+    """
+    try:
+        with open(env_file) as envsettings:
+            for line in envsettings:
+                k, v = line.rstrip('\n').lstrip('export ').split('=', maxsplit=1)
+                os.environ.setdefault(k, v)
+    except FileNotFoundError:
+        pass
--- a/odl_video/envs_test.py
+++ b/odl_video/envs_test.py
@@ -1,6 +1,7 @@
 """Tests for environment variable parsing functions"""
 from unittest.mock import patch
 
+import os
 import pytest
 
 from odl_video.envs import (
@@ -11,6 +12,7 @@ from odl_video.envs import (
     get_key,
     get_list_of_str,
     get_string,
+    parse_env
 )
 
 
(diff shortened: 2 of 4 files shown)


odl_video/settings.py
```diff
@@ -21,6 +21,14 @@
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+try:
+    with open(f'{BASE_DIR}/.env') as envsettings:
+        for line in envsettings:
+            k, v = line.rstrip('\n').lstrip('export ').split('=', maxsplit=1)
+            os.environ.setdefault(k=v)
```
mbertrand: This throws an error: `TypeError: setdefault() got an unexpected keyword argument 'k'`

This works: `os.environ.setdefault(k, v)`

odl_video/settings.py
```diff
@@ -488,7 +497,7 @@
     },
     'redis': {
         "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": CELERY_BROKER_URL,
+        "LOCATION": REDIS_URL,
```
mbertrand: `CELERY_BROKER_URL` will default to `REDIS_URL` if the env. variable isn't set, so maybe it might be best to not change this line, @noisecapella do you want to weigh in?  That seems to be the pattern used in open-discussions & micromasters as well:
```
CELERY_BROKER_URL = get_string("CELERY_BROKER_URL", REDIS_URL)
```

noisecapella: It looks like Giovanni added that line in a commit to add mandatory settings, I don't think `REDIS_URL` has any special meaning to Celery or django-redis. I think the change here makes sense to keep the celery redis broker separate from the redis cache Django

odl_video/envs_test.py
```diff
@@ -148,3 +150,13 @@ def test_get_key():
             b'/q0txR0v1rqmowS1mQIDAQAB\n'
             b'-----END PUBLIC KEY-----\n'
         )
+
+
+def test_parse_env():
+    testpath = 'testenv.txt'
```
noisecapella: Can you use a `TemporaryFile` here or put it into a temporary directory? If it's not done with a context manager or something we should also put the `os.remove` in a `finally` block

odl_video/envs_test.py
```diff
@@ -148,3 +150,13 @@ def test_get_key():
...
+    with open(testpath, 'w') as testfile:
+        testfile.write('FOO_VAR=bar_var\nFOO_NUM=42')
```
noisecapella: Can you change one of the lines to include an `"export "`, and have one of the values include its own `=` sign?

odl_video/envs.py
```diff
@@ -153,3 +153,24 @@ def get_key(name, default):
...
+
+
+def parse_env(env_file):
+    """
+    Parse the env file and set the values in the runtime environment.
+    Simplifies deployment by ensuring env values are present without
+    duplicating across multiple locations and not requiring explicit
+    sourcing.
+
+    :param env_file: path to a file consisting of key=value pairs
```
mbertrand: @blarghmatey The typical format looks like [this](https://github.com/mitodl/odl-video-service/blob/master/odl_video/envs.py#L16-L22), can you change it? Otherwise looks good!

Issue #452, opened 2018-02-26
Update Player to handle YouTube & Cloudfront streams


pdpinch:
This issue builds on #333 and replaces #348 

For public videos, the video stream may be available from YouTube, Cloudfront or both. The player (either backend, front-end or combined) should support the following business logic:

1. If there is no YouTube video, stream from Cloudfront (existing behavior for private videos)
2. If there is a YouTube video and a Cloudfront video, test whether YouTube will work:
   1. If the YouTube API is responding, stream from YouTube. 
   2. if the YouTube API is not responding, stream from Cloudfront
3. If there is a YouTube video and no Cloudfront, stream from YouTube

We can modify the implementation of this logic, as long as the 3 cases mentioned on #451 are covered.

PR #446, opened 2018-02-22
TechTV caption import

mbertrand:
#### What are the relevant tickets?
Closes #416

#### What's this PR do?
Imports captions from a TechTV video into OVS

#### How should this be manually tested?
- Get a token for your dropbox app at https://www.dropbox.com/developers/apps
    - Click the 'Generate token' button
    - The dropbox account must have access to the folder `ODL-Video/TechTV/Captions` for this to work properly: the account owner should be on the 'ODL-Video' dropbox team, and the app should have 'Full Dropbox' access.
    - In your `.env` file, set `DROPBOX_TOKEN` to your token value.

- Install/start MySQL and import the TechTV database dump located at https://www.dropbox.com/home/ODL-Video/TechTV/SQL
   ```
    mysql -u <username> -p techtv <dump_file.sql>
    ```

- Run the django command:
    ```
     docker-compose run web python manage.py techtvimport --aws --user <mysql_user> --db techtv --host 10.0.2.2 --collection 3549 --protected 0
    ```

Three videos should be imported with working caption files in the collection "MIT Diversity Summits"

PR #454, opened 2018-03-01
Import public TechTV collections and set video stream source

### LINKED ISSUE #451: Migrate TechTV Public Collections to OVS
This epic is part of a larger effort to shut down the TechTV web site.

There are three types of public collections on TechTV. Since they are similar, we can migrate them at the same time, applying slightly different business logic:

### acceptance criteria

- [ ] 1. Public videos that cannot be uploaded to YouTube (e.g. due to lack of permission releases). These will only be available through Cloudfront. Example: MIT World
- [ ] 2. Public videos that are sensitive to region-blocking (e.g. must be viewable in China). These will be streamed through YouTube, with a failover to Cloudfront. Example: Blossoms, OCW
- [ ] 3. All public videos that don't fall into case 1 or 2 will be uploaded to YouTube. We won't stream these from CloudFront. Example: PTSG
- [ ] If captions exist in TechTV, they should be migrated to OVS

@emmello will provide a list of collections that fall into 1 and 2 above. All other collections will be handled as 3 above. 

### related issues 
- Update Player to handle YouTube & Cloudfront streams

mbertrand:
~~Review/merge PR #446 first~~

Fixes #451

#### What's this PR do?
Adds a `stream_source` attribute to the `Collection` model and adds additional parameters to the TechTV import script to handle public collections:
  - `noyoutube` parameter:  videos for public collection ids following this parameter will not be uploaded to youtube (stream_source set to `Cloudfront`)
  - `cloudfront` parameter: videos for public collection ids following this parameter will be streamable from cloudfront if youtube is not available (stream_source set to `None`)
  - By default, any public collection id's not included in the above parameters will be streamable from Youtube only (stream_source set to `Youtube`)

Adds/modifies post-save signals to `Collection` and `Video` so that public videos will either be uploaded to or deleted from Youtube depending on the value of the collection's `stream_source`.

- Set `ENABLE_VIDEO_PERMISSIONS=True` in your `.env` file
- Set `DROPBOX_TOKEN` (you'll need a dropbox app) and `DROPBOX_FOLDER` (where the `ODL-Video/TechTV/Captions` dropbox folder is shared, your app will need access to it).
- Install/start MySQL and import the TechTV database dump located at https://www.dropbox.com/home/ODL-Video/TechTV/SQL
   ```
    mysql -u <username> -p techtv <dump_file.sql>
    ```
- Run the command: 
  ```docker-compose run web python manage.py techtvimport --user <mysql_user> --db <mysql_db> --host <mysql_host> --protected 0 --aws --collections 3255```
  - The new collection should have a `stream_source` of `Youtube`, and after videofiles are done being copied from S3 there should be a `YoutubeVideo` object for each `Video` in the collection (this may take a few minutes).

- Run the command: 
  ```docker-compose run web python manage.py techtvimport --user <mysql_user> --db <mysql_db> --host <mysql_host> --protected 0 --aws --cloudfront 3255 --collections 3255```
  - The collection should have a `stream_source` of `None`, and there should still be a `YoutubeVideo` object for each `Video` in the collection

- Run the command: 
  ```docker-compose run web python manage.py techtvimport --user <mysql_user> --db <mysql_db> --host <mysql_host> --protected 0 --aws --noyoutube 3255 --collections 3255```
  - The collection should have a `stream_source` of `Cloudfront`, and there should no longer be any `YoutubeVideo` objects for the collection's videos.

- In the admin console, change the collection's `stream_source` to `None` or `Youtube`.  After a minute or two there should be a `YoutubeVideo` object for each `Video` in the collection

- Every video in the collection should have its permission set to `Public`, regardless of which command above is run.

For #452, the following logic will be used to play public videos:
  - video collection `stream_source` = `None`: play from Youtube if possible, otherwise play from Cloudfront
  - video collection `stream_source` = `Cloudfront`: play from Cloudfront, never Youtube
  - video collection `stream_source` = `Youtube`: play from Youtube, never Cloudfront


noisecapella:
When I do the `--noyoutube` step it doesn't appear to delete any `YouTubeVideo` objects

noisecapella:
For the first step in the manual testing it plays from cloudfront even though the collection `stream_source` says `Youtube`

mbertrand:
@noisecapella the code to conditionally play from youtube if available is in another PR, #460.  This PR just sets up most of the backend plumbing for that.

techtv2ovs/utils.py
```diff
@@ -155,6 +159,23 @@ def run(self):
...
             )
 
+    def get_stream_source(self, collection_id):
+        """
+        Return the appropriate stream_source for transcoded videos
+        Args:
+            collection_id (int): The collection id
+
+        Returns:
+            StreamSource: where the video should be streamed from - Youtube, Cloudfront, or either one (None)
```
noisecapella: The return type should be `str`

techtv2ovs/utils.py
```diff
@@ -155,6 +159,23 @@ def run(self):
...
+        """
+        if self.protected == 1:
```
noisecapella: Is this because youtube videos can't be protected? If so why wouldn't this return `CLOUDFRONT`?

mbertrand:
For the --noyoutube option, looks like that is a consequence of rerunning the command again on the same collection, the delete signal isn't triggered.  The stream_source parameter is set correctly and if I manually save the collection then the signal is triggered and the youtube videos are all deleted.

ui/signals.py
```diff
@@ -46,18 +48,37 @@ def delete_youtube_video(sender, **kwargs):
...
 
-@receiver(post_save, sender=Video)
-def sync_youtube(sender, **kwargs):
+@receiver(post_save, sender=Collection)
+def update_collection_youtube(sender, **kwargs):
     """
+    If a collection's stream source is changed, sync YoutubeVideo objects for public Videos
+    """
+    if settings.ENABLE_VIDEO_PERMISSIONS:
+        for video in kwargs['instance'].videos.filter(is_public=True):
```
noisecapella: Wouldn't we want to do all videos and not just the public ones? From the docstring it looks like `sync_youtube` should delete videos which aren't public

ui/signals.py
```diff
@@ -46,18 +48,37 @@ def delete_youtube_video(sender, **kwargs):
...
-        video = kwargs['instance']
-        yt_video_id = video.youtube_id
-        if video.is_public:
-            if not yt_video_id:
+        sync_youtube(kwargs['instance'])
+
+
+def sync_youtube(video):
+    """
+    Upload a video to youtube if it is public, not already on YouTube, and collection stream_source != cloudfront.
```
noisecapella: Can you add the argument to the docstring?

ui/signals.py
```diff
@@ -46,18 +48,37 @@ def delete_youtube_video(sender, **kwargs):
...
+        if video.original_video is not None:
+            if yt_video is None:
+                upload_youtube_video.delay(video.id)
+            elif yt_video.status in (YouTubeStatus.FAILED, YouTubeStatus.REJECTED):
+                YouTubeVideo.objects.get(id=yt_video.id).delete()
                 upload_youtube_video.delay(video.id)
-        elif yt_video_id:
-            YouTubeVideo.objects.get(id=yt_video_id).delete()
+    elif yt_video is not None:
+        YouTubeVideo.objects.get(id=yt_video.id).delete()
```
noisecapella: After the `delete()` above it calls `upload_youtube_video`, should we do the same here?

ui/views_test.py
```diff
@@ -96,6 +97,7 @@ def test_index(client):
     assert response.url == reverse('collection-react-view')
 
 
+@override_settings(ENABLE_VIDEO_PERMISSIONS=False)
```
noisecapella: Using `override_settings` and the `settings` fixture at the same time might be confusing. Can we set the setting in the test itself instead?

ui/signals.py
mbertrand: It should only be deleted here, because if it gets to this line it means either the video is no longer public or the collection `stream_source` is set to `Cloudfront`, so it should be removed from YouTube.

ui/signals.py
mbertrand: Only public videos should potentially be on Youtube, and any videos that are changed from public to private should be deleted from Youtube via the `update_video_youtube` signal.  I was filtering here to avoid a little overhead in iterating through all the private videos, but I can remove this just in case there's any private videos that somehow get a Youtube video associated with them, maybe if the signal fails.

techtv2ovs/utils.py
mbertrand: ✅

techtv2ovs/utils.py
mbertrand: Correct.  Once `ENABLE_VIDEO_PERMISSIONS` is set to `True`, people will be able to manually  override collection-level permissions for individual videos and make them public, and I think in that case we want it to stream from Youtube if available, Cloudfront otherwise (`stream_source` = `None`). @pdpinch let me know if I've got that wrong.

ui/signals.py
mbertrand: ✅

ui/views_test.py
mbertrand: ✅

PR #459, opened 2018-03-08
Join BASE_DIR for STATIC_ROOT

### LINKED ISSUE #467: `collectstatic` puts files in wrong directory when OVS is deployed via Salt
The Salt deployment of OVS to production causes the static files to be copied to `/home/deploy` which is not the correct folder for the app, because the `STATIC_ROOT` setting is relative.

mbertrand:
Fixes #467

#### What's this PR do?
Fixes a deployment issue where `collectstatic` was copying files to the wrong folder

Nothing should break

### DIFF
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -182,7 +182,7 @@ USE_TZ = True
 # https://docs.djangoproject.com/en/1.11/howto/static-files/
 
 STATIC_URL = '/static/'
-STATIC_ROOT = 'staticfiles'
+STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')
 STATICFILES_DIRS = (
     os.path.join(BASE_DIR, 'static'),
 )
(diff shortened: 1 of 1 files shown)


noisecapella:
What was the deployment issue? We have the same setup with our other apps so we should probably make this fix elsewhere if necessary

mbertrand:
I think the issue was that static files were being copied over to the wrong folder, @blarghmatey do you think there's some other way to prevent that besides altering `STATIC_ROOT` here?

mbertrand:
@noisecapella do you still prefer another solution on the devops side for the staticfiles problem?

noisecapella:
I'm fine with this solution, I'm just curious what this fixes because we might want to fix this in micromasters and open-discussions too

mbertrand:
@noisecapella I created an issue for this, @blarghmatey feel free to expand on the description if you'd like: #467

PR #460, opened 2018-03-08
VideoJS YouTube Player

### LINKED ISSUE #452: Update Player to handle YouTube & Cloudfront streams

mbertrand:
- Closes #452 

#### What's this PR do?
- Videos with a collection `stream_source` of `None` will play from YouTube if available, Cloudfront otherwise.
- Videos with a collection `stream_source` of `Youtube` will only play from YouTube, available or not.
- Videos with a collection `stream_source` of `Cloudfront` will never play from YouTube, only Cloudfront.

- Assign working values for `.env` variables `YT_CLIENT_ID`, `YT_PROJECT_ID`, `YT_CLIENT_SECRET`, `YT_ACCESS_TOKEN`, `YT_REFRESH_TOKEN` (I can send them to you)
- You can run the import command 3x as explained in PR #454.  Or just change the `stream_source` value of an existing collection.
- After each run/stream_source change, wait a few minutes if necessary for videos to upload to Youtube, then try to play a video.  The video should play from the expected source based on the rules above.
- Block off youtube by modifying your `/etc/hosts` file (or ad blocker list).  Then try playing a video.  The video should still play from Cloudfront if its collection `stream_source` is `None` or `Cloudfront`.  It should not be playable if the `stream_source` is `Youtube`.
    - This is done in the front-end on page load by trying to create an `Image` with a src url equal to the first thumbnail for a Youtube video, the assumption being that if that cannot be loaded then the video can't be either.  

Note: I had to manually edit yarn.lock because it was creating duplicate dependencies for video.js that broke the youtube plugin, I think it may be related [to this bug](https://github.com/yarnpkg/yarn/issues/3967).

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -21,17 +21,25 @@ const gaEvents = [
   "ended"
 ]
 
-const makeConfigForVideo = (video: Video): Object => ({
+const makeConfigForVideo = (video: Video, useYouTube: boolean): Object => ({
   autoplay:    false,
   controls:    true,
-  fluid:       false,
+  fluid:       useYouTube,
   playsinline: true,
+  techOrder:   useYouTube ? ["youtube", "html5"] : ["html5"],
   html5:       {
     nativeTextTracks: false
   },
   playbackRates: [0.5, 0.75, 1.0, 1.25, 1.5, 2.0, 4.0],
-  sources:       video.sources,
-  plugins:       {
+  sources:       useYouTube
+    ? [
+      {
+        type: "video/youtube",
+        src:  `https://www.youtube.com/watch?v=${video.youtube_id || ""}`
+      }
+    ]
+    : video.sources,
+  plugins: {
     videoJsResolutionSwitcher: {
       default:      "high",
       dynamicLabel: true
@@ -217,6 +225,29 @@ export default class VideoPlayer extends React.Component<*, void> {
     )
   }
 
+  switchVideoSource = () => {
+    const { video } = this.props
+    if (video.sources.length > 0) {
+      this.player
+        .reset()
+        .src(video.sources)
+        .fluid(false)
+    }
+  }
+
+  imageExists(url: string) {
+    const img = new Image()
+    img.onerror = this.switchVideoSource
+    img.src = url
+  }
+
+  checkYouTube = async () => {
+    const { video } = this.props
+    // Try to load the YouTube video thumbnail image.  Assumes video availability == thumbnail availability
+    const imgUrl = `https://img.youtube.com/vi/${video.youtube_id || ""}/0.jpg`
+    this.imageExists(imgUrl)
+  }
+
   sendEvent = (action: string, label: string) => {
     if (action === "timeupdate") {
       // Track amount played in increments of 60 seconds
@@ -268,6 +299,7 @@ export default class VideoPlayer extends React.Component<*, void> {
 
   componentDidMount() {
     const { video, selectedCorner } = this.props
+
     const cropVideo = this.cropVideo
     const createEventHandler = this.createEventHandler
     const toggleFullscreen = this.toggleFullscreen
@@ -277,10 +309,11 @@ export default class VideoPlayer extends React.Component<*, void> {
         "FullscreenToggle"
       ).prototype.handleClick = toggleFullscreen
     }
+    const useYouTube = video.is_public && video.youtube_id !== null
     this.lastMinuteTracked = null
     this.player = videojs(
       this.videoNode,
-      makeConfigForVideo(video),
+      makeConfigForVideo(video, useYouTube),
       function onPlayerReady() {
         this.enableTouchActivity()
         if (video.multiangle) {
@@ -299,6 +332,9 @@ export default class VideoPlayer extends React.Component<*, void> {
         }
       }
     )
+    if (useYouTube) {
+      this.checkYouTube()
+    }
     this.updateSubtitles()
   }
 
--- a/static/js/components/VideoPlayer_test.js
+++ b/static/js/components/VideoPlayer_test.js
@@ -6,12 +6,17 @@ import sinon from "sinon"
 import { mount } from "enzyme"
 
 import VideoPlayer from "./VideoPlayer"
-import { makeVideo, makeVideoSubtitle } from "../factories/video"
+import {
+  makeVideo,
+  makeVideoSource,
+  makeVideoSubtitle
+} from "../factories/video"
 import * as libVideo from "../lib/video"
 import ga from "react-ga"
 import { FULLSCREEN_API } from "../util/fullscreen_api"
 import { CANVASES } from "../constants"
 import { makeVideoSubtitleUrl } from "../lib/urls"
+import { expect } from "../util/test_utils"
 
 describe("VideoPlayer", () => {
   let video,
@@ -48,6 +53,9 @@ describe("VideoPlayer", () => {
       tracks:        [],
       on:            sandbox.stub(),
       tech_:         {},
+      reset:         sandbox.stub().returns(playerStub),
+      src:           sandbox.stub().returns(playerStub),
+      fluid:         sandbox.stub().returns(playerStub),
       currentTime:   () => 630.5,
       duration:      () => 2400.0,
       videoWidth:    () => 640,
@@ -85,6 +93,7 @@ describe("VideoPlayer", () => {
         controls:    true,
         fluid:       false,
         playsinline: true,
+        techOrder:   ["html5"],
         html5:       {
           nativeTextTracks: false
         },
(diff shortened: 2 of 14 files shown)


pdpinch:
Do the changes to requirements.txt need to be in this PR?

pdpinch:
Do the changes to requirements.txt need to be in this PR?

mbertrand:
@pdpinch no but this PR is still in progress and includes code from PR #451; after that one is approved and merged, this PR will be rebased and cleaned up.

static/js/components/VideoPlayer.js
```diff
@@ -216,6 +228,30 @@ export default class VideoPlayer extends React.Component<*, void> {
     )
   }
 
+  checkYouTube = () => {
```
mbertrand: @aliceriot if you have a moment, could you quickly look this function over and tell me if you think there's a better approach?  The purpose is to make sure the video's youtube URL is reachable, and if not modify the VideoJS configuration to play from Cloudfront instead.

aliceriot: I think this looks relatively straightforward. The only changes I would suggest are rewriting it using the `async / await` syntax, which I think is generally a little bit more readable. If you did that the function would look something like:

```js
checkYouTube = async () => {
  const { video } = this.props
  if (video.youtube_url) {
    let response
    try {
      response = await fetch(video.youtube_url, {
        mode: "no-cors"
      })
    } finally {
      if (!response || !response.ok) {
        player.reset().src(video.sources).fluid(false)
      }
    }
  }
}
```

or something like that, haha

aliceriot:
ok! after much finagling I can confirm that it works. reading code now

static/js/components/VideoPlayer.js
```diff
@@ -216,6 +225,34 @@ export default class VideoPlayer extends React.Component<*, void> {
     )
   }
 
+  imageLoaded = (loaded: boolean) => {
```
aliceriot: I might suggest naming this function more to reflect what it does - maybe `setVideoSource`  or something like that?

static/js/components/VideoPlayer.js
```diff
@@ -216,6 +225,34 @@ export default class VideoPlayer extends React.Component<*, void> {
...
+  imageExists(url: string, callback: Function) {
+    const img = new Image()
+    img.onload = function() {
+      callback(true)
+    }
+    img.onerror = function() {
+      callback(false)
+    }
+    img.src = url
+  }
```
aliceriot: seems to me like there's no reason to call `imageLoaded` with `true` as an argument - you could refactor the `imageLoaded` function to just do the `this.player.reset()...` business - then (I think) you might be able to get away with just doing:

```js
imageExists = (url: string) => {
  const img = new Image()
  img.onerror = this.imageLoaded
  img.src = url
}
```

static/js/components/VideoPlayer_test.js
```diff
@@ -225,17 +235,6 @@ describe("VideoPlayer", () => {
...
-    wrapper.instance().player = playerStub
-    // $FlowFixMe
-    containerStub.parentElement[FULLSCREEN_API.requestFullscreen] = () => {}
-    wrapper.instance().videoContainer = containerStub
-
-    wrapper.instance().toggleFullscreen()
-    sinon.assert.calledWith(wrapper.instance().player.el_.dispatchEvent)
-  })
-
   it("toggleFullScreen on causes player to dispatchEvent", () => {
```
aliceriot: was this just accidentally duplicated?

static/js/components/VideoPlayer_test.js
```diff
@@ -304,4 +303,30 @@ describe("VideoPlayer", () => {
       }
     })
   })
+  ;[false, true].forEach(function(useYoutube) {
+    it(`checkYouTube ${expect(
+      useYoutube
+    )} be called if useYoutube is ${String(useYoutube)}`, async () => {
+      const wrapper = renderPlayer({ useYouTube: useYoutube })
```
aliceriot: small nit: when you're sticking a variable into an object and you want it in the object under the same key as the variable's name there is a little syntax sugar:

```js
const foo = "hey!"
const fooInObj = { foo }
```

then `fooInObj.foo` will have the same value as `foo`.

static/js/components/VideoPlayer_test.js
```diff
@@ -304,4 +303,30 @@ describe("VideoPlayer", () => {
...
+      const checkStub = sandbox.stub(wrapper.instance(), "checkYouTube")
+      wrapper.instance().componentDidMount()
+      sinon.assert.callCount(checkStub, useYoutube ? 1 : 0)
```
aliceriot: this is probably stupid but you could use `Number(useYoutube)` here instead of the ternary :laughing:

static/js/containers/VideoDetailPage.js
```diff
@@ -125,6 +125,7 @@ class VideoDetailPage extends React.Component<*, void> {
         video={video}
         cornerFunc={this.updateCorner}
         selectedCorner={videoUi.corner}
+        useYouTube={video.is_public && video.youtube_id !== null}
```
aliceriot: do we think this is every likely to be set in the UI or something similar? if not, I'd suggest we just compute this value from `video` inside of the component, instead of passing it in as a prop (it's a little more indirect this way IMO)

static/js/components/VideoPlayer.js
mbertrand: sure, I think I'll rename it to `switchVideoSource`

static/js/components/VideoPlayer.js

static/js/components/VideoPlayer_test.js
mbertrand: Yes, it was a dupe, not testing anything not already covered.

static/js/components/VideoPlayer_test.js
mbertrand: Do you mean instead of
```javascript
const wrapper = renderPlayer({ useYouTube: booleanTube })
```
do:
```javascript
const useYouTube = booleanTube
const wrapper = renderPlayer({ useYouTube})
```

aliceriot: right now you have a variable `useYoutube`, if you just rename that variable to `useYouTube` you can use that syntax sugar. It's not a major thing, just wanted to point it out in case you weren't aware :)

mbertrand: Got it, thanks

static/js/components/VideoPlayer_test.js
mbertrand: ✅

static/js/components/VideoPlayer.js
```diff
@@ -237,9 +236,6 @@ export default class VideoPlayer extends React.Component<*, void> {
 
   imageExists(url: string, callback: Function) {
     const img = new Image()
-    img.onload = function() {
-      callback(true)
-    }
     img.onerror = function() {
       callback(false)
```
aliceriot: the callback we're passing in (we're only using this to call `switchVideoSource`) doesn't take a boolean as an argument anymore, so I think you could do either

```js
img.onerror = () => callback()
```

or maybe just directly call

```js
img.onerror = this.switchVideoSource
```

and not bother having this function (`imageExists`) take the callback argument

static/js/components/VideoPlayer_test.js
```diff
@@ -303,30 +303,30 @@ describe("VideoPlayer", () => {
...
+    )} be called if useYoutube is ${String(useYouTube)}`, async () => {
+      video.is_public = useYouTube
+      video.youtube_id = "fakeId"
+      const wrapper = renderPlayer()
       const checkStub = sandbox.stub(wrapper.instance(), "checkYouTube")
       wrapper.instance().componentDidMount()
-      sinon.assert.callCount(checkStub, useYoutube ? 1 : 0)
+      sinon.assert.callCount(checkStub, Number(useYouTube))
     })
   })
```
aliceriot: on this test, I might suggest that we make it fully parameterized, to test that things get called as we expect with all possible combinations of `is_public === true` and `video.youtube_id !== null`

static/js/components/VideoPlayer.js

static/js/components/VideoPlayer_test.js
mbertrand: ✅

mbertrand:
@aliceriot there was a recent conflict with master, just rebased it.

PR #471, opened 2018-03-20
Custom quality selector function for videojs

### LINKED ISSUE #458: Change default bitrate to be highest instead of lowest
As a lecture video viewer, I want to see a high quality video where the text is legible. 

Most of our users are on campus, with access to MIT's high speed network. To avoid complaints about low-quality "blurry" video, we should default to the highest bit rate and only decrease the bitrate if the network is too slow. Especially important for multiangle videos where we are pixel-doubling a quarter-resolution video.

mbertrand:
Closes #458

#### What's this PR do?
Replaces the default VideoJS player `selectQuality` function with a simplified function that only takes available bandwidth into account, not player dimensions, and forces the first 10 seconds of the video to play at the highest available resolution.

- Your `.env` file should have `ET_PRESET_IDS=1504127981769-6cnqhq,1504127981819-v44xlx,1504127981867-06dkm6,1506369700634-2gfhf9`
- Upload a video of moderate length (~10 minutes or more)
- Open the network tab and start playing the video.  You should see `*.ts` video files being downloaded, and `2gfhf9` should be in the filename (the highest bandwidth).
- Try to simulate a lower connection speed
    - in Chrome, use the 'toggle device toolbar' and choose iphone 6/7/8 with mid-tier mobile; or on OSX, you can also use the 'Network Link Conditioner' under 'System Preferences'.  
    - The highest bandwidth should still play (or attempt to) initially, then lower bandwidth video files (`6cnqhq`, `v44xlx`, or `06dkm6`) should start to load for slower speeds. 
    - There will be a delay before switching video quality because VideoJS will continue playing whatever segment its already downloaded, you can FF the video to avoid this.

https://github.com/videojs/videojs-contrib-hls/blob/master/docs/bitrate-switching.md

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -2,7 +2,8 @@
 /* global videojs: true */
 /* global SETTINGS: false */
 import React from "react"
-
+import R from "ramda"
+import _ from "lodash"
 import { makeVideoSubtitleUrl } from "../lib/urls"
 import { videojs } from "../lib/video"
 import type { Video, VideoSubtitle } from "../flow/videoTypes"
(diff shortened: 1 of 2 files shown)


static/js/components/VideoPlayer.js
```diff
@@ -242,11 +243,35 @@ export default class VideoPlayer extends React.Component<*, void> {
...
+    // Always start with highest bandwidth for first 10 seconds
+    if (this.player.tech_.currentTime() < 10) {
+      return _.last(playlists)
+    }
+    // Return playlist with highest bandwidth <= system bandwidth
+    return _.last(
+      R.filter(
+        rep =>
+          rep.attributes.BANDWIDTH <=
+          _.max([
```
noisecapella: This is to make sure at least one of the bandwidth options will always be in this list?

static/js/components/VideoPlayer.js
```diff
@@ -269,6 +294,9 @@ export default class VideoPlayer extends React.Component<*, void> {
             createEventHandler(event, video.key)
           })
         })
+        if (typeof this.tech_.hls !== "undefined") {
```
noisecapella: It is valid to just say `this.tech_.hls !== undefined`. Since ES5 you cannot redefine `undefined`

noisecapella:
I'm having trouble reproducing the 10 seconds high bandwidth at the beginning but I am seeing the bandwidth change when I slow or speed up the bandwidth in my browser FWIW. Maybe I just can't tell with the lecture I'm using

static/js/components/VideoPlayer.js
mbertrand: Yes

static/js/components/VideoPlayer.js

mbertrand:
@noisecapella to check on the first 10 seconds, could try this:
- artificially lower bandwidth before playback
- open up network tab
- Start playing the video
- In the network tab, you should see `*-2gfhf9.ts` files being downloaded.  
- Sometime after the video's first 10 seconds (depends on when the initial video segment finishes), it should switch to a different resolution: at that point the `*.ts` files in the network tab will not have `2gfhf9` in their names.

noisecapella:
It's subtle but it does look like the first 10 seconds are at a higher quality level, according to the file extension

Issue #402, opened 2018-01-09
Present video analytics visualization & data for collection admins


pdpinch:

adorsk:
## Querying Google Analytics: On-demand vs. Pre-queried

When should we pull data from Google Analytics (GA)?

A. At the time a user goes to the 'analytics' detail view in OVS. ("on-demand")
B. Some time before a user goes to the 'analytics' detail. ("pre-queried")

### Pros/Cons for On-Demand

#### Pros
 - no need to store GA data in OVS db. 
 - no need to write/maintain cron job to query from GA.

#### Cons
 - User has to wait for query to GA to resolve. (this is not really an issue. GA is fairly zippy, on the order of a few seconds at most).

### Pros/Cons for Pre-Queried

#### Pros
 - No need to wait for GA. (again, this is not a long wait)

#### Cons
 - need to store GA data in OVS db. 
 - need to write/maintain cron job to query from GA.

With the above pros/cons in mind, on-demand is the simpler way to go.

adorsk:
A possible Implementation plan:

1. Write a utility function in the ovs django app to query GA. Perhaps Something like this:

  ```
  def get_video_analytics_data_from_ga(video_id, timespan=None):
    # oauth dance...
    # execute analytics query...
    # format data...
    return video_analytics_data
  ```

2. Write an endpoint that proxies to the utility function.
3. Write front-end logic that renders the analytics data the endpoint returns.

Other considerations:

1. We could render charts/tables (or whatever other vizzes we want) in the backend. But imho it is better to do this on the front-end, for:
   1. better dx.
   2. flexibility to add other interactions (selecting date range, filtering tables, etc.)

mbertrand:
@adorsk all the above sounds good!  If it turns out that on-demand queries are too slow then we can switch to another approach.

roberthouse54:
There are tabs to view either a graph or table. We do not need pagination on the table, as most should have <100 rows. We will add a link to the analytics page next to the other links on the Video Details page. The overflow menu on the collection page will also have a link to the analytics page.

![image](https://user-images.githubusercontent.com/20047260/37681484-30008a4a-2c5d-11e8-81e8-5140992d3ae1.png)

![image](https://user-images.githubusercontent.com/20047260/37681508-45d2bb68-2c5d-11e8-9d28-70f3a7f200fd.png)

[...]

PR #472, opened 2018-03-20
Video Analytics Backend

adorsk:
Partially implements #402 

#### What's this PR do?
1. Adds backend utility functions to query Google Analytics for event stats for a particular video
2. Adds an API endpoint to expose analytics data.

1. In `<ROOT>/.env, set...
   a. GA_DIMENSION_CAMERA=dimension1
   b. GA_VIEW_ID=<the ID of an analytics view that provides video event data>
   c. GA_KEYFILE=<path to the keyfile for Google service account that has access to the analytics view>

2. After starting the app, go to `http://localhost:8089/api/v0/videos/<SOME VIDEO KEY>/analytics/`. You should see the DjangoRestFramework view, w/ analytics data corresponding to that video.

#### Known Issues
1. video keys that begin with a letter, rather than a number, appear to have that letter capitalized when event labels are sent to google analytics. This means that queries for videos with leading-letter keys fail to find matching events. >:/   I do not know yet where this capitalization is happening.

#### Suggested Reviewers
@mbertrand perhaps?


adorsk:
### Re: capitalization issue from above
The culprit was the react-ga library. See: https://github.com/react-ga/react-ga/issues/24.
Related: https://xkcd.com/979/

Fix is to initial react-ga object in `static/js/util/google_analytics.js` like this:
```
ga.initialize(SETTINGS.gaTrackingID, { debug: debug , titleCase: false})
```

ui/utils.py
```diff
@@ -170,3 +177,119 @@ def write_x509_files():
...
+            'Could not parse analytics response') from exc
+
+
+def get_google_analytics_client():
+    """Gets a Google Analytics client.
+
+    Returns:
+        analytics_client: An analytics client
+    """
+    _check_files_exist([settings.GA_KEYFILE])
```
mbertrand: The MIT certificate and private key files are generated on app startup [here](https://github.com/mitodl/odl-video-service/blob/6beb8c88b35247215f88f1f3312a18263285a825/ui/apps.py#L27).  The same would need to be done for the GA keyfile, with `settings.GA_KEYFILE` equaling the contents of the file (same as `MIT_WS_CERTIFICATE` and `MIT_WS_PRIVATE_KEY`), but in that case you can skip the file creation and use the method `ServiceAccountCredentials.from_json_keyfile_dict` instead of `ServiceAccountCredentials.from_json_keyfile_name`), something like this:

```
try:
    credentials = ServiceAccountCredentials.from_json_keyfile_dict(json.loads(settings.GA_KEYFILE))
```

ui/utils.py
```diff
@@ -170,3 +177,119 @@ def write_x509_files():
...
+    try:
+        credentials = ServiceAccountCredentials.from_json_keyfile_name(
+            settings.GA_KEYFILE)
+        analytics_client = build('analyticsreporting', 'v4',
+                                 credentials=credentials)
+        return analytics_client
+    except Exception as exc:  # pylint: disable=broad-except
+        raise GoogleAnalyticsException('Something went wrong with creating a'
```
mbertrand: Maybe add a unit test for this line to increase diff coverage

ui/utils.py
```diff
@@ -170,3 +177,119 @@ def write_x509_files():
...
+def get_video_analytics(video_key):
+    """Get video analytics data from Google Analytics."""
+    ga_client = get_google_analytics_client()
+    ga_response = ga_client.reports().batchGet(
+        body=generate_google_analytics_query(video_key)).execute()
+    try:
+        return parse_google_analytics_response(ga_response)
+    except Exception as exc:
+        raise GoogleAnalyticsException(
+            'Could not parse analytics response') from exc
```
mbertrand: Maybe add a unit test to cover this exception

ui/utils.py
```diff
@@ -170,3 +177,119 @@ def write_x509_files():
...
+    Returns:
+        analytics_query: a query dict suitable to use as the body of an
+        analytics 'batchGet' request.
+    """
+    if not settings.GA_VIEW_ID:
+        raise RuntimeError("Missing 'settings.GA_VIEW_ID'.")
+    ga_view_id = settings.GA_VIEW_ID
+    # https://developers.google.com/analytics/devguides/reporting/core/v3/reference
+    START_DATE = '2005-01-01'
+    END_DATE = '9999-01-01'
```
mbertrand: This will return all available data, which should be fine.  But it might be worth getting some clarification from @pdpinch on whether the graph should show all data by default, or just the last X days, and/or if the user should be able to specify a date range.

ui/utils.py
```diff
@@ -170,3 +177,119 @@ def write_x509_files():
...
+                    {'name': 'ga:eventAction'},
+                    {'name': 'ga:dimension1'}
+                ],
+                'dimensionFilterClauses': [
+                    {
+                        'filters': [
+                            {
+                                'dimensionName': 'ga:eventLabel',
+                                'operator': 'EXACT',
+                                'expressions': [video_key]
```
mbertrand: Good catch on the capitalization issue.  While adding `titleCase: false` to the react-ga initialization will stop the video keys from being capitalized, it will also cause a split in the analytics data for the videos that begin with a letter (before and after fix).  Another possible option is to use `video_key.capitalize()` here, which should avoid that data split.   Might be worth getting input from @pdpinch on the preferred approach.

adorsk: re: @mbertrand's note about splitting analytics data.

To me, the tradeoff here is between these things:
  
    1. CHANGE FRONT-END: We change our front-end code so that ga data is not altered.
        PROS:
            1. Future analytics data matches what we expect.
        CONS:
            1. We have to handle events that happen before this change. some options here:
                1. ignore 'legacy' events 
                2. change the backend to search for both new + legacy events (easy, just change backend GA query)
    2. DON'T CHANGE FRONT-END: We leave front-end code as-is.
        a. PROS:
            1. analytics data is consistent (no legacy vs new formats)
        b. CONS:
            1. We have to change the backend to search for legacy events (easy)
            2. If we add future analytics queries, we must remeber that event data is alteread by react-ga.


I personally favor <CHANGE FRONT-END> and ignoring legacy events, since our earliest events only start in 2017-10:
https://analytics.google.com/analytics/web/#report/content-event-overview/a5145472w162315911p163330947/%3F_u.date00%3D20150314%26_u.date01%3D20180320/

@pdpinch, whaddya think?

ui/utils.py
adorsk: leaving as-is for now (return all available data), since we can change this easily later.

pdpinch: As an MVP approach, I think we should return all data. I'm not expecting a lot of traffic so in order for our users to make any inference from the data, they'll probably need all of it.

ui/utils.py
pdpinch: I don't think we put a lot of care into deciding whether the events should be `t...` or `T...` so I think consistent data from Google is more valuable than lower-case t. 

That said, I would like to have the front-end modified so that it will be obvious to the next engineer what to expect. Can we specify that the events be upper-case? 

Does that make sense?

adorsk: Roger that, sticking with capitalized events.
I'll add a comments which tell future engineers that events get capitalized.

adorsk:
@pdpinch RE: KEYFILES
For this PR to work, we will need to add some secrets to Vault.

Specifically, we will need keyfiles for the service account(s) which have access to:

1. OVS (video) CI: https://analytics.google.com/analytics/web/#embed/report-home/a5145472w162315911p163330947/
1. OVS (video) Prod: https://analytics.google.com/analytics/web/#embed/report-home/a5145472w162315911p163330947/

Questions:
1. Do these service accounts exist?
    1. If no, how do we want to create these accounts/document them??
    2. If yes, can you send me their keyfiles via KeyBase?

adorsk:
@mbertrand 
Updated PR per your earlier comments. Can you review the changes?
(changing label to 'Needs Review')

ui/utils.py
```diff
@@ -170,3 +178,126 @@ def write_x509_files():
...
+
+def generate_google_analytics_query(video_key):
+    """Generates a Google Analytics query.
+
+    Returns:
+        analytics_query: a query dict suitable to use as the body of an
+        analytics 'batchGet' request.
+    """
+    if not settings.GA_VIEW_ID:
+        raise RuntimeError("Missing 'settings.GA_VIEW_ID'.")
```
mbertrand: See earlier comment about `MANDATORY_SETTINGS`

ui/utils.py
```diff
@@ -170,3 +178,126 @@ def write_x509_files():
...
+
+def get_google_analytics_client():
+    """Gets a Google Analytics client.
+
+    Returns:
+        analytics_client: An analytics client
+    """
+    try:
+        if not settings.GA_KEYFILE_JSON:
+            raise RuntimeError("missing 'settings.GA_KEYFILE_JSON'")
```
mbertrand: This works fine, but for consistency it might be better to remove this and add `GA_VIEW_ID` and `GA_KEYFILE_JSON` to [settings.MANDATORY_SETTINGS](https://github.com/mitodl/odl-video-service/blob/624ef5e5bd86783a312b7deb8a359b7f1df7e6c7/odl_video/settings.py#L414) instead.

ui/utils.py
```diff
@@ -170,3 +178,126 @@ def write_x509_files():
...
+                'viewId': ga_view_id,
+                'dateRanges': [{'startDate': START_DATE, 'endDate': END_DATE}],
+                'metrics': [{'expression': 'ga:totalEvents'}],
+                'dimensions': [
+                    # [adorsk, 2018-03]
+                    # Achtung! a high degree of implicit coupling to
+                    # dimension names that have been manually set in GA.
+                    # Hard-coding for now.
+                    {'name': 'ga:eventAction'},
+                    {'name': 'ga:dimension1'}
```
mbertrand: `dimension1` is available as a django setting (`settings.GA_DIMENSION_CAMERA`)

adorsk: Ah, nice catch. Updated here, and in test as well.

adorsk:
@mbertrand Some changes from the last review, can you take another look?

Also, related, secrets have been added to vault and put into salt: https://github.com/mitodl/salt-ops/pull/573

ui/utils_test.py
```diff
@@ -1,11 +1,21 @@
...
+
 import pytest
 from zeep.exceptions import Fault
 
 from ui import factories
-from ui.exceptions import MoiraException
-from ui.utils import write_to_file, get_moira_client, user_moira_lists, has_common_lists
+from ui.exceptions import MoiraException, GoogleAnalyticsException
+from ui.utils import (
+    write_to_file, get_moira_client, user_moira_lists,
```
mbertrand: very minor thing, feel free to ignore here if you prefer: if most methods are on separate lines, do them all that way for consistency:
```python
from ui.utils import (
    write_to_file, 
    get_moira_client, 
    user_moira_lists,
```

ui/views.py
```diff
@@ -334,6 +338,22 @@ class VideoViewSet(ModelDetailViewset):
         ui_permissions.HasVideoPermissions,
     )
 
+    @detail_route()
+    def analytics(self, request, key=None):
+        """get video analytics data"""
+        # pylint: disable=unused-argument
+        if 'throw' in request.GET:
+            return HttpResponse(status=500)
+        if 'mock' in request.GET:
```
mbertrand: Instead of putting this here, might it be possible to mock `get_video_analytics` in `views_test.py`, to keep test code and app code separate? Something like:
```python
mocker.patch('ui.views.get_video_analytics', side_effect= generate_mock_video_analytics_data)
```
and place the `generate_mock_video_analytics_data` function in the same test file?

ui/utils.py
```diff
@@ -170,3 +179,148 @@ def write_x509_files():
...
+        times.add(time_)
+        channels.add(channel)
+    return {
+        'times': sorted(list(times)),
+        'channels': sorted(list(channels)),
+        'views_at_times': views_at_times,
+    }
+
+
+def generate_mock_video_analytics_data(n=24, seed=42):
```
mbertrand: Could this go in one of the test files instead (like `views_test.py`)?

ui/views.py
adorsk: I put this 'mock' conditional in here for doing integration testing w/ front-end code. (for testing from the client-end sans the GoogleAnalytics query in the back-end).

Perhaps I should remove this, or at least explain its purpose better?

ui/utils.py
adorsk: I think this depends on whether we decide to remove the 'mock' parameter (per above comment).

If we decide to remove it, I agree, we should move this out.

If we decide to leave it, I think we should keep this here, because it's part of a 'real' view.

ui/views.py
mbertrand: Okay, I assumed it was for unit-testing.  Since it has a purpose outside of that I'm fine with leaving it here.

ui/utils_test.py
adorsk: sounds good, updated to put methods on separate lines

adorsk:
also moved GA settings into mandatory settings, and removed corresponding checks in utils functions.

ui/views.py
adorsk: Sounds good, I'll leave it in and add a descriptive comment.

ui/utils.py
mbertrand: Forgot that for anything in `MANDATORY_SETTINGS`, a dummy value has to be placed in the `tox.ini` file, under the `setenv=` section.  Once you add these the travis tests should pass.

mbertrand: GA_DIMENSION_CAMERA, GA_KEYFILE_JSON, GA_VIEW_ID

adorsk:
@mbertrand I think we're all set now (added settings to the tox.ini, tests pass).

Anything else?

mbertrand:
@adorsk it's good to go, last step is to squash commits before merging: https://github.com/mitodl/handbook/blob/master/development-process.md#step-3-merging

adorsk:
@mbertrand excellent,  I'll squash down, and then merge it in. Thanks for your comments and patience with this one. Hopefully future PRs will be smaller!

PR #478, opened 2018-03-22
Switch from celery.get_task_logger() to logging.getLogger() for tasks

### LINKED ISSUE #462: raise to Sentry when copying to AWS fails
See the [thread in Slack](https://mitodl.slack.com/archives/G5T80SA7L/p1520612987000634):

This error appears in the logs:

```
[2018-03-09 16:45:19,000: ERROR/ForkPoolWorker-3] cloudsync.tasks.monitor_watch_bucket[05dd3d90-25a4-4e93-b9b4-5ade645aa313]: AWS error when ingesting file from watch bucket MIT-MIT-Math FC-lec-mit-0000-2018mar08-1635-MM.mp4: {'source_url': ['Enter a valid URL.']}
```

But it didn't get raised to Sentry. It would be nice to be notified about errors like this.

mbertrand:
- Fixes #462 
- Removes a stray `log.exception()` line

#### What's this PR do?
Replaces `celery.get_task_logger` with `logging.getLogger`

Nothing should break.
To test that logged exceptions make it to sentry:
- Set `SENTRY_DSN` in your `.env` file to the correct value
- Modify `cloudsync.tasks.monitor_watch_bucket` to always log an exception:
   `log.exception("This is a test")`)
- In a shell, run `monitor_watch_bucket.delay()`
- Check Sentry for your error

### DIFF
--- a/cloudsync/youtube.py
+++ b/cloudsync/youtube.py
@@ -255,5 +255,4 @@ class YouTubeApi:
         Returns:
             int: 204 status code if successful
         """
-        log.exception("Deleting YT video id %s", video_id)
         return self.client.videos().delete(id=video_id).execute()
--- a/cloudsync/tasks.py
+++ b/cloudsync/tasks.py
@@ -1,6 +1,7 @@
 """
 Tasks for cloudsync app
 """
+import logging
 import os
 import re
 from urllib.parse import unquote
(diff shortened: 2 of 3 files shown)


cloudsync/youtube.py
```diff
@@ -255,5 +255,4 @@ def delete_video(self, video_id):
         Returns:
             int: 204 status code if successful
         """
-        log.exception("Deleting YT video id %s", video_id)
```
mbertrand: This shouldn't be here

PR #479, opened 2018-03-23
Make `ENABLE_VIDEO_PERMISSIONS` affect front-end video edit form only

### LINKED ISSUE #474: Remove `ENABLE_VIDEO_PERMISSIONS` requirement for Youtube & public URL functionality
Currently, the `ENABLE_VIDEO_PERMISSIONS` setting must be set to `True` for the following to happen:
  - Display options on the video UI form to override collection permissions, make videos public
  - Sync public videos with Youtube
  - Allow public URL's to be viewed

This should be changed so that the setting only affects the video edit UI form.

mbertrand:
Closes #474

#### What's this PR do?
- Removes all the `if settings.ENABLE_VIDEO_PERMISSIONS` statements affecting YouTube uploads and permissions, leaves it only for the front-end VideoEditForm. 

- Set `ENABLE_VIDEO_PERMISSIONS=False` in your `.env` file
- Change a video to `is_public=True` on its admin page
- In a shell, run the following:
    ````
    from cloudsync.tasks import upload_youtube_videos
    upload_youtube_videos.delay()
    ````
- After a few minutes, go to the `Youtube Videos` admin page.  There should be a `YouTubeVideo` object for your `Video`, and its `id` should not be `None` after it has completed uploading. 
- In an incognito tab, visit the video detail page.  It should play from YouTube (look for the logo at bottom right), and not require login.
- Log in and click the 'EDIT' button on the video detail page.  You should not see any options for overriding the collection permissions.
- Upload a subtitle file for the video.
- Wait a minute then go to `https://www.youtube.com/watch?v=<YoutubeVideo.id>`, there should be a `CC` button on the video toolbar that when activated shows the captions you uploaded.
- Return to the admin page for the video, and set `is_public` to `False`.  
- Check the `YouTube Videos` admin page, there should no longer be a `YouTubeVideo` object there for that `Video`
- Check `https://www.youtube.com/watch?v=<YoutubeVideo.id>` again, you should see `This video is unavailable`
- Set `ENABLE_VIDEO_PERMISSIONS=True` in your `.env` file and restart the docker containers.
- Go to the video detail page and click 'EDIT'.  You should see options for overriding the collection permissions.  If you make it public and save it, the video should be uploaded to Youtube again.


odl_video/settings.py
```diff
@@ -438,15 +438,8 @@
...
-
-if ENABLE_VIDEO_PERMISSIONS:
-    MANDATORY_SETTINGS += [
-        'YT_ACCESS_TOKEN',
-        'YT_REFRESH_TOKEN',
-        'YT_CLIENT_ID',
-        'YT_CLIENT_SECRET',
-        'YT_PROJECT_ID',
+    'YT_ACCESS_TOKEN',
+    'YT_REFRESH_TOKEN',
```
gsidebo: Just wondering: why are these two settings required to run the app, and why are the other 3 YT not required?

mbertrand: The optional 3 settings are used only when running the `oauthtokens` command, which is what will give you values for `YT_ACCESS_TOKEN` and `YT_REFRESH_TOKEN`.  These two are required for handling all Youtube API requests.

mbertrand: Actually `YT_CLIENT_ID` and `YT_CLIENT_SECRET` are required too, not sure how I missed that.  Only `YT_PROJECT_ID` is optional.  Will update.

mbertrand: Updated `app.json` and `settings.py` to indicate that `YT_CLIENT_ID` and `YT_CLIENT_SECRET` are required.

PR #493, opened 2018-03-27
Handle possible YT video dupes while checking statuses

### LINKED ISSUE #486: Duplicate Youtube videos cause an IndexError when checking their status
If a Youtube video is flagged as a duplicate, the API will not return any information about it, as if it doesn't exist.  This will cause an `IndexError` during the `update_youtube_statuses` task.  This exception should be caught and logged, and the status of the `YoutubeVideo` should be changed to `failed` so that it won't keep repeating.

mbertrand:
Fixes #486 

#### What's this PR do?
Catches `IndexError` when checking the status of a `YoutubeVideo` and sets the status to `YoutubeStatus.FAILED`.

- In the admin page for a video, set `is_public` to `True`
- in a django shell:
  ```python
  from cloudsync.tasks import upload_youtube_videos
  upload_youtube_videos.delay()
```
- Wait a few minutes or so for the video to upload to Youtube.
- Check the status of the `YoutubeVideo` object, hopefully it is `processed`
- Repeat the above steps, uploading the same video again.
- This time the status of the new ``YoutubeVideo` should be `failed`
- There should be an exception logged by celery.


aliceriot:
will this work with an existing video which is already uploaded? I have a video which is uploaded to YouTube already, and which has `status: processed`. I ran the `upload_youtube_videos` command, but I don't see another YouTube video object and the status on that one is still processed - not sure if I messed something up, haha.

aliceriot:
I don't see a celery exception either

mbertrand:
@aliceriot I think you might need `ENABLE_VIDEO_PERMISSIONS=True` in your `.env` file, and you need to create a new `Video` (same videofile as the previous one if you still have it, or just upload a new one twice).  Then run `upload_youtube_videos.delay()`

aliceriot:
Ok, I'll give it another try

PR #528, opened 2018-04-06
fix text field regressions from mdc upgrade

adorsk:
https://github.com/mitodl/odl-video-service/pull/519
https://github.com/mitodl/odl-video-service/pull/513

#### What's this PR do?
Fixes regression introduced in https://github.com/mitodl/odl-video-service/pull/513
Regression was text fields were no longer full width.

Visit a section of the app that has text fields. Confirm that the text fields are full width.


aliceriot:
guess I should have looked a little closer when reviewing #513 :woman_facepalming:

Issue #377, opened 2017-11-28
Add a quality selector button to VideoJS


mbertrand:
Add a button that allows users to manually select a video quality.  An available plugin is here: https://www.npmjs.com/package/videojs-hls-quality-selector

pdpinch:
Just checking in on this: am I right that we have a quality selector for mp4 videos, but not for HLS?

mbertrand:
Correct.

mbertrand:
Also should look into changing the default bitrate for multiangle videos to the highest available.

pdpinch:
I opened a separate issue for changing the default bitrate: https://github.com/mitodl/odl-video-service/issues/458

I'm also going to move this into the backlog.

PR #556, opened 2018-04-19
Dorska/quality selector

adorsk:
#377 

#### What's this PR do?
Adds a new control to the VideoPlayer control bar that allows the user to select a video quality setting.

  1. Ensure that you have a video that's at least a few minutes long. I'm a fan of https://ia802702.us.archive.org/16/items/fireplace/FireplaceGlenLakeMi.11-06-07_512kb.mp4 .
  2. Check that the selected bitrate changes after ten seconds:
     1. Open up dev tools, and activate (1) 'Network Conditions' section and (2) the 'Network' tab.
     2. Start the video playing. you should start to see lines in the Network tab like `video_...-2gfh90...ts` . The "`2gfh90`" indicates that the highest bitrate is being used.
     3. Let ten seconds go by, and then in the 'Network Conditions' tab, set 'Network throttling' to a slower value (like 'slow 3G').
     4. Ensure that you see the lines in the 'Network' tab change. Probably to something like `video_...-06dkm6...ts`.
  3. Check that the you can set the bitrate.
     1. Reload the page, and start the video playing again.
     2. Let ten seconds go by.
     3. Now click the 'HD' button in the video player controls and select a value.
     4. Ensure that the 'Network' tab shows different `.ts` lines.
     5. Ensure that different selections in the menu result in different `.ts` lines.
  4. Check that the auto bitrate works.
     1. Reload the page, and start the video playing again.
     2. Let ten seconds go by.
     3. Now click the 'HD' button in the video player controls and select a value.
     4. Then select the 'auto' value.
     5. Let ten seconds go by, and then in the 'Network Conditions' tab, set 'Network throttling' to a slower value (like 'slow 3G').
     6. Ensure that you see different `.ts` lines.
     7. Disable the 'Network throttling', and ensure that you see different `.ts` lines.

Normally videojs adaptively determines which bitrate to use, per:
https://github.com/videojs/videojs-contrib-hls/blob/master/docs/bitrate-switching.md
We can override the normal behavior by toggling active bitrates. The videojs-contrib-quality-levels plugin provides an API for doing this toggling.
And in this PR we add a control that allows humans to interact with that API.

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -286,17 +286,31 @@ class VideoPlayer extends React.Component<*, void> {
     })
   }
 
-  selectPlaylist = (opts: { defaultPlaylistSelector: () => any }) => {
-    const { defaultPlaylistSelector } = opts
+  selectPlaylist = () => {
+    const sortByBandwidth = R.sortBy(R.path(["attributes", "BANDWIDTH"]))
+    const playlists = sortByBandwidth(
+      this.player.tech_.hls.playlists.master.playlists
+    )
     // Always start with highest bandwidth for first 10 seconds
     if (this.player.tech_.currentTime() < 10) {
-      const sortByBandwidth = R.sortBy(R.path(["attributes", "BANDWIDTH"]))
-      const playlists = sortByBandwidth(
-        this.player.tech_.hls.playlists.master.playlists
-      )
       return _.last(playlists)
     }
-    return defaultPlaylistSelector()
+    // Return active playlist with highest bandwidth <= system bandwidth,
+    // or first active playlist otherwise.
+    const activePlaylists = R.filter((rep) => !rep.disabled, playlists)
+    return (
+      _.last(
+        R.filter(
+          ((rep) => {
+            return rep.attributes.BANDWIDTH <= _.max([
+              this.player.tech_.hls.systemBandwidth,
+              playlists[0].attributes.BANDWIDTH
+            ])
+          }),
+          activePlaylists
+        )
+      ) || activePlaylists[0]
+    )
   }
 
   selectInitialQualityLevel() {
@@ -319,7 +333,7 @@ class VideoPlayer extends React.Component<*, void> {
     }
     const useYouTube = video.is_public && video.youtube_id !== null
     this.lastMinuteTracked = null
-    const self = this
+    const selectPlaylist = this.selectPlaylist.bind(this)
     this.player = videojs(
       this.videoNode,
       makeConfigForVideo(video, useYouTube, embed),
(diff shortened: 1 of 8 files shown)


pdpinch:
@adorsk if this is ready for review, can you add the "Needs Review" label? That's our most reliable signal that a PR is ready.

static/js/components/VideoPlayer.js
```diff
@@ -285,27 +286,24 @@ class VideoPlayer extends React.Component<*, void> {
...
+    const { defaultPlaylistSelector } = opts
     // Always start with highest bandwidth for first 10 seconds
     if (this.player.tech_.currentTime() < 10) {
+      const sortByBandwidth = R.sortBy(R.path(["attributes", "BANDWIDTH"]))
+      const playlists = sortByBandwidth(
+        this.player.tech_.hls.playlists.master.playlists
+      )
       return _.last(playlists)
     }
-    // Return playlist with highest bandwidth <= system bandwidth
```
mbertrand: This overrides the native HLS algorithm so that the video player's size in the browser is ignored; and should still be used unless someone manually selects a quality level.

static/js/components/VideoPlayer.js
```diff
@@ -339,7 +337,12 @@ class VideoPlayer extends React.Component<*, void> {
           })
         })
         if (this.tech_.hls !== undefined) {
-          this.tech_.hls.selectPlaylist = selectPlaylist
+          const _originalSelectPlaylist = this.tech_.hls.selectPlaylist
```
mbertrand: This was overriden in a previous PR so that the size of the video in the browser is ignored when determining the optimal resolution to play.  Can you revert to using that custom algorithm instead of the default one?

adorsk: Ah, I see what you mean.

Updated to use a modified version of the previous custom algorithm. The modification is: the algorithm will now select the highest *active* playlist that is less than the available system bandwidth, or the first active playlist if no active playlist is <= system bandwidth.

The quality selector widget sets all non-selected playlists to disabled. If 'auto' is selected it sets all playlists to active.

static/js/components/VideoPlayer.js
```diff
@@ -294,34 +295,45 @@ class VideoPlayer extends React.Component<*, void> {
...
+              playlists[0].attributes.BANDWIDTH
+            ])
+          }),
+          activePlaylists
+        )
+      ) || activePlaylists[0]
     )
   }
 
+  selectInitialQualityLevel() {
```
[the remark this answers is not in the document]
adorsk: ah, nice catch. removed.

static/js/lib/videojs_hls_quality_selector.js
```diff
@@ -0,0 +1,135 @@
+/* Adapted from https://github.com/chrisboustead/videojs-hls-quality-selector .
```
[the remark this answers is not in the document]
adorsk: I wasn't sure on whether this would be worth it. The original plugin just tests that it can register against videojs:
https://github.com/chrisboustead/videojs-hls-quality-selector/blob/master/test/plugin.test.js

We could write tests, but I think it would mostly look like a lot of mocking for videojs functions.

Do you think it's worth pursuing? Either way is fine with me.

adorsk:
@mbertrand reverted `requirements.txt` to take out `-e` flags, should work with travis now per bug fix from pip 10.0.1 (https://github.com/pypa/pip/issues/5251).

Ready for review again.

adorsk:
Closed #377.

Issue #463, opened 2018-03-12
display an error in the front-end when API calls fail


pdpinch:
On the collection page, when the API call to list video fails, we display the misleading message: "You have not added any videos yet"

Instead, we should:

- [ ] raise an exception to Sentry
- [ ] display an error along the lines of "Sorry, we were unable to load the data necessary to process your request. Please reload the page. If this happens again, please contact `<support email>`"

pdpinch:
@roberthouse54 Should we use a toast message for this?

can you give me a 2nd opinion on the text of the error message?

roberthouse54:
I think we could go with something like:

"Oops, something went wrong. We were not able to load the page. Please refresh the page. If this happens again, please contact <support email>."

I think a toast message could work.

adorsk:
This is how I'm interpreting the expected behavior, based on above comments:

### Case 1: error unrelated to permissions
1. log sentry exception
2. render this message: "Sorry, we were unable to load the data necessary to process your request. Please reload the page. If this happens again, please contact <support email>"

### Case 2: error related to permissions
1. Render this message: "You don't have permissions to view this collection."

adorsk:
#570 and #571 have been merged and close this issue.

adorsk:
Correction, still waiting on #571

adorsk:
#571 finished, closing here.

PR #570, opened 2018-04-27
add error message for collections page

adorsk:
Partially implements #463 .

#### What's this PR do?
1. Creates an `ErrorMessage` component.
2. Displays this component with a message on the `CollectionsListPage` when there is an api failure.

1. Go to the collections page `/collections/`
2. Block the url for the collections api. See [this page](    https://stackoverflow.com/questions/27863094/how-to-block-a-url-in-chromes-developer-tools-network-monitor
) for info on how to block.
3. Reload the page.
4. Ensure that the error message looks reasonable.

### DIFF
--- a/static/js/containers/CollectionListPage.js
+++ b/static/js/containers/CollectionListPage.js
@@ -18,6 +18,7 @@ import type { Collection, CollectionsPagination } from "../flow/collectionTypes"
 import withPagedCollections from "./withPagedCollections"
 import LoadingIndicator from "../components/material/LoadingIndicator"
 import Paginator from "../components/Paginator"
+import ErrorMessage from "../components/ErrorMessage"
 
 export class CollectionListPage extends React.Component<*, void> {
   props: {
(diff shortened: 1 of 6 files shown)


static/js/containers/CollectionListPage.js
```diff
@@ -87,7 +88,14 @@ export class CollectionListPage extends React.Component<*, void> {
       return null
     }
     if (currentPageData.status === "ERROR") {
-      return <div className="collection-list-page-error">Error!</div>
+      const supportEmail = SETTINGS.support_email_address || ""
+      return (
+        <ErrorMessage>
+          <p>Sorry, we were unable to load the data necessary to process your request. Please reload the page.</p>
```
mbertrand: This content is pretty generic, maybe it could be the default in the `ErrorMessage` component itself?  And customizable if we want to display something slightly different on certain pages?

adorsk: @mbertrand  Hmm...I would prefer to keep the `ErrorMessage` component itself simple, rather than have logic in the `ErrorMessage` component to check whether it was given content or not.

Perhaps we can wait to create generic messages until we have another use case for this particular message?

PR #571, opened 2018-04-27
Dorska/collection 403

adorsk:
Depends on #570 . Wait for that to merge before reviewing.

Partially implements #463 .

#### What's this PR do?
Adds an error message to the `CollectionDetailPage`

1. Ensure that you have (a) a collection, and (b) a user who can *not* access that collection.
2. Log in as the unauthorized user and go to the Collection detail page.
3. You should see an error message.
4. Ensure that the Collection detail page renders as expected for authorized users.

### DIFF
--- a/static/js/components/errorMessages.js
+++ b/static/js/components/errorMessages.js
@@ -1,0 +1,22 @@
+// @flow
+/* global SETTINGS: false */
+import React from "react"
+import ErrorMessage from "./ErrorMessage"
+
+
+export const UnableToLoadData = () => {
+  // $FlowFixMe
+  const supportEmail:string = SETTINGS.supportEmail
+  return (
+    <ErrorMessage>
+      <p>Sorry, we were unable to load the data necessary to process your request. Please reload the page.</p>
+      {
+        supportEmail ? (
+          <p>If this happens again, please contact&nbsp;
+            <a href={`mailto:${supportEmail}`}>{supportEmail}</a>.
+          </p>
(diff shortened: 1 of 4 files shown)


adorsk:
@mbertrand re: "unknown error" message, this our placeholder when the response has no error text:
https://github.com/mitodl/odl-video-service/blob/981095f959b7068502e04cd9ac659340ff8db7db/static/js/containers/CollectionDetailPage.js#L136 .

I think this is a side-effect of blocking the url, via the chrome dev tools. In theory, we should always get an error message back when the error is coming from Django.

Do you think this is sufficient? Or should I modify the text here?

mbertrand:
@adorsk My assumption based on the issue description is that the error should be the same/similar as to what's displayed on the collections page when this happens: "Sorry, we were unable to load the data necessary to process your request. Please reload the page. If this happens again, please contact ".  

 "Unknown Error" doesn't seem very informative to the end user.  I also get this same message if I artificially throw an exception, for instance by adding 
```python
raise Exception('SOMETHING TERRIBLE HAS HAPPENED')
```
to `CollectionViewSet.get_queryset`

adorsk:
@mbertrand ah, I see what you mean about there being no message when there's a 500 error. (which is surprising to me, I had thought django + hammock would pipe us down a message body).

I consolidated the error  message from #570 , and added it here.

Ready for review again.

adorsk:
squashed and merged.

PR #587, opened 2018-05-01
Per-user moira list cache

### LINKED ISSUE #518: Improve efficiency of permission checking via moira lists
Checking for permission via moira-lists every time a user opens a video or collection page is inefficient, and could slow down the page load if there are multiple lists to be checked.  The following might help:

- Cache the moira-lists that a user has view/edit access to for a certain amount of time (a few hours?)
- If not cached yet, check multiple moira-lists concurrently instead of consecutively.

mbertrand:
- Fixes #518 

#### What's this PR do?
- Implements a per-user cache of moira lists that a user does and does not have access to, in order to reduce the frequency of moira service calls.

- There should be no obvious change to the user in how permissions work.
- Create a collection with `odl-engineering` as the view or admin moira list.
- In django admin, create a staff (but non-superuser) account for someone on the `odl-engineering` moira list (most likely your own MIT email)- username and email should both be `<name>@mit.edu`
- In an incognito browser window, log in as that non-superuser and go to a video detail page from that collection above.  You should be able to view it.  If `odl-engineering` is the admin moira list, you should be able to go to the collection detail page.  
- Add the following to your `hosts` file:
  ```
  127.0.0.1       moiraws.mit.edu
  ```
- Now go to the other video detail pages in the collection as the non-superuser.  The video detail page should load without any errors because the moira lists you have access to should already be cached.
- Check out the master branch from git and try the above step again.  Now you should get a `MoiraException`.
- Remove the change to your `hosts` file.


ui/utils.py
```diff
@@ -65,27 +82,65 @@ def get_moira_user(user):
...
+    Args:
+        user (django.contrib.auth.User): the Django user.
+
+    Returns:
+        UserMoiraMembership: An object containing all known lists the user does & doesn't have access to.
+    """
+    return cache.get(MOIRA_CACHE_KEY.format(user_id=user.id)) or query_moira_lists(user)
+
+
 def has_common_lists(user, list_names):
```
adorsk: This function now has a side effect of altering the moira_cache. It may make sense to note this in the function's docstring. Your call.

PR #598, opened 2018-05-02
add timestamps to models

### LINKED ISSUE #525: Most/all OVS models should have `created_on` and `modified_on` fields
Add the Micromasters `TimeStampedModel` to OVS, and make most/all OVS models inherit it.  Some models have an existing `created_at` field - either change it to `created_on` or use`created_at` in the `TimeStampedModel`.

From MicroMasters:
```python
class TimestampedModel(Model):
    """
    Base model for create/update timestamps
    """
    objects = TimestampedModelManager()
    created_on = DateTimeField(auto_now_add=True)  # UTC
    updated_on = DateTimeField(auto_now=True)  # UTC

    class Meta:
        abstract = True
```

adorsk:
Fixes #525 

#### What's this PR do?
Adds timestamp fields (`created_at`, `updated_at`) to all models in ovs.

Confirm that the models for `mail`, `ui`, and `techtv2ovs` all have `created_at`/`updated_at` fields.

Note that only a few models currently expose timestamp fields in the Django admin. This PR does not alter the admin.


odl_video/models.py
```diff
@@ -0,0 +1,62 @@
...
+        if "updated_on" not in kwargs:
+            kwargs["updated_on"] = now_in_utc()
+        return super().update(**kwargs)
+
+
+class TimestampedModelManager(Manager):
+    """
+    Subclassed manager for TimestampedModel
+    """
+    def update(self, **kwargs):
```
mbertrand: This method (and `TimestampedModelQuerySet.update`) don't seem to be covered by unit tests yet, can you add one to cover both when `updated_at` is and isn't in `kwargs`?  That should bring up the code coverage.

ui/models.py
```diff
@@ -76,7 +77,7 @@ def all_admin(self, user):
             models.Q(owner=user)).distinct()
 
 
-class Collection(models.Model):
+class Collection(TimestampedModel):
```
mbertrand: The default ordering for `Collection` is `-created_at`, which is probably still fine, but maybe check with @pdpinch to see if he'd prefer `updated_at` instead.

ui/models.py
```diff
@@ -120,14 +120,13 @@ def for_owner(cls, owner):
         return cls.objects.filter(owner=owner)
 
 
-class Video(models.Model):
+class Video(TimestampedModel):
```
mbertrand: The default ordering for `Video` is `-created_at`, which is probably still fine, but maybe check with @pdpinch to see if he'd prefer `updated_at` instead.

odl_video/models.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    Subclassed QuerySet for TimestampedModelManager
+    """
+    def update(self, **kwargs):
+        """
+        Automatically update updated_on timestamp when .update(). This is because .update()
+        does not go through .save(), thus will not auto_now, because it happens on the
+        database level without loading objects into memory.
+        """
+        if "updated_on" not in kwargs:
```
mbertrand: Replace `updated_on` with `updated_at`

adorsk: ah, nice catch.

ui/models.py
mbertrand: I think it would affect the order of the collections returned, so instead of the most recently created collections at the top, it would be the most recently updated.

adorsk:
@mbertrand ready for review again.

PR #600, opened 2018-05-02
add status messages to embed page

### LINKED ISSUE #178: Embed video doesn't show 'Video is processing, check back later' text
Maybe one way to work around this is to hide the embed text until this is available

adorsk:
Fixes #178 .

#### What's this PR do?
Adds status messages to the embed page.

1. Go to the embed page for a video, ensure that it renders normally.
2. Change the video's status to 'error', ensure that the embed page shows a reasonable message.
3. Change the video's status to 'uploading', ensure that the embed page shows a reasonable message.

### DIFF
--- a/static/js/containers/VideoEmbedPage.js
+++ b/static/js/containers/VideoEmbedPage.js
@@ -8,8 +8,9 @@ import { actions } from "../actions"
 import VideoPlayer from "../components/VideoPlayer"
 import type { Video, VideoUiState } from "../flow/videoTypes"
 import { initGA, sendGAPageView } from "../util/google_analytics"
+import { videoIsProcessing, videoHasError } from "../lib/video"
 
-class VideoEmbedPage extends React.Component<*, void> {
+export class VideoEmbedPage extends React.Component<*, void> {
   props: {
     dispatch: Dispatch,
     video: Video,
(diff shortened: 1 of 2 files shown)


aliceriot:
functionality works!

static/js/containers/VideoEmbedPage.js
```diff
@@ -27,28 +28,71 @@ class VideoEmbedPage extends React.Component<*, void> {
   }
 
   render() {
```
aliceriot: this is kind of a style thing, but I prefer to have `render` be the last method declared on a React component - would you mind terribly moving it to the bottom of the class? :smile:

static/js/containers/VideoEmbedPage.js
```diff
@@ -27,28 +28,71 @@ class VideoEmbedPage extends React.Component<*, void> {
...
+
+  renderBody () {
+    const { video, videoUi } = this.props
+    const videoStatus = this.getVideoStatus(video)
+    if (videoStatus === 'PROCESSING') {
+      return this.renderProcessingMessage()
+    }
+    else if (videoStatus === 'ERROR') {
+      return this.renderErrorMessage()
+    }
```
aliceriot: why not just check `videoIsProcessing(video)` and `videoHasError(video)` right here, instead of having the separate `getVideoStatus` function?

adorsk:
@aliceriot ready for review again.

static/js/containers/VideoEmbedPage.js
aliceriot: In the tests there's a `video` object created, can't you just mutate that to make the different test cases?

aliceriot: Hm, I suppose there's a trade-off there. To me, writing the proxy function just to enable easier testing is undesirable, because we're trading off code readability for ease of testing.  But if you feel strongly this is the right way to go about it I'm fine with it.

PR #611, opened 2018-05-04
change 'Only me' => 'Only owner', to clarify permissions behavior

### LINKED ISSUE #444: 'Only me' permission is confusing if set by a user who is not the owner
To reproduce:
 - create a collection
 - change the edit permission on the collection to the `odl-engineering` moira list
 - create a user who is on that moira list and log in as them
 - As that user change the permission of the collection to 'Only me'

'Only me' makes me think it will reassign the ownership to the logged in user. Instead we should change the text to 'Only the owner'`, at least for other users

adorsk:
Fixes #444 .

#### What's this PR do?
1. Relabels 'Only me' => 'Only owner', to reflect actual permissions behavior. 

1. Confirm that labels in Collection edit dialog say 'Only owner'.
2. Confirm that labels in Video edit dialog say 'Only owner'.

### DIFF
--- a/static/js/components/dialogs/CollectionFormDialog.js
+++ b/static/js/components/dialogs/CollectionFormDialog.js
@@ -195,7 +195,7 @@ class CollectionFormDialog extends React.Component<*, void> {
             <h4>Who can view videos?</h4>
             <Radio
               id="view-only-me"
-              label="Only me"
+              label="Only owner"
               radioGroupName="view-perms"
               value={PERM_CHOICE_NONE}
               selectedValue={collectionForm.viewChoice}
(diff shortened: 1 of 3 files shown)


noisecapella:
Can you tweak the CSS to give a bit more space to the text?
screenshot from 2018-05-04 15-29-34

We should also change the 'Only me' below

adorsk:
@noisecapella ah, nice catch on the other 'only me' label.

- Updated so that all labels should now be 'only owner'.

I don't understand what you mean about the CSS spacing. Do you mean between the label and the text input?

noisecapella:
I mean the 'Only owner' is wrapped to two lines instead of staying on one. It looks like `generalStyles.scss` sets a `88px` width for labels, maybe that should be increased a bit? cc @roberthouse54

adorsk:
@noisecapella re: CSS, I see what you mean now (on my screen it's < 88px).

I changed the label width to 'auto', which should fix this.

adorsk:
squashing & then merging.

PR #621, opened 2018-05-08
Fix login redirect

### LINKED ISSUE #614: Bring back the login url/view
Revert the removal of `/login` URL/view and make that the default `LOGIN_URL` regardless of whether or not Touchstone is enabled.

### LINKED ISSUE #620: After authenticating via Touchstone, the login page doesn't redirect to `next` URL
- Go to `/collections` URL
- Log in via Touchstone.
- Redirected to `/login/?next=/collections` instead of `/collections`, and login form is visible.

The login view needs to be modified to redirect an authenticated user to the `next` URL if present or `/collections` otherwise.

mbertrand:
- Fixes #614
- Fixes #620

#### What's this PR do?
- Brings back the login view; if an authenticated user goes to the login page (as happens after Touchstone login), redirects user to the url specified by the `next` parameter if present or `/collections` otherwise.

- In an incognito tab, go to `/collections`.  You should be redirected to the login page.  Fill out the form and submit.  You should be redirected to `/collections`.
- While still logged in, go to `/login/?next=/collections/<collection_id>`.  You should be redirected to that collection detail page.
- While still logged in, go to `/login/`.  You should be redirected to the `/collections` page.

### DIFF
--- a/ui/templates/registration/login.html
+++ b/ui/templates/registration/login.html
@@ -7,16 +7,7 @@
 {% if form.errors %}
 <p>Your username and password didn't match. Please try again.</p>
 {% endif %}
-
-{% if next %}
-    {% if user.is_authenticated %}
-    <p>Your account doesn't have access to this page. To proceed,
-    please login with an account that has access.</p>
-    {% else %}
-    <p>Please login to see this page.</p>
-    {% endif %}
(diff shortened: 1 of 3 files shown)


blarghmatey:
:+1:

Issue #503, opened 2018-04-02
add paging UI to collection page


pdpinch:

adorsk:
### Implementation Note Re: url format
I'm planning to put the paged videos endpoint at (a) `/videos/?collection_key=<some_key`, rather than (b) `/collections/<some_key>/videos/`.

I choose (a) because:
1. It's cleaner in the code.
2. It gives us more flexibility later. For example, if we want to show a user all of her videos across collections, not just in a specific collection.

pdpinch:
I'm concerned about changing URLs. We have existing collection URLs out in the wild (e-mail, web pages) as well as Google Analytics history. 

We should at least look back at the decision-making process for the /collections/ end point to remind ourselves what our reasoning was for that.

adorsk:
@pdpinch Sorry, I should have clarified. [My comment above](https://github.com/mitodl/odl-video-service/issues/503#issuecomment-387464164) isn't about changing existing URLs; instead it's about creating a new URL that lists videos.

roberthouse54:
Design is in #521

PR #622, opened 2018-05-08
add videos pagination backend

adorsk:
#503 

#### What's this PR do?
1. Adds backend logic for videos pagination.

Note that there are no front-end changes yet.

1. Ensure that you have...
   a. a superuser with >= 3 videos
   b. a non-superuser with >= 3 videos
2. in `odl_video/settings.py`, find the line for `PAGE_SIZE_COLLECTIONS`, and make the default value be `2`.
3. As the superuser, go to http://localhost:8089/api/v0/videos/ . You should see all the videos, in paged format.
4. As the non-superuser, go to http://localhost:8089/api/v0/videos/ . You should see only videos for the non-superuser (or public videos, or videos which moira list criteria).
5. try different values for parameters `page_size=`, `page=` and `ordering=` (`title`, `-title`, `created_at`, `-created_at`) - the results should update accordingly.

The meaty bit of this PR involved configuring/testing video permissions. However, this work will give us the infrastructure to do private videos (videos visible to only their owner), should we choose to pursue private videos.


adorsk:
ah, nice catch. This becomes a little more complicated than I thought at first. But perhaps still tractable.

Per our IRL convo just now, I'm going to pursue the track of getting a user's full list hierarchy.

It looks like the Moira web service offers an endpoint that can give us this. See: https://moiraws.mit.edu/moiraws/services/moira?wsdl, `getUserListMembership`; this endpoint has a `recursiveSearch` option that allows for returning a user's fully-resolved list hierarchy.

That is, if there is a list structure like this:
```
- List-Grandparent
   - List-Parent
     - List-Child
        - Member-Bob
```
then `getUserListMembership(Bob, recursiveSearch=True)` will return `[List-Grandparent, List-Parent, List-Child]`.

We can then compare a user's fully-resolved lists against the lists specified in videos/collections.
(this should also cut down on Moira requests too)

adorsk:
Now that #625  has landed, this is ready for review again.

Issue #521, opened 2018-04-06
Collections pagination


pdpinch:
As an admin with access to a lot of collections, I need to be able to load the home page (/collections) and the side nav without waiting for 1000+ items. 

In order to minimize design changes and not block the import of TechTV collections, we can add simple pagination to the home page, and limit the sidenav to the first 10 items. We'll add a "more" link in the sidenav that will just load the /collections page.

### acceptance criteria

- [x] add pagination to the collections API
- [x] change the sidenav to request only the first 10 items 
- [x] add a "more" button to the sidenav that links to /collections
- [x] update the /collections page with a basic paging control

### related issues
- [x] move the "Create a collection" button in the side nav above the list of collections
- [ ] it would be useful to allow sorting of the API results. If DRF makes this easy, we should do it

roberthouse54:
Just to add a note. Elaine has said that one of her main use cases to find videos is that she does an in browser text search to find a video. It is possible that this will be a use case for other users, so we may at some point want to add a basic search feature. As the way she does it now will no longer be possible when we add pagination.

emmello:
It's more to find a collection that I do a search, though in something like the Blossoms collection with a lot of videos, it would be nice to be able to search through pages, but I think that's probably more work than it is worth.

roberthouse54:
Please update the pagination style to the following.

![image](https://user-images.githubusercontent.com/20047260/39821910-501967a2-5377-11e8-9d0c-6716b04dca22.png)

This should be added to both the collection page, and the collections list. Place the pagination at the bottom, aligned right.

![image](https://user-images.githubusercontent.com/20047260/39822160-ea4c5c44-5377-11e8-8693-34a7bc16bdf6.png)

![image](https://user-images.githubusercontent.com/20047260/39822226-0cea5922-5378-11e8-99cd-45e86952b52f.png)

pdpinch:
I'm going to close this. The UI changes in #626 don't seem to be released yet, but I see them on CI.

PR #626, opened 2018-05-09
pagination styling

adorsk:
Partially implements #521 .

#### What's this PR do?
1. Updates paginator styling.
2. Re-positions paginator on `/collections` page.

1. Go to `http://localhost:8089/collections/` and confirm that it matches design guide in [#521](https://github.com/mitodl/odl-video-service/issues/521#issuecomment-387768595).

### DIFF
--- a/static/js/components/Paginator.js
+++ b/static/js/components/Paginator.js
@@ -8,22 +8,17 @@ class Paginator extends React.Component<*, void> {
     const { currentPage, totalPages } = this.props
     return (
       <div className="paginator">
-        <span className="paginator-current-range">
-          Page <span className="paginator-current-page">{currentPage}</span>
-          &nbsp;of&nbsp;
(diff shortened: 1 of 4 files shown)


static/js/components/Paginator.js
```diff
@@ -8,22 +8,17 @@ class Paginator extends React.Component<*, void> {
...
-        </span>
-      
-        <span className="paginator-buttons"
-          style={{
-            marginLeft: '.5em',
-            position:   'relative',
-            top:        '-.1em',
-          }}
-        >
+        <div className="contents" style={{position: "relative"}}>
```
mbertrand: Minor nitpick: maybe `position: relative` could be added to a new style class, and that class could be added here, instead of using `style`?  `<div className="contents contents_relative">`

adorsk: Let's keep the `style` here for now. I'll go ahead and merge.

PR #629, opened 2018-05-11
Paginate videos on collection detail page.

### LINKED ISSUE #503: add paging UI to collection page

adorsk:
Closes #503.

#### What's this PR do?
1. Adds ersatz pagination to videos on collection detail pages. See [my comment here](https://github.com/mitodl/odl-video-service/blob/dorska/ersatz_video_pagination/static/js/components/PaginatedVideoList.js#L3) for why I say 'ersatz'.

1. Adjusts styling class names for paginate video lists.

1. Fixes `Paginator` styling

1. Consolidates pagination type definitions for both video pagination and collections pagination.

1. Make a collection with a bunch of videos. For example, with some shell commands like this:
  ```
import ui.factories
collection = ui.factories.CollectionFactory()
collection.save()
for i in range(100): ui.factories.VideoFactory(collection=collection)
  ```
2. Go to the detail page for the collection. You should see a paginated list that looks reasonable.
3. Create a collection with no videos and view its detail page. It should look reasonable.

### DIFF
--- a/static/js/containers/CollectionDetailPage.js
+++ b/static/js/containers/CollectionDetailPage.js
@@ -8,7 +8,7 @@ import _ from "lodash"
 import DocumentTitle from "react-document-title"
 
 import WithDrawer from "./WithDrawer"
-import VideoCard from "../components/VideoCard"
+import PaginatedVideoList from "../components/PaginatedVideoList"
 import Button from "../components/material/Button"
 import EditVideoFormDialog from "../components/dialogs/EditVideoFormDialog"
 import ShareVideoDialog from "../components/dialogs/ShareVideoDialog"
@@ -21,7 +21,6 @@ import * as ErrorMessages from "../components/errorMessages"
 
 import { actions } from "../actions"
 import * as collectionUiActions from "../actions/collectionUi"
-import { getActiveCollectionDetail } from "../lib/collection"
 import { DIALOGS } from "../constants"
 
 import type { Collection } from "../flow/collectionTypes"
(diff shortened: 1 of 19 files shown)


static/js/containers/CollectionDetailPage.js
```diff
@@ -41,6 +42,13 @@ export class CollectionDetailPage extends React.Component<*, void> {
     showDialog: Function
   }
 
+  VIDEO_LIST_PAGE_SIZE:number
+
+  constructor(props:any) {
+    super(props)
+    this.VIDEO_LIST_PAGE_SIZE = 8
```
mbertrand: Should be based on `settings.PAGE_SIZE_VIDEOS` instead?

adorsk:
@mattbert

[your comment above](https://github.com/mitodl/odl-video-service/pull/629#pullrequestreview-119537142) makes me realize I missed a few issues.

The blank-page/reload-on-first-page issue will happen when:

1. a video is deleted
2. a video is added

Deletion is especially problematic because it would would disrupt the current page. For example, if we delete the last video on a partial page there is currently no logic to load the previous page.

This makes me think that another UI solution might be more appropriate. For example, something like showing the first X videos w/ a 'load more' button that fetches the remaining videos.

In the meanwhile, I think it makes sense to switch back to the normal VideoList in this PR.

That is, replace `PaginatedVideoList` with `VideoList` [here](https://github.com/mitodl/odl-video-service/pull/629/files#diff-f17f7a2c02d970b5f592ba4988b1bb6fL176)

pdpinch:
closing this since we haven't merged it in nearly 7 years

PR #642, opened 2018-05-29
Public video download links

### LINKED ISSUE #639: Download links for public videos
As a Blossoms user, I'd like to be able to download videos using original TechTV URLs.

mbertrand:
- Closes #639 

#### What's this PR do?
- Creates download links for public videos

- If you don't have any imported TechTV collections/videos, create one in a shell for a normal public video that already exists:
  ```python
  from ui.models import Video
  from techtv2ovs.models import *
  my_video = Video.objects.get(<your_video_id>)
  my_video.is_public=True
  my_video.save()
  ttv_collection = TechTVCollection.objects.create(id=100, title='My TTV colletion')
  TechTVVideo.objects.create(id=1, ttv_id=101, ttv_collection=ttv_collection, video=my_video, status='Complete')
   ```
- Go to `/videos/101`, you should get the detail page for the video via the TechTV id (`ttv_id`)
- Go to `/videos/101/download`, `videos/101/download.mp4`, `videos/101/download.avi`, they should all result in the same file being downloaded with the name `<video-title-slug>.<actual_extension_of_file>`
- Go to `/videos/<ovs_video_key>`, you should get the detail page for the video via the video uuid key(`video.hexkey`)
- Go to `/videos/<ovs_video_key>/download`, `videos/<ovs_video_key>/download.mp4`, `videos/<ovs_video_key>/download.avi`, they should all result in the same file being downloaded with the name `<video-title-slug>.<actual_extension_of_file>`
- If you change the video to `is_public=False`, the download links should all return 404's

### DIFF
--- a/ui/models.py
+++ b/ui/models.py
@@ -231,6 +231,20 @@ class Video(TimestampedModel):
                       key=lambda x: EncodingNames.MP4.index(x.encoding) if x.encoding in EncodingNames.MP4 else 0)
 
     @property
+    def download(self):
+        """
+        Return the most appropriate videofile for downloading.
+
+        Returns:
+            VideoFile: The videofile most appropriate for download (highest quality MP4 transcode or original upload)
+        """
+        files = sorted(self.videofile_set.exclude(encoding=EncodingNames.HLS),
+                       key=lambda x: EncodingNames.MP4.index(x.encoding) if x.encoding in EncodingNames.MP4 else 5)
+        if files:
+            return files[0]
+        return None
+
+    @property
     def sources(self):
         """
         Generate a sources dict for VideoJS
--- a/ui/models.py
+++ b/ui/models.py
@@ -238,8 +238,12 @@ class Video(TimestampedModel):
         Returns:
             VideoFile: The videofile most appropriate for download (highest quality MP4 transcode or original upload)
         """
-        files = sorted(self.videofile_set.exclude(encoding=EncodingNames.HLS),
-                       key=lambda x: EncodingNames.MP4.index(x.encoding) if x.encoding in EncodingNames.MP4 else 5)
+        files = sorted(
+            self.videofile_set.exclude(encoding=EncodingNames.HLS),
+            key=lambda x: EncodingNames.MP4.index(x.encoding) if x.encoding in EncodingNames.MP4 else len(
+                EncodingNames.MP4
+            )
+        )
         if files:
             return files[0]
         return None
--- a/ui/urls.py
+++ b/ui/urls.py
@@ -23,9 +23,12 @@ urlpatterns = [
 
     url(r'^videos/(?P<video_key>[0-9a-f]{32})/$', views.VideoDetail.as_view(), name='video-detail'),
     url(r'^videos/(?P<video_key>[0-9a-f]{32})/embed/$', views.VideoEmbed.as_view(), name='video-embed'),
+    url(r'^videos/(?P<video_key>[0-9a-f]{32})/download', views.VideoDownload.as_view(), name='video-download'),
     url(r'^videos/(?P<video_key>\d+)(-.+)?/?$', views.TechTVDetail.as_view(), name='techtv-detail'),
     url(r'^videos/(?P<video_key>.+)/private/?$', views.TechTVPrivateDetail.as_view(), name='techtv-private'),
+    url(r'^videos/(?P<video_key>\d+)(-.+)?/download', views.TechTVDownload.as_view(), name='techtv-download'),
     url(r'^embeds/(?P<video_key>\d+)(-.+)?/?$', views.TechTVEmbed.as_view(), name='techtv-embed'),
+
 
     url(r'^transcode/', include('dj_elastictranscoder.urls')),
     url(r'^api/v0/upload_videos/$', views.UploadVideosFromDropbox.as_view(), name='upload-videos'),
--- a/ui/views.py
+++ b/ui/views.py
@@ -1,11 +1,13 @@
 """Views for ui app"""
 import json
+from os.path import splitext
 
+import requests
 from django.conf import settings
 from django.contrib.auth.views import redirect_to_login
 from django.contrib.auth.views import LoginView as DjangoLoginView
 from django.core.exceptions import PermissionDenied
-from django.http import HttpResponse
+from django.http import HttpResponse, StreamingHttpResponse
 from django.shortcuts import (
     get_object_or_404,
     redirect,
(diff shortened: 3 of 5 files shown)


ui/urls.py
```diff
@@ -23,10 +23,13 @@
 
     url(r'^videos/(?P<video_key>[0-9a-f]{32})/$', views.VideoDetail.as_view(), name='video-detail'),
     url(r'^videos/(?P<video_key>[0-9a-f]{32})/embed/$', views.VideoEmbed.as_view(), name='video-embed'),
+    url(r'^videos/(?P<video_key>[0-9a-f]{32})/download', views.VideoDownload.as_view(), name='video-download'),
```
noisecapella: Can we add a `/$` or `/?$` at the end here and below for consistency?

ui/models.py
```diff
@@ -230,6 +230,20 @@ def transcoded_videos(self):
...
+    @property
+    def download(self):
+        """
+        Return the most appropriate videofile for downloading.
+
+        Returns:
+            VideoFile: The videofile most appropriate for download (highest quality MP4 transcode or original upload)
+        """
+        files = sorted(self.videofile_set.exclude(encoding=EncodingNames.HLS),
+                       key=lambda x: EncodingNames.MP4.index(x.encoding) if x.encoding in EncodingNames.MP4 else 5)
```
noisecapella: Can we replace `5` with `len(EncodingNames.MP4)` if that makes sense here?

ui/models.py
```diff
@@ -230,6 +230,20 @@ def transcoded_videos(self):
...
 
+    @property
+    def download(self):
+        """
+        Return the most appropriate videofile for downloading.
+
+        Returns:
+            VideoFile: The videofile most appropriate for download (highest quality MP4 transcode or original upload)
+        """
+        files = sorted(self.videofile_set.exclude(encoding=EncodingNames.HLS),
```
noisecapella: If this is meant to only include MP4 related encodings should we use `encoding__in=EncodingNames.MP4` here instead?

ui/views.py
```diff
@@ -154,6 +157,32 @@ def get_context_data(self, video, **kwargs):  # pylint: disable=arguments-differ
         return context
 
 
+class VideoDownload(VideoDetail):
+    """
+    Download a public video
+    """
+    def download(self, video):
```
noisecapella: Can you add the `video` argument to the docstring?

ui/views.py
```diff
@@ -154,6 +157,32 @@ def get_context_data(self, video, **kwargs):  # pylint: disable=arguments-differ
...
+        """
+        Return the response as a file download
+        Returns:
+            HttpResponse: the response videofile content and a file attachment header
```
noisecapella: Looks like this should be a `StreamingHttpResponse`

ui/views.py
```diff
@@ -154,6 +157,32 @@ def get_context_data(self, video, **kwargs):  # pylint: disable=arguments-differ
...
+        """
+        video_file = video.download
+        if not video_file:
+            raise Http404()
+        _, ext = splitext(video_file.s3_object_key)
+        video_bytes = requests.get(video_file.cloudfront_url, stream=True)
+        response = StreamingHttpResponse(
+            (chunk for chunk in video_bytes.iter_content(512 * 1024)),
```
noisecapella: Would just passing the `iter_content(...)` call directly work here, instead of wrapping in a generator?

ui/views.py
```diff
@@ -154,6 +157,32 @@ def get_context_data(self, video, **kwargs):  # pylint: disable=arguments-differ
         return context
 
 
+class VideoDownload(VideoDetail):
```
noisecapella: I think we are subclassing `VideoDetail` elsewhere to reuse the `get_context_data` method and template. But here we are producing raw data so I'm not sure it's necessary to subclass that instead of a regular `View`

ui/views.py
```diff
@@ -154,6 +157,32 @@ def get_context_data(self, video, **kwargs):  # pylint: disable=arguments-differ
...
+        _, ext = splitext(video_file.s3_object_key)
+        video_bytes = requests.get(video_file.cloudfront_url, stream=True)
+        response = StreamingHttpResponse(
+            (chunk for chunk in video_bytes.iter_content(512 * 1024)),
+            content_type='video/mp4')
+        response['Content-Disposition'] = 'attachment; filename={}.{}'.format(slugify(video.title), ext)
+        return response
+
+    def get(self, request, *args, **kwargs):
+        video = get_object_or_404(Video, key=kwargs['video_key'], is_public=True)
```
noisecapella: Do we need to do any additional permission check here? It looks like this will only work for cloudfront urls; are all cloudfront videos necessarily public?

ui/views.py
```diff
@@ -182,6 +211,18 @@ def get(self, request, *args, **kwargs):
         return conditional_response(self, ttv_videos[0].video, *args, **kwargs)
 
 
+class TechTVDownload(VideoDownload):
+    """
+    Public video download for a TechTV-based URL
+    """
+    def get(self, request, *args, **kwargs):
+        ttv_videos = get_list_or_404(
```
noisecapella: If there are two items in this list does it matter which one we pick?

ui/urls.py
mbertrand: The URL needs to handle possible variations like `/download.mp4` or `/download.qt`

ui/models.py
mbertrand: sure

ui/models.py
mbertrand: It needs to include `EncodingNames.ORIGINAL` too (not in `MP4`).

ui/views.py
noisecapella: Nevermind, I missed the very obvious `is_public=True` here

mbertrand: This is intended to download from cloudfront URL's; only public videos are allowed to be downloaded via these links.

ui/views.py
mbertrand: No

ui/views.py
mbertrand: ✅

ui/views.py
mbertrand: ✅

noisecapella:
Functionality works great

ui/views.py
mbertrand: Yes, that seems to work

ui/views.py
mbertrand: ✅

ui/views.py
mbertrand: All videos in the list should be identical (same S3 files copied over to multiple OVS `Video` objects).

mbertrand:
@noisecapella ready for review again

PR #679, opened 2018-08-23
Increase moira retrieval limit

### LINKED ISSUE #641: Only first 1000 of a user's moira lists are returned
The function `moira.user_list_membership` should be called with the `max_return_count` keyword argument and a value of something like 10000 or more (default is 1000).

mbertrand:
Fixes #641 

#### What's this PR do?
Sets the ` max_return_count=100000` in the `moira_list_membership` call

In a django shell:

```python
from django.contrib.auth.models import User
from ui.utils import query_moira_lists

user, _ = User.objects.get_or_create(username='<Peter email>', email='<Peter email>')
len(query_moira_lists(user))
```
The length should be about 3354.

### DIFF
--- a/ui/utils.py
+++ b/ui/utils.py
@@ -91,7 +91,7 @@ def query_moira_lists(user):
     moira_user = get_moira_user(user)
     moira = get_moira_client()
     try:
-        list_infos = moira.user_list_membership(moira_user.username, moira_user.type)
+        list_infos = moira.user_list_membership(moira_user.username, moira_user.type, max_return_count=100000)
         list_names = [list_info['listName'] for list_info in list_infos]
         return list_names
     except Exception as exc:  # pylint: disable=broad-except


ui/utils.py
```diff
@@ -91,7 +91,7 @@ def query_moira_lists(user):
     moira_user = get_moira_user(user)
     moira = get_moira_client()
     try:
-        list_infos = moira.user_list_membership(moira_user.username, moira_user.type)
+        list_infos = moira.user_list_membership(moira_user.username, moira_user.type, max_return_count=100000)
```
amir-qayyum-khan: Does this have effect on performance of api?

mbertrand: It could cause a performance hit only for users who belong to many thousands of moira lists.  Without this change, those users may be denied access to collections/videos that they should be able to view.

PR #682, opened 2018-08-23
Formatting of javascript with fmt

mbertrand:
None

#### What's this PR do?
Contains formatting changes in javascript file resulting from running `docker-compose run watch npm run-script`

Nothing should break


aliceriot:
you may want to also go ahead and add `npm run fmt:check` to `travis/js_test.sh`, like we have it set up in MM and OD. that will cause travis to fail if a PR is submitted where someone hasn't yet run `fmt`.

PR #687, opened 2018-08-29
Remove IS_OSX now that Docker for Mac is used by everyone

noisecapella:
See https://github.com/mitodl/cookiecutter-djangoapp/pull/136

#### What's this PR do?
Since Docker for Mac is now used by everyone we don't need to detect OS X to handle it differently

Nothing should break

### DIFF
--- a/scripts/envs.sh
+++ b/scripts/envs.sh
@@ -1,19 +1,6 @@
 #!/usr/bin/env bash
 set -e -o pipefail
 
-# Set IS_OSX to true or false depending on which environment we're running in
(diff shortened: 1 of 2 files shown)


amir-qayyum-khan:
setting up odl videos on docker for mac with this branch for first time.....

amir-qayyum-khan:
getting issue ```[2018-09-04 14:44:30,776: ERROR/MainProcess] consumer: Cannot connect to redis://redis:6379/4: Error -2 connecting to redis:6379. Name or service not known..```

PR #688, opened 2018-08-29
Configure raven.js

noisecapella:
See also https://github.com/mitodl/open-discussions/pull/1142

#### What's this PR do?
Configures raven.js so Sentry will catch front end errors

In the PR build, go to the homepage and open up the browser console. Type:

    var body = document.getElementsByTagName("body")
    var script = document.createElement("script")
    script.innerHTML = "throw new Error('something unique to make it obviously a test')"
    body.append(script)

You should see an exception in the console. If you open up the stack trace of the exception there should be some mention of raven (not sure if this is true for all browsers). Go to Sentry and verify that it caught an error with the error text from a CI environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "prettier-eslint-cli": "^4.7.0",
     "prop-types": "^15.5.10",
     "ramda": "^0.24.1",
+    "raven-js": "^3.26.4",
     "react": "^15.6.1",
     "react-addons-shallow-compare": "^15.6.0",
     "react-document-title": "^2.0.3",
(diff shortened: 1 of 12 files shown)


aliceriot:
it works!

mysentryerror

aliceriot:
haha, that screenshot should have been for the discussions pr. is there a PR build set up for odl-video?

noisecapella:
Ah right, I'm not sure PR builds are set up for odl-video. I think we'll just have to merge it and test it on CI

PR #709, opened 2018-10-02
Enable anonymous access to collection pages

### LINKED ISSUE #706: anonymous access to collection pages
As a collection admin who has made their videos public, I'd like users to be able to see a list of the videos without logging in either. 

Currently, collection pages require an authenticated user. We can remove this restriction for all collections, whether they contain public or private videos. 

There is limited risk in exposing the thumbnails when the video detail pages will still be access controls. However, it would be a better UX if anonymous access to the collection API only returned the public videos.   

### acceptance criteria
- [ ] remove the requirement for an authenticated user on the /collections/[collection_id] route. This may also require an nginx change in production to remove the Touchstone requirement
- [ ] hide any UI elements that require an authenticated user, like upload and settings. 
- [ ] allow anonymous access to the collections API so the left nav will work, and video thumbnails will display (ideally, only public video thumbnails)
- [ ] test (on RC?) that Touchstone login still functions for private videos

annagav:
Fix #706 

#### What's this PR do?
Allow anonymous user to view a collection page.

Create a collection with public and private videos. Try to view the collection detail page from an incognito window. You should be able to see the public videos.

### DIFF
--- a/static/js/containers/CollectionDetailPage.js
+++ b/static/js/containers/CollectionDetailPage.js
@@ -71,7 +71,11 @@ export class CollectionDetailPage extends React.Component<*, void> {
       <DocumentTitle title={collection ? `OVS | ${collection.title}` : "OVS"}>
         <WithDrawer>
           <VideoSaverScript />
-          <div className="collection-detail-content">{this.renderBody()}</div>
+          <div className="collection-detail-content">
+            {collectionError
+              ? this.renderError(collectionError)
+              : this.renderBody()}
+          </div>
         </WithDrawer>
       </DocumentTitle>
     )
--- a/static/js/containers/CollectionDetailPage.js
+++ b/static/js/containers/CollectionDetailPage.js
@@ -71,11 +71,7 @@ export class CollectionDetailPage extends React.Component<*, void> {
       <DocumentTitle title={collection ? `OVS | ${collection.title}` : "OVS"}>
         <WithDrawer>
           <VideoSaverScript />
-          <div className="collection-detail-content">
-            {collectionError
-              ? this.renderError(collectionError)
-              : this.renderBody()}
-          </div>
+          <div className="collection-detail-content">{this.renderBody()}</div>
         </WithDrawer>
       </DocumentTitle>
     )
--- a/ui/permissions_test.py
+++ b/ui/permissions_test.py
@@ -191,24 +191,11 @@ def test_is_collection_owner_user_superuser(collection_owner_permission, collect
 def test_is_collection_admin_anonymous_users(collection_permission, collection, request_data_anon):
     """
     Test for HasCollectionPermissions.has_object_permission to verify that
-    anonymous users do not have permission.
+    anonymous users have permission.
     """
     assert collection_permission.has_object_permission(request_data_anon.request,
                                                        request_data_anon.view,
-                                                       collection) is False
-
-
-def test_is_collection_admin_user_different_from_collecton_owner(mock_user_moira_lists,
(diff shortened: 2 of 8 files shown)


ui/permissions_test.py
```diff
@@ -191,24 +191,11 @@ def test_is_collection_owner_user_superuser(collection_owner_permission, collect
 def test_is_collection_admin_anonymous_users(collection_permission, collection, request_data_anon):
     """
     Test for HasCollectionPermissions.has_object_permission to verify that
-    anonymous users do not have permission.
+    anonymous users have permission.
```
amir-qayyum-khan: it would be nice to add test where user change permission from private to public and assert that user other user can view

static/js/containers/CollectionDetailPage.js
```diff
@@ -71,11 +71,7 @@ export class CollectionDetailPage extends React.Component<*, void> {
       <DocumentTitle title={collection ? `OVS | ${collection.title}` : "OVS"}>
         <WithDrawer>
           <VideoSaverScript />
-          <div className="collection-detail-content">
-            {collectionError
-              ? this.renderError(collectionError)
```
amir-qayyum-khan: why we remove `renderError`. How we are going to handle errors now/

amir-qayyum-khan:
please add test to demonstrate that if collection is change from private to public. it is visible to anonymous user.

mbertrand:
As an anonymous user, I can see all video thumbnails in the collection even if some of them are not public.  One possible way to fix this is to adjust `ui.serializers.CollectionSerializer`:

```python
    videos = serializers.SerializerMethodField()
.....
    def get_videos(self, obj):
        if self.context.get('request') and self.context.get('request').user.is_anonymous:
            videos = obj.videos.filter(is_public=True)
        else:
            videos = obj.videos.all()
        return [SimpleVideoSerializer(video).data for video in videos]
```

ui/views.py
```diff
@@ -115,7 +137,10 @@ def get_context_data(self, **kwargs):  # pylint: disable=arguments-differ
     def get(self, request, *args, **kwargs):
         collection_key = kwargs['collection_key']
         if collection_key:
-            get_object_or_404(Collection, key=collection_key)
+            collection = get_object_or_404(Collection, key=collection_key)
```
mbertrand: I think this can be simplified to:
```python
if collection_key:
    get_object_or_404(Collection, key=collection_key)
    return super().get(request, *args, **kwargs)
```

ui/views_test.py
```diff
@@ -269,7 +269,7 @@ def test_collection_viewset_permissions(logged_in_apiclient):
         reverse('models-api:collection-detail', kwargs={'key': uuid4().hex}),
     )
     for url in urls:
-        assert client.get(url).status_code == status.HTTP_403_FORBIDDEN
+        assert client.get(url).status_code == status.HTTP_200_OK
```
mbertrand: This test fails with a 404 because it is looking for a collection that doesn't exist, try this:

```python
def test_collection_viewset_permissions(logged_in_apiclient):
    """
    Tests the list of collections for an anonymous_user
    """
    client.logout()
    collection = CollectionFactory()
    urls = (
        reverse('models-api:collection-list'),
        reverse('models-api:collection-detail', kwargs={'key': collection.hexkey}),
    )
    for url in urls:
        assert client.get(url).status_code == status.HTTP_200_OK
        assert client.post(url, {'owner': 1}).status_code == status.HTTP_403_FORBIDDEN
```

PR #719, opened 2018-10-22
Change no-videos message for anonymous

### LINKED ISSUE #718: Change empty collection/video message for anonymous user
screen shot 2018-10-19 at 2 34 18 pm

annagav:
Fix #718

#### What's this PR do?
If an anonymous user tries to view a collection's detail page and there are no public videos the message should be different: 
screen shot 2018-10-22 at 9 54 05 am

screen shot 2018-10-22 at 10 17 05 am

Try to access a collections page with no public videos.

### DIFF
--- a/static/js/containers/CollectionDetailPage.js
+++ b/static/js/containers/CollectionDetailPage.js
@@ -195,11 +195,12 @@ export class CollectionDetailPage extends React.Component<*, void> {
   renderEmptyVideoMessage(isAdmin: boolean) {
     let message = "There are no public videos available for viewing."
     if (isAdmin) {
-      message = "There are no videos yet. Click the button above to add videos from a linked Dropbox account."
+      message =
+        "There are no videos yet. Click the button above to add videos from a linked Dropbox account."
     }
     return (
       <div className="no-videos">
-        <p>{ message }</p>
+        <p>{message}</p>
       </div>
     )
   }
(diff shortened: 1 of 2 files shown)


static/js/containers/CollectionDetailPage.js
```diff
@@ -192,13 +192,15 @@ export class CollectionDetailPage extends React.Component<*, void> {
     )
   }
 
-  renderEmptyVideoMessage() {
+  renderEmptyVideoMessage(isAdmin: boolean) {
+    let message = "There are no public videos available for viewing."
```
pdpinch: We've confused some users who don't realize that they need to login. Can we add something like this (not sure if this will work with react) 
```suggestion
    let message = "There are no public videos available for viewing. Please <a href="/login">login</a> to view private videos"
```

Ideally that login link would direct them back to the current page after logging in.

annagav:
@amir-qayyum-khan, have you tried restarting your container?

amir-qayyum-khan:
it is working now, thanks @annagav. I rebased. Just follow @pdpinch comment and add some tests

annagav:
@amir-qayyum-khan, I rebased the branch.

static/js/containers/CollectionDetailPage.js
mbertrand: Might be useful to make the URL `/login/?next=/collections/<collection_key>` so the user is redirected back to the collection after logging in.

PR #722, opened 2018-11-05
Upgrade requests

shaidar:
#### What's this PR do?
Upgrade `requests` based on potential security vulnerability.


pdpinch:
Don't we need to update requirements.txt? Did you run pip-compile?

Issue #740, opened 2019-03-14
embedded videos in Residential MITx from video.odl.mit.edu not playing


pdpinch:
from [ZenDesk #26917](https://odl.zendesk.com/agent/tickets/26917)

> Students in 6.036 are reporting spinning play wheels, when trying to play videos embedded with an iframe from video.odl.mit.edu ; this error appears in the js console:
> 
```
VM606 www-widgetapi.js:100 Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://www.youtube.com') does not match the recipient window's origin ('https://video.odl.mit.edu').
```
> 
> for example, due to this embed:
> 
```
<iframe src="https://video.odl.mit.edu/videos/0a4d3a43a8524616a2e027852808aed0/embed/" width="560" height="315" frameborder="0" allowfullscreen="1"></iframe>
```
> 
> Example which reproduces the error:
>
> https://lms.mitx.mit.edu/courses/course-v1:MITx+6.036+2019_Spring/courseware/Week6/week6_video/1
>
> Playing directly at the video.odl.mit.edu site still works, as a backup.

mbertrand:
This issue is described in the VideoJS-Youtube plugin [here](https://github.com/videojs/videojs-youtube/issues/532), and a potential solution is on [StackOverflow](https://stackoverflow.com/questions/54786686/videojs-youtube-video-not-playing-in-an-iframe-in-desktop-chrome).

PR #741, opened 2019-03-15
fix embedded video not playing

tasawernawaz:
#740 

Not sure, how can test this. At least tests should pass.

### DIFF
--- a/static/js/components/dialogs/ShareVideoDialog.js
+++ b/static/js/components/dialogs/ShareVideoDialog.js
@@ -61,7 +61,7 @@ class ShareVideoDialog extends React.Component<*, void> {
             label="Embed HTML"
             id="video-embed-code"
             rows="4"
-            value={`<iframe src="${videoEmbedUrl}" width="560" height="315" frameborder="0" allowfullscreen></iframe>`}
+            value={`<iframe src="${videoEmbedUrl}" width="560" height="315" frameborder="0" allow="autoplay" allowfullscreen></iframe>`}
           />
           <Checkbox
             label={`Start at ${formatSecondsToMinutes(startTime)}`}


pdpinch:
I don't want to slow down this PR at all, but I'm curious if there is something we should do to get this type of error to show up in Sentry. 

I see errors like https://sentry.io/share/issue/8ebe19ea8e04425584f6b6ef76a5ed79/ and https://sentry.io/share/issue/e964694727e9435490feb03b53447468/ but nothing related to Residential MITx

tasawernawaz:
@pdpinch May I should have asked this question sometime earlier, but I was confuse what actually `Residential MITx` refers to?

mbertrand:
@pdpinch @tasawernawaz  The error `VM606 www-widgetapi.js:100 Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://www.youtube.com') does not match the recipient window's origin ('https://video.odl.mit.edu').` is not responsible for the lack of playability, it occurs even when the video does play on another browser and/or when `autoplay=true` is added to the iframe, making it playable in Chrome.  It does not seem to affect usability.

Raven.js is showing up in the stack trace, not sure why it isn't showing up in sentry but my guess is because the error is originating from the Youtube script, there is a plugin issue about it here: https://github.com/videojs/videojs-youtube/issues/67

pdpinch:
Ok, if we can't get it into Sentry, that's OK. If this fixes the issue, let's merge it and move on. 

In terms of the fix -- will folks who have the current embed code need to update it?

tasawernawaz:
@pdpinch  This will surely fix for the new embedded code, @mbertrand can you answer for the existing issue ?

mbertrand:
@pdpinch @tasawernawaz Yes I think anyone using the old iframe code will need to add `allow="autoplay"` so that it will be playable in Chrome again.

PR #743, opened 2019-03-21
views added for moira list and users

### LINKED ISSUE #739: Add REST API views for checking moira list membership
Add views for determining:

- [ ] The members of a moira list (`/api/v0/moira/list/<list_name>`)
- [ ] The moira lists a user belongs to (`/api/v0/moira/user/<user_name or email>`)

These views should only be accessible to admins.

tasawernawaz:
fixes #739 

#### What's this PR do?
Add following views
- The members of a moira list `/api/v0/moira/list/<list_name>`
- The moira lists a user belongs to `/api/v0/moira/user/<user_name or email>`

Test above views on your local.

### DIFF
--- a/ui/urls.py
+++ b/ui/urls.py
@@ -37,4 +37,8 @@ urlpatterns = [
     url(r'^api/v0/upload_subtitles/$', views.UploadVideoSubtitle.as_view(),
         name='upload-subtitles'),
     url(r'^api/v0/', include((router.urls, 'models-api'))),
+
+    url(r'^api/v0/moira/user/(?P<username_or_email>[-\w]+$|[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+)$',
+        views.MoiraListsForUser.as_view(), name='member-lists'),
+    url(r'^api/v0/moira/list/(?P<list_name>[-\w]+)$', views.UsersForMoiraList.as_view(), name='list-members'),
 ]
--- a/ui/urls.py
+++ b/ui/urls.py
@@ -38,7 +38,7 @@ urlpatterns = [
         name='upload-subtitles'),
     url(r'^api/v0/', include((router.urls, 'models-api'))),
 
-    url(r'^api/v0/moira/user/(?P<username_or_email>[-\w]+$|[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+)$',
+    url(r'^api/v0/moira/user/(?P<username_or_email>[-\w]+$|.*@.*)$',
         views.MoiraListsForUser.as_view(), name='member-lists'),
     url(r'^api/v0/moira/list/(?P<list_name>[-\w]+)$', views.UsersForMoiraList.as_view(), name='list-members'),
 ]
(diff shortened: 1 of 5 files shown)


ui/views_test.py
```diff
@@ -1230,3 +1230,67 @@ def test_terms_of_service_for_logged_in_user(mock_user_moira_lists, owns_collect
...
+    [
+        reverse('member-lists', kwargs={'username_or_email': 'test_user'}),
+        reverse('list-members', kwargs={'list_name': 'test_name'}),
+    ]
+)
+def test_moira_list_views_permission(logged_in_apiclient, url):
+    """
+    Tests that only authenticated users with admin permissions can call MoiraListsForUser and UsersForMoiraList
+    """
+    client, user = logged_in_apiclient
```
mbertrand: I thinl you'll need to add this for test to pass:
```python
mocker.patch('ui.views.list_members', return_value=[])
```

ui/views.py
```diff
@@ -494,3 +497,40 @@ def get(self, request, *args, **kwargs):
...
+    """
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAdminUser,)
+
+    def get(self, request, username_or_email):
+        """Get and return the list names"""
+        regex = r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)"
+
+        email = username_or_email
+        if not username_or_email.endswith("@mit.edu"):
```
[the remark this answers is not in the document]
tasawernawaz: This will break when supplied argument is a valid email address but not ends with `@mit.edu`, e.g `tasawernawaz@gmail.com`

ui/urls.py
```diff
@@ -37,4 +37,8 @@
     url(r'^api/v0/upload_subtitles/$', views.UploadVideoSubtitle.as_view(),
         name='upload-subtitles'),
     url(r'^api/v0/', include((router.urls, 'models-api'))),
+
+    url(r'^api/v0/moira/user/(?P<username_or_email>[-\w]+$|[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+)$',
```
mbertrand: Are you sure this regex will cover all valid emails?  For instance I think "&" is a valid character in an email. Might be safer to use a less stringent regex here.

ui/views.py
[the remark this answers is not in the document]
mbertrand: This seems to work with all emails:
```python
    def get(self, request, username_or_email):
        """Get and return the list names"""

        email = username_or_email
        if '@' not in email:
            email = "{username}@mit.edu".format(username=username_or_email)

        try:
            user = User.objects.get(Q(username=username_or_email) | Q(email=email))
        except User.DoesNotExist:
            return Response(status.HTTP_404_NOT_FOUND)

        user_lists = query_moira_lists(user)
        return Response(data={'user_lists': user_lists})
```

but you can keep it this way if you prefer.  However in that case it might be best to make use of django's built in email validator if you stick with this approach, rather than a custom regex (`from django.core.validators import validate_email`)

pdpinch:
Yes please. If we must have validation, don’t write your own code. 

Google a bit and you’ll see how hard it is to write a regex to validate emails.

ui/urls.py
tasawernawaz: I have updated this to a basic regex.

tasawernawaz:
@pdpinch @mbertrand I have updated email validation and URL pattern according to your suggestions.

PR #763, opened 2019-08-01
Added request to auto-add HLS videos to edX when appropriate

### LINKED ISSUE #762: Automatically add HLS videos to edX via API
**Depends on https://github.com/mitodl/edx-platform/issues/139**

We want to be able to take OVS videos that have already been uploaded/encoded and add them to edX, where a course author can insert them into course content via Studio

Acceptance criteria:
- [ ] edX-related app settings added (edX OVS user access token, edX API key, base edX URL)
- [ ] Optional edX course key field added to video collections model 
- [ ] Backend method created to add an HLS video to edX* via the new endpoint added as part of https://github.com/mitodl/edx-platform/issues/139
- [ ] HLS videos automatically posted to edX if the collection has a course key and transcoding/etc succeeds

\* - at least to begin with, we can probably just construct this HTTP request "manually" instead of adding this functionality to the edX API client and adding that library as a requirement to OVS

gsidebo:
Closes #762 

#### What's this PR do?
When a video is added to a collection and successfully transcoded, this PR adds an API request to add the HLS-encoded video to edX (triggered automatically when the model object is created)

- Run devstack with the `edx-video-api` package installed (details in the [project README](https://github.com/mitodl/edx-video-api))
- Add settings to your `.env`:
    ```
    EDX_API_KEY=PUT_YOUR_API_KEY_HERE
    EDX_ACCESS_TOKEN=<access token value for some admin user in edX>
    # OSX users
    EDX_BASE_URL=http://host.docker.internal:18000
    # Linux users
    EDX_BASE_URL=http://localhost:18000
    ```
- Set the `edx_course_id` value for some collection to a course id in your LMS instance (e.g.: `course-v1:edX+DemoX+Demo_Course`). This can be done in Django admin
- Add a video to that collection

When the video is done uploading and transcoding, a successful request should be made to add the HLS-encoded video to edX. You should see that request in the celery logs.

I added the necessary config files to allow devs to run a container-ized Jupyter notebook. I added it so I could do some of my own data manipulation and it made enough sense just to leave those files in the PR


pdpinch:
@mbertrand can you review this?

Can I also get your opinion on the authentication/authorization scheme? I had told Gavin that we should use a single privileged user, but I'm having some second thoughts about it. But we don't have any python-socialauth infrastructure in this project, do we?

mbertrand:
@pdpinch that's correct, no python-socialauth in OVS, just Touchstone.  The authentication via EdX API key seems ok to me at first glance, I'll let you know if I change my mind after testing this out a bit.

ui/utils.py
```diff
@@ -378,3 +379,33 @@ def list_members(list_name):
...
+    return urljoin(url_base, url_path)
+
+
+def edx_settings_configured():
+    """Returns True if all the necessary settings for adding a video to edX via API are properly configured"""
+    return all((
+        settings.EDX_BASE_URL,
+        settings.EDX_HLS_API_URL,
+        settings.EDX_ACCESS_TOKEN,
+        settings.EDX_API_KEY
```
mbertrand: These should all be added to `app.json`

odl_video/settings.py
```diff
@@ -539,3 +539,8 @@ def get_all_config_keys():
 MOIRA_CACHE_TIMEOUT = get_int('MOIRA_CACHE_TIMEOUT', 10800)
 HIJACK_ALLOW_GET_REQUESTS = True
 HIJACK_LOGOUT_REDIRECT_URL = '/admin/auth/user'
+
+EDX_ACCESS_TOKEN = get_string('EDX_ACCESS_TOKEN', None)
```
mbertrand: Seems like the tokens generated automatically in edx have short lifespans, but the expiration can be manually changed.  I guess this is okay as long as we pick a token with a far-future expiration date?

ui/api.py
```diff
@@ -53,3 +54,42 @@ def process_dropbox_data(dropbox_upload_data):
...
+    Posts an HLS video to edX via API using attributes from a video file
+
+    Args:
+        video_file (ui.models.VideoFile): An HLS-encoded video file
+
+    Returns:
+        requests.models.Response: The API response
+    """
+    assert edx_settings_configured(), "edX settings need to be configured"
+    assert video_file.can_add_to_edx(), "This video file is not of the correct type"
```
mbertrand: remove asserts?

ui/models.py
```diff
@@ -16,12 +17,29 @@
...
+        as_filter(bool): Filter the bucket by the key
+    """
+    bucket = get_bucket(bucket_name)
+    if not as_filter:
+        bucket.delete_objects(Delete={'Objects': [{'Key': key}]})
+    else:
+        for obj in bucket.objects.filter(Prefix=key):
+            obj.delete()
+
+
```
mbertrand: Seems like `tasks.py` is a better place for this than `models.py`?

ui/tasks.py
```diff
@@ -1,24 +1,21 @@
...
-        bucket.delete_objects(Delete={'Objects': [{'Key': key}]})
-    else:
-        for obj in bucket.objects.filter(Prefix=key):
-            obj.delete()
+@app.task
+def post_hls_to_edx(video_file_id):
+    """Loads a VideoFile and calls our API method to add it to edX"""
+    video_file = VideoFile.objects.filter(id=video_file_id).select_related("video__collection").first()
+    if not video_file:
+        log.error("VideoFile with id %s doesn't exist", video_file_id)
```
mbertrand: This condition could use a unit test

odl_video/settings.py
gsidebo: Yeah, that's what I was thinking. If a token expires, it will be easy to manually update this setting

ui/models.py
gsidebo: It's a task, but it's only used by these models, and it's specifically designed _for_ models. Having this in `tasks.py` also created a circular import (tasks -> api -> models -> tasks). I felt it was cleaner/more intuitive to move this into `models.py` than to inline import both the api and models in the `post_hls_to_edx` task.

mbertrand: 👍

ui/api.py
gsidebo: I just put this here as a bit of extra assurance. This function can be called from anywhere, and I wanted to make it explicit that it should only be used under these conditions

mbertrand: 👍

gsidebo:
@mbertrand Ready again

PR #767, opened 2019-08-05
Changed edX auto-add to use edxval library endpoints

### LINKED ISSUE #765: Make edX requests via edxval endpoints instead of new library
PR #763 (issue #762) added a feature where an HLS video could be automatically added to edX via HTTP API after adding it to a collection. That edX API was added by a new plugin app, and unfortunately we recently discovered that an API was already exposed by the original video library and that a plugin app was not needed. This issue covers the work to use those API endpoints from the original library (`edx-val`) instead of the API endpoints that were added by the new library (`edx-video-api`)

gsidebo:
Closes #765 

#### What's this PR do?
Changed edX auto-add to use `edx-val` library endpoints instead of using endpoints exposed by a new library

- Run devstack (the edX instance needs to have `edx-val` installed, and the `xpro/ironwood` branch already lists it as a requirement, so it should be ready to go)
- Add an access token for some edX admin user (Django admin path: `/admin/oauth2/accesstoken/`). **THINGS TO NOTE:** (1) edX has 2 different oauth libraries installed. This access token must be an object of the model represented by that Django admin path, and not the other access token model. (2) Feel free to create a new Client for this. The URLs don't matter.
- Add settings to your `.env`:
    ```
    EDX_ACCESS_TOKEN=<access token value for some admin user in edX>
    # OSX users
    EDX_BASE_URL=http://host.docker.internal:18000
    # Linux users
    EDX_BASE_URL=http://localhost:18000
    ```
- Set the `edx_course_id` value for some collection to a course id in your LMS instance (e.g.: `course-v1:edX+DemoX+Demo_Course`). This can be done in Django admin
- Add a video to that collection, wait it to transcode and for celery to process it once it finishes. If it's successful, you should see a celery log line for the successful request to, e.g.: `celery_1  | [2019-08-05 21:07:13,135: INFO/ForkPoolWorker-2] Task ui.tasks.post_hls_to_edx[4d4c0327-e272-4470-a984-9af23959f74f] succeeded in 0.1495262999960687s: 201` (the `201` is the response code)

### DIFF
--- a/app.json
+++ b/app.json
@@ -33,16 +33,12 @@
       "description": "Access token value for an edX admin user",
       "required": false
     },
-    "EDX_API_KEY": {
-      "description": "edX API key",
-      "required": false
-    },
     "EDX_BASE_URL": {
       "description": "Base URL for the edX API",
       "required": false
     },
     "EDX_HLS_API_URL": {
-      "default": "/api/course_videos/",
+      "default": "/api/val/v0/videos/",
       "description": "URL path for the HLS video endpoint on edX",
       "required": false
     },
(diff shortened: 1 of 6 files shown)


annagav:
My devstack is breaking, so I could not test the functionality, maybe someone has a working one on `xpro/ironwood`. But the code looks good!

PR #784, opened 2019-10-24
add mp4 and webm to transcode settings

### LINKED ISSUE #785: xPRO: add transcode settings for MP4, WebM
As a user of open edX as part of the xPRO service, I'd like to be able to download mp4 versions of videos when I can't play back HLS.

pdpinch:
We have heard from some open edX users having trouble with HLS video. Perhaps these options will help?

fixes #785 

#### What's this PR do?

adds two new codecs to our elastic transcode settings. 

There will need to be a subsequent PR to push these encodings to open edX Studio. 

I don't know how/if these encodings will be an option in the OVS player. 

- Is it valid JSON?
- Does it break AWS? I'm not sure the settings are all valid options

So many ZenDesk tickets about video playback problems in xPRO.

### DIFF
--- a/config/et_presets.json
+++ b/config/et_presets.json
@@ -162,5 +162,87 @@
       "PaddingPolicy": "NoPad",
       "SizingPolicy": "Fit"
     }
+  },
+  {
+    "Name": "MP4_AV_1_0Mbps",
+    "Description": "MP4 for low network bandwidth",
+    "Container": "mp4",
+    "Video": {
+      "BitRate": "900",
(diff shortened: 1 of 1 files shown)


pdpinch:
@mbertrand I'd like to try these setting to see if it helps us with some support issues with xPRO. 

1. Do you think this will work?
2. What would the impact be (if any) on regular OVS functionality?

pdpinch:
I don't think our problem is missing mp4 encodes. Our problem is not enough low bitrates.

Issue #788, opened 2019-11-02
batch re-encode HLS video


pdpinch:
As an admin, I'd like an easy, scalable way to re-encode videos for some collections, especially when we add or change the encoding presets (see #787). 

#### Acceptance criteria

- [ ] option for re-encoding just one video, by ID
- [ ] option for re-encoding all videos in a collection, by collection ID (or collection name?)
- [ ] after the video is re-encoded, need a way to discard the old encodings

#### out of scope

- I think it's best not to have an option to re-encode all the videos on the site, since that is a expensive undertaking

mbertrand:
Do you want a management command?  devops would need to run it since RC and PR are not on heroku.  

Or I could add a boolean 'schedule_encode' field on both Video and Collection objects, and set up a recurring celery task to reencode any of those objects with `schedule_encode=True` (and then set that to False again).

`after the video is re-encoded, need a way to discard the old encodings` - I think we are just adding additional encodings?  The m3u8 file for a video will regenerated too and any unused encodings won't be in it.

pdpinch:
I like the idea of adding the "schedule_encode" field. Can we add that as a filter in the admin so there's a way to find collections and videos that are in one state or the other?

In the interests of time, we may only be adding new encoding now, but I'd like to have an option to replace existing encodings with entirely new ones.

mbertrand:
@pdpinch Unfortunately there's a complication: Elastic Transcoder will not overwrite existing files, so a retranscode will always fail (`The specified object could not be saved in the specified bucket because an object by that name already exists`).  Some possible workarounds:

- Rewrite the code to specify temporary s3 bucket files for ET to write to, and when finished copy to the correct S3 location and delete the temporary ones.
- Remove the original transcode files before retranscoding.  This is obviously not ideal because then the video will be unplayable until the retranscode is complete.

mbertrand:
Here is the proposed approach:

- Create a flag on Collection and Video: boolean field ‘schedule_reencode’
- Create a recurring task to look for videos and collections with schedule_reencode=True.
  - check flagged collections first:
     - set flag to True for every video in the collection
     - reset collection flag to False
  - Trigger a new encode job for each flagged video.
  - Assign new video status - ‘Re-encoding’ 
  - Reset flag to False
- Modify the existing encoding job metadata:
  - a temporary prefix should be added to the output filenames to avoid transcode errors that occur when trying to overwrite existing output from the previous transcode
[...]

PR #791, opened 2019-11-06
RFC for retranscodes

mbertrand:
[Rendered](https://github.com/mitodl/odl-video-service/blob/mb/rfcretranscode/docs/rfcs/0001-video-retranscoding.md)

#### Relevant Issues
#788 

#### Summary
Sugfested approach for supporting the retranscoding of existing videos

### DIFF
--- a/docs/rfcs/0001-video-retranscoding.md
+++ b/docs/rfcs/0001-video-retranscoding.md
@@ -1,0 +1,43 @@
+## Video Re-transcoding
+
+#### Abstract
+Provide a way to trigger re-transcoding of specified videos, or all videos in a collection, if new presets are needed.
+
+
+#### Architecture Changes
+
+- Create a flag on Collection and Video: boolean field ‘schedule_retranscode’
+- Create a recurring task to look for videos and collections with `schedule_retranscode=True`.
+  - check flagged collections first:
+     - set `schedule_retranscode=True` for every Video in the Collection, then
+     - set `schedule_retranscode=False` on the Collection
+  - Trigger a new encode job for each flagged Video.
+  - Assign a new video status - ‘Retranscoding’
+  - Set `schedule_retranscode=False` on the Video
+- Modify the generation of encoding job metadata:
+  - a temporary prefix should be added to the output filenames to avoid transcode errors that occur when trying to overwrite existing output from the previous transcode
+- Videos being re-transcoded should still be playable (front end needs to accept the new reencoding status as playable)
+- Modify the task that checks on video transcode statuses
+  - should handle the new Retranscoding status
+  - There will now be 2+ encode jobs for the video instead of just 1, the most recent is the one that needs to be checked for progress.
+- When a retranscoding is complete:
+  - copy output from the temporary S3 subfolder to the correct subfolder, then delete the temporary files
+  - Change video status back to complete
+- If a retranscoding task fails:
+  - Log the error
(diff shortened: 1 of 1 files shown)


docs/rfcs/0001-video-retranscoding.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### Architecture Changes
+
+- Create a flag on Collection and Video: boolean field ‘schedule_retranscode’
+- Create a recurring task to look for videos and collections with `schedule_retranscode=True`.
+  - check flagged collections first:
+     - set `schedule_retranscode=True` for every Video in the Collection, then
+     - set `schedule_retranscode=False` on the Collection
+  - Trigger a new encode job for each flagged Video.
+  - Assign a new video status - ‘Retranscoding’
```
pdpinch: Is "video status" an existing field on the video model?

mbertrand: Yes, this is just adding a new possible value for it.

mbertrand: adding a new failure status too - 'Retranscode failed'

docs/rfcs/0001-video-retranscoding.md
```diff
@@ -0,0 +1,43 @@
...
+  - Set `schedule_retranscode=False` on the Video
+- Modify the generation of encoding job metadata:
+  - a temporary prefix should be added to the output filenames to avoid transcode errors that occur when trying to overwrite existing output from the previous transcode
+- Videos being re-transcoded should still be playable (front end needs to accept the new reencoding status as playable)
+- Modify the task that checks on video transcode statuses
+  - should handle the new Retranscoding status
+  - There will now be 2+ encode jobs for the video instead of just 1, the most recent is the one that needs to be checked for progress.
+- When a retranscoding is complete:
+  - copy output from the temporary S3 subfolder to the correct subfolder, then delete the temporary files
+  - Change video status back to complete
```
pdpinch: @gsidebo I think we also need to update edX when the retranscode is complete, since I believe the HLS cloudfront URL will change. Does it make sense to add that to this process as well?

docs/rfcs/0001-video-retranscoding.md
```diff
@@ -0,0 +1,43 @@
...
+- If a retranscoding task fails:
+  - Log the error
+  - Set video status back to complete (original transcode should still work)
+
+#### Security/Other Considerations
+
+If for some reason the copying of retranscoded S3 files is interrupted or fails partway, the video may be rendered unplayable, since some of the original transcode files will be overwritten.
+
```
pdpinch: Can we turn on versioning for this bucket? I know it's expensive, but it's an easy way to get. backup.

pdpinch: Not sure where logging belongs, but I'd like to have a way to follow along with the retrancoding status

mbertrand: >Can we turn on versioning for this bucket?

@shaidar @blarghmatey that bucket would be the value of `VIDEO_S3_TRANSCODE_BUCKET`

mbertrand: @pdpinch another option is to always rename the transcoded output files instead of overwriting the original ones, but that would mean updating the video links in xpro too.

PR #792, opened 2019-11-07
Support for retranscoding videos

### LINKED ISSUE #788: batch re-encode HLS video

mbertrand:
Closes #788 

#### What's this PR do?
- Adds a new feature flag: `FEATURE_RETRANSCODE_ENABLED`
- Adds `schedule_retranscode` field to `Video` and `Collection`
- Adds a new recurring task (if flag is enabled) that will create new encode jobs for videos with `schedule_retranscode=True`
- Modifies existing transcode functions to properly process videos being retranscoded again instead of for the first time.

- Get required .env values like AWS keys, buckets, etc from heroku CI
- set `FEATURE_RETRANSCODE_ENABLED=True` in your .env file
- Upload 1 or more new short videos, it should work as before - click the resolution selector on the video player and take note of the available options.
- Shut down docker containers, add these additonal presets to `ET_PRESET_IDS` in your .env file: `1351620000001-200050,1351620000001-200040`, then restart containers.
- Go to django admin and set `schedule_retranscode=True` on one collection, and on one video in a different collection.  IMPORTANT: choose collections/videos you haven't watched in at least 24 hours to avoid cloudfront caching.
- If you check the videos in the scheduled collection, all their `schedule_retranscode` should now be `True` too.
- Either wait for the `schedule_retranscodes` task to run on its own, or start it manually.  Videos with `schedule_retranscode=True` should be set back to `False`, their status changed to `Retranscoding`, and a new EncodeJob should be visible on their admin detail page.
- Wait for `update_video_statuses` task to run (might need to run a few times depending on how long the transcode takes).  Eventually the video statuses should change from 'Retranscoding' to 'Complete'
- Go to one of the videos that has been retranscoded.  Click the resolution selector on the video player and you should see 2 more options than before.


annagav:
@mbertrand After the retranscode it is showing this:
Screen Shot 2019-11-13 at 9 40 55 AM

mbertrand:
@annagav the new transcodes should be added to the existing list of `ET_PRESET_IDS`:
`1351620000001-200050,1351620000001-200040,1351620000001-200060,1504127981769-6cnqhq,1504127981867-06dkm6,1504127981921-c2jlwt` (the first two are the new ones).

mbertrand:
just modified the above comment to match what should be on RC

Issue #810, opened 2019-12-09
Order encode presets from highest to lowest 


pdpinch:
Some video players (like the edX player) default to the first bitrate in the m3u8 playlist. This means that users on high-bitrate connections will start playing the lowest quality video and never have reason to request a better video. 

The OVS player has code to always choose the highest bitrate. 

We should change the settings to be highest bitrate to lowest.

pdpinch:
We've recently discovered that this issue also affects the OVS player when used in Chrome. So it's worth fixing for all videos, not just xPRO videos.

pdpinch:
https://github.com/mitodl/salt-ops/pull/1084 addresses this issue going forward, but we need to retroactively update the existing playlists. 

Although we could just re-encode all the videos, it would be more cost effective to just re-order the items in the `video__index.m3u8` files.

mbertrand:
@pdpinch are you sure you want to do this to all videos?  I can't reproduce the Chrome behavior described above, the highest resolution plays by default for me on both OSX and Windows 10.

pdpinch:
Is there a downside to doing it for all video? 

Elaine has reported incorrect playback in Chrome Windows, but I’m not 100% sure this is a fix. 

> @pdpinch are you sure you want to do this to all videos? I can't reproduce the Chrome behavior described above, the highest resolution plays by default for me on both OSX and Windows 10.

mbertrand:
Might as well do all videos just to make them consistent

PR #811, opened 2019-12-17
m3u8 reorder task

abeglova:
https://github.com/mitodl/odl-video-service/issues/810
#### What's this PR do?
This pr reorders m3u8 playlists (video__index.m3u8 on S3) from highest bandwidth to lowest bandwidth so that the highest bandwidth video is selected first.
Set 
ET_PRESET_IDS=1351620000001-200050,1351620000001-200040,1504127981769-6cnqhq,1504127981819-v44xlx,1504127981867-06dkm6,1504127981921-c2jlwt

in your .env file.  Upload a video. Check that the newly created video__index.m3u8 file in {VIDEO_S3_TRANSCODE_BUCKET}/transcoded/{video_key} s3 folder has the smallest bandwidth first.

From the python shell, run

```
from cloudsync.tasks import *
sort_transcoded_m3u8_files()
```

The file on s3 should be updated and should now be sorted to have the highest  bandwidth first.

(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/cloudsync/tasks.py
+++ b/cloudsync/tasks.py
@@ -21,6 +21,7 @@ from cloudsync.youtube import YouTubeApi, API_QUOTA_ERROR_MSG
 from ui.models import Video, YouTubeVideo, VideoSubtitle, Collection
 from ui.constants import VideoStatus, YouTubeStatus, StreamSource
 from ui.utils import get_bucket
+from ui.encodings import EncodingNames
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


cloudsync/tasks.py
```diff
@@ -352,3 +352,37 @@ def parse_content_metadata(response):
...
+            lines = file_content.split(delimiter)
+            header = lines.pop(0)
+
+            if str.strip(header) != "#EXTM3U":
+                continue
+
+            try:
+                lines.sort(key=lambda line: -int(re.search(r'RESOLUTION=(\d+)', line).group(1)))
+            except AttributeError:
+                continue
```
mbertrand: Maybe add a log statement here to give an indication of what happened, or will this be a common noisy error?

cloudsync/tasks.py
```diff
@@ -352,3 +352,37 @@ def parse_content_metadata(response):
...
+
+            delimiter = '#EXT-X-STREAM-INF:'
+            lines = file_content.split(delimiter)
+            header = lines.pop(0)
+
+            if str.strip(header) != "#EXTM3U":
+                continue
+
+            try:
+                lines.sort(key=lambda line: -int(re.search(r'RESOLUTION=(\d+)', line).group(1)))
```
mbertrand: Might also need to sort by bandwidth, there are two codecs with the same resolution but one has a lower bandwidth than the other:
```
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=471000,RESOLUTION=320x176,CODECS="avc1.42001e,mp4a.40.2"
video_1351620000001-200050.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=682000,RESOLUTION=320x176,CODECS="avc1.42001e,mp4a.40.2"
video_1351620000001-200040.m3u8
```

After reordering, the lower bandwidth codec precedes the higher bandwidth code in some videos but not others

cloudsync/tasks.py
abeglova: I don't think this will be a common error, I adding a log statement

mbertrand:
PS A benefit of chunking out the videos would be we could process a subset of videos too, instead of all of them, for testing purposes.

Something like:
```python
@app.task(bind=True)
def sort_video_indexes(self):
    index_tasks = celery.group(
            [
                sort_transcoded_m3u8_files.si(video_ids)
                for video_ids in chunks(
                    Video.objects.order_by("id").values_list("id", flat=True),
                    chunk_size=settings.ELASTICSEARCH_INDEXING_CHUNK_SIZE,
                )
            ]
    .....


@app.task(bind=True)
def sort_transcoded_m3u8_files(video_ids):
    for video in Video.objects.filter(id__in=video_ids).iterator():
        ....
```

mbertrand:
One way to reduce the number of videos processed on production is to filter by those that have an HLS encoding:

`for video in Video.objects.filter(videofile__encoding=EncodingNames.HLS).iterator():`

PR #835, opened 2020-03-20
Enabled edX course ID editing for collections

### LINKED ISSUE #830: Allow collection owners to connect edX courses
[Trello](https://trello.com/c/FsbxNfz5/2-allow-collection-owners-to-connect-edx-courses)

As an xPRO media member, I'd like to be able to create collections myself and set which course they connect to.

Currently, I have to ask an admin to do this for me.

gsidebo:
Closes #830 

#### What's this PR do?
Allows users to add/edit the `edx_course_id` value for a collection if **either** of the following are true:
- They are superusers/staff
- They are a member of a collection's admin Moira lists and have a certain permission group added to their user record

Test the superuser/staff case:
- Confirm that you can add/edit the edx Course ID value for a new/existing collection if your user has the `is_staff` and/or `is_superuser` flags set to `True`

Test the non-superuser/staff case:
- Log in a user with `is_staff` and`is_superuser` set to `False`
- Add that user is added to one of the admin Moira lists for some collection
- Visit that collection detail page in the app and click the gear icon. You **should not** see the edX Course ID field at the bottom of the pop up
- In Django admin, add the "edX Course Admin" group for your user (that group should have been created in a data migration)
- Visit the same page and open up the form
- You **should** see the edX Course ID field at the bottom of the pop up, and changing the value and submitting the form should change the value in the database

ss 2020-03-20 at 14 30 00


ui/utils.py
```diff
@@ -127,6 +127,24 @@ def user_moira_lists(user):
     return list_names
 
 
+def list_members(list_name):
```
gsidebo: Moved. No changes made

ui/views.py
```diff
@@ -240,10 +249,6 @@ class HelpPageView(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         js_settings = default_js_settings(self.request)
-        if not self.request.user.is_anonymous:
-            has_admin = Collection.objects.all_admin(self.request.user).exists()
-            is_staff_or_superuser = ui_permissions.is_staff_or_superuser(self.request.user)
-            js_settings["editable"] = has_admin or is_staff_or_superuser
```
gsidebo: I could not find a single place where this context variable was being used for the help page or the TOS page

docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
       NODE_ENV: 'development'
       PORT: 8087
       COVERAGE_DIR: htmlcov
-      DATABASE_URL: postgres://postgres@db:5432/postgres
+      DATABASE_URL: ${DATABASE_URL:-postgres://postgres@db:5432/postgres}
```
mbertrand: I noticed that this line is still `DATABASE_URL: postgres://postgres@db:5432/postgres` in xpro and open, is there a reason for changing it here?

gsidebo: In this project and in xPro, I was continuously getting a Postgres error, likely brought about as some kind of side effect of our Python/Postgres version upgrades. The error was something like "Database is uninitialized and superuser password is not specified.". Allowing for a `.env` override of the `DATABASE_URL` setting allowed me to just specify the password in the connection string itself.

Issue #842, opened 2020-03-23
Remove 'public' option for videos in front end


gsidebo:
https://trello.com/c/BiJVx34C/18-remove-public-option-in-video-settings-ui-for-now

PR #843, opened 2020-03-23
Removed 'public' option for videos in front end

gsidebo:
Closes https://github.com/mitodl/odl-video-service/issues/842

#### What's this PR do?
See title!

Edit all aspects of any video via the form in the front end and confirm that those updates are still working

**OLD** 
ss 2020-03-23 at 14 59 25

**NEW**
ss 2020-03-23 at 17 17 50


pdpinch:
We’d like to bring the feature back, but it will almost certainly be modified — at least in UI and possibly in functionality. 

I don’t feel strongly, but my instinct is that it’s not worth the effort for a feature flag.

PR #856, opened 2020-03-25
Add keyboard control to video player

### LINKED ISSUE #854: Video keyboard control
Need to add keyboard support for the video lectures.

Here are the possible controls avaialble:

- [ ]  Space bar toggles play/pause.
- [ ] - Right and Left Arrow keys seek the video forwards and back.
- [ ] - Up and Down Arrow keys increase and decrease the volume.
- [ ] - M key toggles mute/unmute.
- [ ] - F key toggles fullscreen off and on. (Does not work in Internet Explorer, it seems to be a limitation where scripts cannot request fullscreen without a mouse click)
- [ ] - Double-clicking with the mouse toggles fullscreen off and on.
- [ ] - Number keys from 0-9 skip to a percentage of the video. 0 is 0% and 9 is 90%.

annagav:
Fix #854

#### What's this PR do?
add `videojs-hotkeys` plugin

Open any video and try to control volume, skip seconds, toggle full screen etc.

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -397,8 +397,8 @@ class VideoPlayer extends React.Component<*, void> {
     this.updateSubtitles()
 
     this.player.hotkeys({
-      volumeStep: 0.1,
-      seekStep: 5,
+      volumeStep:                0.1,
+      seekStep:                  5,
       enableModifiersForNumbers: false
     })
   }
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -395,6 +395,12 @@ class VideoPlayer extends React.Component<*, void> {
       this.checkYouTube()
     }
     this.updateSubtitles()
+
+    this.player.hotkeys({
+      volumeStep: 0.1,
+      seekStep: 5,
+      enableModifiersForNumbers: false
+    })
   }
 
   componentDidUpdate() {
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -366,6 +366,11 @@ class VideoPlayer extends React.Component<*, void> {
       makeConfigForVideo(video, useYouTube, embed, startTime),
       function onPlayerReady() {
         this.enableTouchActivity()
+        this.hotkeys({
+          volumeStep:                0.1,
+          seekStep:                  5,
+          enableModifiersForNumbers: false
+        })
         if (video.multiangle) {
           setCustomDimension(SETTINGS.ga_dimension_camera, selectedCorner)
           this.on("loadeddata", cropVideo)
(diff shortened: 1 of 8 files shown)


ui/templates/base.html
```diff
@@ -37,5 +37,6 @@
     </ul>
   {% endif %}
   {% block content %}{% endblock %}
+  <script src="https://cdn.sc.gl/videojs-hotkeys/latest/videojs.hotkeys.min.js"></script>
```
mbertrand: Instead of including it this way, can you add it to `package.json` via:
```
docker-compose run watch yarn add videojs-hotkeys@^0.2.25
```
and then add the following to `static/lib/video.js`:
```
require("videojs-hotkeys")
```

pdpinch:
Does this change the UI at all? How should users find out about these keyboard commands?

annagav:
@pdpinch, not there are no changes to the ui. The controls are pretty standard, you can control videos the same way on Youtube.

static/js/components/VideoPlayer.js
```diff
@@ -395,6 +395,12 @@ class VideoPlayer extends React.Component<*, void> {
       this.checkYouTube()
     }
     this.updateSubtitles()
+
+    this.player.hotkeys({
+      volumeStep:                0.1,
+      seekStep:                  5,
+      enableModifiersForNumbers: false
+    })
```
mbertrand: Might be better to call this function only when the player is ready, to make sure it's initialized:
```javascript
      function onPlayerReady() {
        this.enableTouchActivity()
        this.hotkeys({
          volumeStep:                0.1,
          seekStep:                  5,
          enableModifiersForNumbers: false
        })
```

static/js/components/VideoPlayer_test.js
```diff
@@ -66,6 +66,7 @@ describe("VideoPlayer", () => {
       fluid:         sandbox.stub().returns(playerStub),
       width:         sandbox.stub(),
       height:        sandbox.stub(),
+      hotkeys:       sandbox.stub(),
```
mbertrand: if `this.hotkeys(...)` is called within `onPlayerReady`, you can change the "uses videojs on mount with the right arguments" test as follows:
```javascript
        const hotkeysStub = sandbox.stub()
        args[2].call({
          enableTouchActivity: enableTouchActivityStub,
          on:                  onStub,
          tech_:               { hls: {} },
          hotkeys:             hotkeysStub
        })
        sinon.assert.calledWith(hotkeysStub, {
          volumeStep:                0.1,
          seekStep:                  5,
          enableModifiersForNumbers: false
        })
```

static/js/containers/VideoDetailPage_test.js
```diff
@@ -42,8 +42,25 @@ describe("VideoDetailPage", () => {
...
+        return this.tracks
+      },
+      removeRemoteTextTrack: function(track) {
+        this.tracks.splice(this.tracks.indexOf(track), 1)
+      },
+      addRemoteTextTrack: function(track) {
+        this.tracks.push({ src: track.src, addEventListener: function() {} })
+      }
+    }
+    sandbox.stub(libVideo, "videojs").returns(playerStub)
```
mbertrand: if this.hotkeys(...) is called within onPlayerReady, you can remove this addition to the test

PR #858, opened 2020-03-25
Fix video source switch failover

### LINKED ISSUE #857: Youtube failover to cloudfront no longer works
If Youtube is unavailable for some reason, VideoJS should switch to the Cloudfront source for the video.  This no longer works, `player.reset()` is no longer recognized as a function.

mbertrand:
Closes #857 

#### What's this PR do?
Modifies the `VideoPlayer.switchVideoSource` function to call `player.src()` instead of `player.reset().src()`

- Set up a video to play from Youtube:
  -  In django admin, create a new YoutubeVideo object, associate with an existing video, enter any valid Youtube video id, and set status to `processed`
  - Set the related video to `is_public=True`
- Go to the video detail page, it should play the Youtube video.
- Using an adblocker like uBlock, blacklist `*youtube.com`
- Go to the video detail page again, it should play the Cloudfront version.

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -269,7 +269,7 @@ class VideoPlayer extends React.Component<*, void> {
   switchVideoSource = () => {
     const { video } = this.props
     if (video.sources.length > 0) {
-      this.player.reset().src(video.sources)
+      this.player.src(video.sources)
     }
   }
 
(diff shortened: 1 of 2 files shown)


gsidebo:
@mbertrand Works when I use `*youtube.com` as the blacklist pattern, but when I use `www.youtube.com` (as is the default when you block a domain from the Chrome dev tools Network tab, the page looks like this and clicking the play icon does nothing. The thumbnail is not the cloudfront version. Should we consider that a won't-fix?

mbertrand:
@gsidebo assumption is that if youtube is blocked in a country, it won't just be "www.youtube.com", code checks to see if the thumbnail for a video is loadable (`img.youtube.com`) to determine if the failover should occur.

PR #860, opened 2020-03-25
Force non-native HLS playback to fix quality selector in Edge, Safari

### LINKED ISSUE #859: No quality setting in Safari
As a Safari user I would like to be able to change the video quality so I can control the bitrate I'm seeing.

This works in Chrome and Firefox, where there is a gear menu at the bottom right of the player. The menu does not appear in Safari, and possibly not in MS Edge either.

mbertrand:
- Fixes #859

#### What's this PR do?
Forces non-native HLS playback, because it is not supported by the videojs-hls-quality-selector plugin.  Edge and Safari will use native playback by default otherwise.

In Safari or Edge, open a video detail page.  The control bar should have a button to select different quality levels, and it should work.
Everything should remain normal in FF and Chrome.

https://www.npmjs.com/package/videojs-hls-quality-selector

### DIFF
--- a/static/js/components/VideoPlayer.js
+++ b/static/js/components/VideoPlayer.js
@@ -37,7 +37,7 @@ const makeConfigForVideo = (
   techOrder:   useYouTube ? ["youtube", "html5"] : ["html5"],
   html5:       {
     nativeTextTracks: false,
-    hls: {
+    hls:              {
       overrideNative: true
     }
   },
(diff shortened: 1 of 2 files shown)


pdpinch:
Wow, that’a big trade off. We have to give up native HLS playback in order to get bitrate control?


What does that mean for user experience?

mbertrand:
@pdpinch I didn't notice any difference at all.  The video still plays via HLS, but if I understand correctly, uses VideoJS's included HLS plugin instead of the browser's own HLS code.  I think the reason why the quality selector already works in FF and Chrome is because it doesn't play natively in those.

PR #864, opened 2020-03-27
Add videojs-annotation-comments plugin and put it behind a feature flag

noisecapella:
https://trello.com/c/YlaO3AvR/25-annotation-research

#### What's this PR do?
Adds the videojs-annotation-comments plugin and put it behind a feature flag so that we can try it out

Nothing should break by default. Set `FEATURE_VIDEOJS_ANNOTATIONS=True` and you should see a simple annotations UI when you look at videos using videojs. You can make changes but they will not be persisted anywhere and will go away when the browser refreshes.

### DIFF
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -11,9 +11,6 @@
     {% load raven %}
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
     <link rel="stylesheet" type="text/css" href="{% static 'hijack/hijack-styles.css' %}" />
-      {% if video_annotations %}
-        <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
-      {% endif %}
       <script type="text/javascript">
       var SETTINGS = {{ js_settings_json|safe }};
     </script>
--- a/ui/templates/base.html
+++ b/ui/templates/base.html
@@ -11,7 +11,7 @@
     {% load raven %}
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
     <link rel="stylesheet" type="text/css" href="{% static 'hijack/hijack-styles.css' %}" />
-      <script type="text/javascript">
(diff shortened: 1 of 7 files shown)


ui/views.py
```diff
@@ -70,8 +71,10 @@ def default_js_settings(request):
         ),
         "support_email_address": settings.EMAIL_SUPPORT,
         "ga_dimension_camera": settings.GA_DIMENSION_CAMERA,
+        "video_annotations": video_annotations_feature_flag,
```
gsidebo: Do we need this redundancy? Can we just check for `FEATURES["VIDEOJS_ANNOTATIONS"]` in the template?

ui/templates/base.html
```diff
@@ -11,7 +11,10 @@
     {% load raven %}
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
     <link rel="stylesheet" type="text/css" href="{% static 'hijack/hijack-styles.css' %}" />
-    <script type="text/javascript">
+      {% if video_annotations %}
+        <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
```
gsidebo: I'm guessing this means that this `videojs-annotation-comments` plugin requires jQuery? Could you not add this dependency as a `require` call in `static/js/lib/video.js`?

noisecapella:
Comments should be addressed. The template variable was actually set as a JS variable so the jquery link was never inserted, which is probably why it didn't work for you. Can you try it again?

ui/templates/base.html
```diff
@@ -11,7 +11,7 @@
     {% load raven %}
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
     <link rel="stylesheet" type="text/css" href="{% static 'hijack/hijack-styles.css' %}" />
-    <script type="text/javascript">
+      <script type="text/javascript">
```
gsidebo: Seems like an unintentional change

PR #879, opened 2020-04-22
Fix TechTV embed URLs

mbertrand:
N/A

#### What's this PR do?
Fixes Techtv embed URL's

```Python
from ui.models import *
from techtv2ovs.models import *
ttc = TechTVCollection.objects.create(id=123, collection=Collection.objects.first())
ttv = TechTVVideo.objects.create(ttv_id=987, ttv_collection=ttc, video=Video.objects.last(), video_status='Complete', status = "Complete", title="TTV Video")
```
 - Go to `http://localhost:8089/embeds/987/`, it should load the video embed page
- Go to `http://localhost:8089/collections/123/videos/987/`, it should load the video detail page

### DIFF
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -57,6 +57,11 @@ class App extends React.Component<*, void> {
           path={`${match.url}videos/:videoKey/embed/`}
           component={this.renderVideoEmbedPage}
         />
+        <Route
+          exact
+          path={`${match.url}embeds/:videoKey/`}
+          component={this.renderVideoEmbedPage}
+        />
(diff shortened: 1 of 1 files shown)


pdpinch:
Is it possible to add a test to avoid this breaking again?

mbertrand:
@pdpinch a test for this would require adding in something like the `TestIntegrationHelper` from open-discussions (in whole or in part) into OVS and adjusting accordingly, which may take some time.  If this needs to be deployed ASAP it might be better to do that in a separate PR, but if there's no hurry then I can work on including it here.

PR #883, opened 2020-04-29
Add optional logging to file, not stdio

markbreedlove:

https://trello.com/c/qEbo79DI/78-add-odl-video-service-uwsgi-log-to-fluentd

#### What's this PR do?

Add an optional file destination for logging, instead of the default stderr stream. This can be used to keep app log messages separate from uWSGI middleware log messages, which gives us better control over the formatting of the lines that we want to have JSON-formatted in production. The messages coming from uWSGI itself can continue to have their own particular format.

There should be no change in behavior if you change nothing in your configuration. If you define the `ODL_VIDEO_LOG_FILE` environment variable, you should see a log getting written out that contains only the messages from our application's `logger` calls, not mixed in with the uWSGI messages.

### DIFF
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -234,6 +234,7 @@ else:
 # Logging configuration
 LOG_LEVEL = get_string('ODL_VIDEO_LOG_LEVEL', 'INFO')
 DJANGO_LOG_LEVEL = get_string('DJANGO_LOG_LEVEL', 'INFO')
+LOG_FILE = get_string('ODL_VIDEO_LOG_FILE', None)
 
 # For logging to a remote syslog host
 LOG_HOST = get_string('ODL_VIDEO_LOG_HOST', 'localhost')
--- a/odl_video/logging.py
+++ b/odl_video/logging.py
@@ -4,9 +4,15 @@ from structlog_sentry import SentryJsonProcessor
 import structlog
 from odl_video import settings
 
+log_config_args = {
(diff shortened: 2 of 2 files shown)


blarghmatey:
This is fine as a first pass, but we'll likely want to add in a logging dictconfig and specify a rotating file handler, unless you intend to handle the rotation with logrotate.

blarghmatey:
:+1:

markbreedlove:
I'll check quickly about adding the rotating file handler. If it's convenient, then it will be better than using the copy-truncate method, which is what I'd do if I used logrotate.

markbreedlove:
It's turning out to be more complex than I'd hoped and not working as advertised in a test in the interpreter, so I can make a ticket to follow up and investigate the rotating handler.

odl_video/settings.py
```diff
@@ -234,6 +234,7 @@
 # Logging configuration
 LOG_LEVEL = get_string('ODL_VIDEO_LOG_LEVEL', 'INFO')
 DJANGO_LOG_LEVEL = get_string('DJANGO_LOG_LEVEL', 'INFO')
+LOG_FILE = get_string('ODL_VIDEO_LOG_FILE', None)
```
mbertrand, line 237: This should be added to `app.json`

blarghmatey: It doesn't make sense to add it to app.json because we wouldn't ever want to use a log file on Heroku. This setting is only useful on EC2.

PR #909, opened 2020-08-20
Make collection autocomplete field for video

### LINKED ISSUE #906: use autocomplete_fields for collection in video admin
As an admin, I sometimes need to move dozens of videos between collections. I can do this in the admin, but it is cumbersome to choose a collection from the drop-down list with hundreds of items. 

Instead of a drop-down list, we should try the new in django 2 feature, autocomplete_fields. See https://docs.djangoproject.com/en/2.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields

annagav:
Fix #906
#### What's this PR do?
Add autocomplete field to VideoAdmin model, to search for collections easier.

go to admin, find an existing video, try to update the collection field.

### DIFF
--- a/ui/admin.py
+++ b/ui/admin.py
@@ -161,7 +161,7 @@ class VideoAdmin(admin.ModelAdmin):
         VideoSubtitlesInline,
         VideoThumbnailsInline
     ]
-    autocomplete_fields = ['view_lists']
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

PR #970, opened 2021-11-12
Increased encrypted char fields to text fields

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
We discovered in RC that secret keys are larger than 100 chars so we widened this into a text field.

Migrations should run and the app should start


annagav (approved):
👍

PR #1028, opened 2023-01-04
Only upload videos to edx if the collection defines course ID

collinpreston:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Fixes an issue where OVS would attempt to register videos with edx even if the collection the video was uploaded to did not define a course ID.

1. Follow the readme to setup OVS and edx locally.
2. Create a collection but do not define a course ID for the collection.
3. Upload a video to the collection.
4. Verify that the logs for OVS do not show "This video file cannot be added to edX".

It is also worth confirming that you can reproduce the error locally by taking the master branch of OVS and repeating the steps above in order to verify that is  "This video file cannot be added to edX" is output to the OVS logs. 

Also verify that, with a course ID defined for the collection, uploaded videos are registered with edx.

### DIFF
--- a/ui/signals.py
+++ b/ui/signals.py
@@ -93,7 +93,7 @@ def add_video_to_edx(sender, instance, created, **kwargs):
     """
     If a Video was updated with a status of COMPLETE, we can now upload the related VideoFiles.
     """
-    if instance.status == VideoStatus.COMPLETE and instance.collection.edx_course_id is not None:
+    if instance.status == VideoStatus.COMPLETE and instance.collection.edx_course_id != '':
         ovs_tasks.post_video_to_edx.delay(instance.id)
 
 
--- a/ui/signals.py
+++ b/ui/signals.py
@@ -93,7 +93,7 @@ def add_video_to_edx(sender, instance, created, **kwargs):
     """
     If a Video was updated with a status of COMPLETE, we can now upload the related VideoFiles.
     """
-    if instance.status == VideoStatus.COMPLETE:
+    if instance.status == VideoStatus.COMPLETE and instance.collection.edx_course_id is not None:
         ovs_tasks.post_video_to_edx.delay(instance.id)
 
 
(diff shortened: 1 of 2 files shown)


ui/signals.py
```diff
@@ -93,7 +93,10 @@ def add_video_to_edx(sender, instance, created, **kwargs):
     """
     If a Video was updated with a status of COMPLETE, we can now upload the related VideoFiles.
     """
-    if instance.status == VideoStatus.COMPLETE:
+    if (
+        instance.status == VideoStatus.COMPLETE
+        and instance.collection.edx_course_id != ""
```
mbertrand, line 98: `edx_course_id` could be `None`, so might be safer to do this:
```python
and instance.collection.edx_course_id
```

ui/signals_test.py
```diff
@@ -91,11 +92,13 @@ def test_youtube_sync_redo_failed(mocker, video_with_file, status):
     assert mock_delete.call_count == expected_count
 
 
-def test_edx_video_file_signal(mocker):
+@pytest.mark.parametrize("edx_course_id", ["123", ""])
```
mbertrand, line 95:
```python
@pytest.mark.parametrize("edx_course_id", ["123", "", None])
```

ui/signals_test.py
```diff
@@ -106,7 +109,10 @@ def test_edx_video_file_signal(mocker):
     )
     video.status = VideoStatus.COMPLETE
     video.save()
-    patched_edx_task.assert_called_once_with(video.id)
+    if edx_course_id == "":
```
mbertrand, line 112:
```python
if not edx_course_id:
```

mbertrand (changes requested):
` ui.tasks.post_video_to_edx` is still being called because my collection has a `edx_course_id` of `None` instead of ""

ui/signals.py
```diff
@@ -93,7 +93,10 @@ def add_video_to_edx(sender, instance, created, **kwargs):
     """
     If a Video was updated with a status of COMPLETE, we can now upload the related VideoFiles.
```
mbertrand, line 94: "we can now post the video to edx"?

mbertrand (approved):
👍

Issue #1027, opened 2022-12-23
Can't upload more than 5 files from dropbox


pdpinch:
### Steps to Reproduce

1. Create a new collection
2. Upload the 32 videos from https://www.dropbox.com/sh/27n8kzccjxv7dm1/AADn0oY1V2gjjZXkbuP8gu_Sa?dl=0

### Expected Behavior

- All 32 videos get queued for transcode

### Actual Behavior

- Only 5 videos got queued
- I have to reload the page to see what's been queued

pdpinch:
The number of videos that get queued seems to be variable. On a subsequent attempt, I uploaded 16 and 7 were queued.

pdpinch:
No apparent errors in Sentry, but when I watch in my browser, I see a 504 response from https://video.odl.mit.edu/api/v0/upload_videos/

mbertrand:
I think the [celery task to stream videos to S3 and start transcoding jobs](https://github.com/mitodl/odl-video-service/blob/f4076d2025d5e1db9c0a91c09f141a845b564343/ui/api.py#L49-L52) is not running asynchronously:

```
        # Kick off chained async celery tasks to transfer file to S3, then start a transcode job
        task_result = chain(
            tasks.stream_to_s3.s(video.id), tasks.transcode_from_s3.si(video.id)
        )()
```

The `()` at the end should probably be replaced with `.apply_async()`

pdpinch:
@collinpreston @mbertrand do you think we can close this?

mbertrand:
I think so, it seems to be working well now

PR #1030, opened 2023-01-05
call the celery task as async

collinpreston:
#### Pre-Flight checklist
- [x] Testing
 
https://github.com/mitodl/odl-video-service/issues/1027

#### What's this PR do?
This PR changes the chain of celery tasks, that are called when adding videos, to be asynchronous.  It also removes any dependency on those tasks in the returned response in order to (hopefully) speed up the response from the `upload` endpoint.

1. Have OVS running locally.
2. Have a large number of videos and an empty collection created in OVS.
3. In one operation, select all of your videos from dropbox and add them to your collection.
4. Verify that all of the videos are added to the collection and encoded properly.

I was not able to reproduce the error from the issue when testing locally.  Using the same files I use locally, I can reproduce the error consistently in RC.  I would expect that testing this PR locally should be limited to just ensuring that nothing is broken.  Testing to see whether this change resolves the issue seen in RC and Production will need to wait until this is deployed to those environments.

### DIFF
--- a/ui/api.py
+++ b/ui/api.py
@@ -47,14 +47,13 @@ def process_dropbox_data(dropbox_upload_data):
                 bucket_name=settings.VIDEO_S3_BUCKET,
             )
         # Kick off chained async celery tasks to transfer file to S3, then start a transcode job
-        task_result = chain(
+        chain(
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
LGTM 👍

PR #1032, opened 2023-01-12
Reorganizing the Dockerfile a bit to be a little more coherent. Remov…

Ardiea:
Reorganizing the Dockerfile a bit to be a little more coherent. Removing some un-needed steps from the Dockerfile. Added a step to explicitly copy webpack-stats.json into the 'production' build.

#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

OL Inf 1249 https://github.com/mitodl/ol-infrastructure/issues/1249 

#### What's this PR do?
Tweaks the docker file further beyond Tobias' changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM node:13.13.0 as node
-ENV NODE_ENV=development
+ENV NODE_ENV=production
 RUN apt-get update && apt-get install libelf1 -y
 COPY . /src
 WORKDIR /src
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
LGTM

Dockerfile
```diff
@@ -1,5 +1,12 @@
-FROM python:3.9.13 AS base
+FROM node:13.13.0 as node
+ENV NODE_ENV=development
```
blarghmatey, line 2: Is the intent to override this value in the pipeline definition? Otherwise the produced assets won't be minified.

Ardiea: Oh good catch. This was a failed threadchase from debugging yesterday. This should be 'production'.

collinpreston (approved):
Looks good!  Thank you

PR #1064, opened 2023-06-27
fix,version: Update version constraints for Poetry to be more accurate

blarghmatey:
# What are the relevant tickets?
N/A

# Description (What does it do?)
There were a number of constraints using `>=` which is dangerous. Updated to use `^` which constrains to be within a given major or minor version specifier.

# How can this be tested?
Deploy to RC and verify that the app works properly


pyproject.toml
```diff
@@ -32,63 +32,63 @@ packages = []
...
+sentry-sdk = "^0.14.3"
+smart-open = "^1.5.7"
+structlog = "^20.1.0"
+structlog-sentry = "^1.2.2"
+urllib3 = "^1.24.2"
 uwsgi = "2.0.21"
 
 
 [tool.poetry.group.dev.dependencies]
-black = "22.8.0"
```
mbertrand, line 75: Can we keep this pegged to a particular version, so as to avoid potential failures when the default `black` criteria change?

poetry.lock
```diff
@@ -2276,32 +2412,32 @@ files = [
 
 [[package]]
 name = "psycopg2"
-version = "2.8.6"
+version = "2.9.6"
```
mbertrand, line 2415: version `2.8.6` is specified in the pyproject.toml file

mbertrand (commented):
Some preliminary feedback

mbertrand (approved):
👍

PR #1083, opened 2024-03-19
Add an additional layer of retries to get a moira client beyond the default 3 for urllib

mbertrand:
Potentially addresses https://github.com/mitodl/hq/issues/3810#issuecomment-2007305350 

The current request to get a moira client will retry up to 3x (the urllib library default) before giving up.  This PR adds an additional layer of retries, defaulting to 3 (3x3 = 9) which can be adjusted via a new setting (`MOIRA_RETRIES`).

### How can this be tested?
Unit tests should pass.  You would need current .env settings for `MIT_WS_CERTIFICATE_FILE` and `MIT_WS_PRIVATE_KEY_FILE` to be able to connect to moira successfully but without that you should see repeated attempts to connect in the log.


mbertrand:
Closing, moira issue has been fixed

PR #1162, opened 2025-03-10
unique_together constrain added for CollectionEdxEndpoint bridge Table

marslanabdulrauf:
https://github.com/mitodl/hq/issues/4102

This PR adds `unique_together` constrain in `CollectionEdxEndpoint` bridge table to avoid entries for same collection and edx_endpoint

### How can this be tested?
1. Go to admin and collection edx endpoint -> /admin/ui/collectionedxendpoint
2. Try to add new entry with existing rows collection and edx_endpoint

### Additional Context
We need to **delete any Duplicate entry before merging** this PR
We can identify duplicates by running this in django shell:
```
from django.db.models import Count
from ui.models import CollectionEdxEndpoint

CollectionEdxEndpoint.objects.values('collection', 'edx_endpoint').annotate(count=Count('id')).filter(count__gt=1)
```


mbertrand (approved):
Works well, but wondering if the ability to add/edit/remove `CollectionEdxEdpoint` objects directly should be prohibited? Either by removing `CollectionEdxEdpointAdmin` altogether or making it read-only (see below), so that collections can be assigned to edx endpoints only via the inline collection edx endpoint section under Collection Admin?

But I'm not familiar with how these associations are typically created, if they are usually done via `CollectionEdxEdpointAdmin` then okay to leave as is (so approving now in case that's true)

```python
class CollectionEdxEndpointAdmin(admin.ModelAdmin):
    """CollectionEdxEndpoint admin"""
....
    def has_add_permission(self, request, obj=None):
        """Prevent adding new objects"""
        return False
        
    def has_change_permission(self, request, obj=None):
        """Allow viewing but not editing objects"""
        return False
        
    def has_delete_permission(self, request, obj=None):
        """Prevent deleting objects"""
        return False
```
Screenshot 2025-03-10 132428

marslanabdulrauf:
Thanks @mbertrand for taking some time and review the PR.

May be someone who is using OVS can answer better on [above comment](https://github.com/mitodl/odl-video-service/pull/1162#pullrequestreview-2671734159)

Cc @pdpinch @cachob

cachob:
@marslanabdulrauf / @mbertrand 

Those associations are done manually. Currently when I get a request to add videos to a collection, I create those endpoints (either for xpro or mitxonline) using that admin in django that you pointed out. It would be great if this step could be automated because someone like Chris B (who usually works on this) tends to forget this step as a not so technical user.

mbertrand:
@marslanabdulrauf I don't currently have shell access to production, since it is not on heroku, but devops might be able to help with that, will talk to him about running those shell commands.

marslanabdulrauf:
Thanks @mbertrand

PR #1184, opened 2025-04-15
SECURE_CROSS_ORIGIN_OPENER_POLICY setting added

marslanabdulrauf:
https://github.com/mitodl/hq/issues/7110#issuecomment-2801529301

This PR:
1. Adds `SECURE_CROSS_ORIGIN_OPENER_POLICY` in settings
2. And make it configureable via environment

image

### How can this be tested?
1. Checkout to this branch
2. run `docker-compose up --build -d`
3. Check network calls and verify if the response headers have `cross-origin-opener-policy` header and it is set to the value you provided

### DIFF
--- a/odl_video/settings.py
+++ b/odl_video/settings.py
@@ -572,7 +572,7 @@ DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 SECURE_CROSS_ORIGIN_OPENER_POLICY = get_string(
     "SECURE_CROSS_ORIGIN_OPENER_POLICY",
-    "same-origin",
+    "same-origin-allow-popups",
 )
 
 # Hijack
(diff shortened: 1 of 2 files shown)


odl_video/settings.py
```diff
@@ -570,5 +570,10 @@ def get_all_config_keys():
 
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
+SECURE_CROSS_ORIGIN_OPENER_POLICY = get_string(
+    "SECURE_CROSS_ORIGIN_OPENER_POLICY",
+    "same-origin",
```
blarghmatey, line 575: We should default this to the value that we want, which is "same-origin-allow-popups", otherwise we risk regressing again.

PR #1188, opened 2025-04-25
pre-commit configs updated with Ruff

marslanabdulrauf:
https://github.com/mitodl/hq/issues/7143

This PR:
1. Replaces pylint with Ruff
2. Updates pre-commit
3. Runs pre-commit on all files


umar8hassan (approved):
Looks Good 👍

PR #1189, opened 2025-04-30
Video matching from edX to OVS fixed

marslanabdulrauf:
https://github.com/mitodl/hq/issues/5334

This PR:
1. Fix the matching query to find videos

### How can this be tested?
1. Make sure your setup is complete and integrated with edX

**Management Command**
1. Replace [Line](https://github.com/mitodl/odl-video-service/blob/master/ui/api.py#L106) with `"edx_video_id": str(uuid4(),` make sure to import `uuid4`
3. Upload video --  Verify from edX. you can also use API `/api/val/v0/videos/` to get the list of videos from edX
4. OVS video key and edX video id should be different
5. Run the management command `python manage.py sync_video_key_with_edx`
6. Validate that now edX and OVS have same keys

**Overall Testing**
1. Please follow the steps from above till you have different keys on both systems
2. Run re-transcode on the Video to verify if everything is working fine
3. Run Management command to sync the keys
4. Run re-transcode again to see if everything is working as expected.

### DIFF
--- a/ui/management/commands/sync_video_key_with_edx.py
+++ b/ui/management/commands/sync_video_key_with_edx.py
@@ -1,5 +1,6 @@
 """Management command to sync video keys with edX"""
 
+from datetime import datetime
 from urllib.parse import urlencode
 import requests
 from django.core.management.base import BaseCommand
@@ -60,15 +61,16 @@ class Command(BaseCommand):
                     )
                     self.stdout.write(self.style.ERROR(str(exc)))
 
-            group_by_key = {}
+            latest_videos_by_created_at = {}
             for video in course_videos:
                 key = video.get("encoded_videos", [{}])[0].get("url", "").split("/")[-2]
-                group_by_key.setdefault(key, []).append(video)
-
-            latest_videos_by_created_at = {
-                key: max(videos, key=lambda x: x.get("created", "1970-01-01"))
-                for key, videos in group_by_key.items()
-            }
+                created = datetime.fromisoformat(video.get("created", "1970-01-01"))
+                if (
+                    key and key not in latest_videos_by_created_at
+                ) or created > datetime.fromisoformat(
+                    latest_videos_by_created_at[key].get("created", "1970-01-01")
+                ):
+                    latest_videos_by_created_at[key] = video
 
             for vid_key, vid in latest_videos_by_created_at.items():
                 Video.objects.filter(
--- a/ui/management/commands/sync_video_key_with_edx.py
+++ b/ui/management/commands/sync_video_key_with_edx.py
@@ -61,8 +61,14 @@ class Command(BaseCommand):
                     self.stdout.write(self.style.ERROR(str(exc)))
 
             for vid in course_videos:
-                Video.objects.filter(title=vid.get("client_video_id")).update(
-                    key=vid.get("edx_video_id")
+                vid_key = (
+                    vid.get("encoded_videos", [{}])[0].get("url", "").split("/")[-2]
+                )
+                Video.objects.filter(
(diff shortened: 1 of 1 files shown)


ui/management/commands/sync_video_key_with_edx.py
```diff
@@ -61,8 +61,14 @@ def handle(self, *args, **options):
...
 
             for vid in course_videos:
-                Video.objects.filter(title=vid.get("client_video_id")).update(
-                    key=vid.get("edx_video_id")
+                vid_key = (
+                    vid.get("encoded_videos", [{}])[0].get("url", "").split("/")[-2]
+                )
+                Video.objects.filter(
+                    title=vid.get("client_video_id"), key=vid_key
+                ).update(key=vid.get("edx_video_id"))
```
umar8hassan, line 69: @marslanabdulrauf, as discussed, if we have duplicate data in edX after re-transcoding, I assume we should update the OVS Video object with the latest relevant `edx_video_id`. However, in this loop, the Video object is updated with the `2nd` ID it gets from the EdX API. Because subsequent loops do not find any object with `key=vid_key` after its key has been updated. We get a time field, `created`, for each video in the EdX API results.
Can you confirm if we still have duplicate data in EdX? And we should use the latest? Is the video list from EdX sorted, or should we sort on `created`?

umar8hassan (changes requested):

ui/management/commands/sync_video_key_with_edx.py
```diff
@@ -60,9 +60,22 @@ def handle(self, *args, **options):
...
-                    key=vid.get("edx_video_id")
+            group_by_key = {}
+            for video in course_videos:
+                key = video.get("encoded_videos", [{}])[0].get("url", "").split("/")[-2]
+                group_by_key.setdefault(key, []).append(video)
+
+            latest_videos_by_created_at = {
+                key: max(videos, key=lambda x: x.get("created", "1970-01-01"))
+                for key, videos in group_by_key.items()
+            }
```
umar8hassan, line 71: we should avoid nested loops if there can be many videos per course. maybe try something like this: 
```
for video in course_videos:
    url = video.get("encoded_videos", [{}])[0].get("url", "")
    if not (key := url.split("/")[-2]):
        continue

    current_created = video.get("created", "1970-01-01")
    latest = latest_videos_by_key.get(key)

    if not latest or current_created > latest.get("created", "1970-01-01"):
        latest_videos_by_key[key] = video
```

umar8hassan (changes requested):

ui/management/commands/sync_video_key_with_edx.py
```diff
@@ -60,9 +61,23 @@ def handle(self, *args, **options):
...
-                    key=vid.get("edx_video_id")
+            latest_videos_by_created_at = {}
+            for video in course_videos:
+                key = video.get("encoded_videos", [{}])[0].get("url", "").split("/")[-2]
+                created = datetime.fromisoformat(video.get("created", "1970-01-01"))
+                if (
+                    key and key not in latest_videos_by_created_at
+                ) or created > datetime.fromisoformat(
+                    latest_videos_by_created_at[key].get("created", "1970-01-01")
+                ):
```
umar8hassan, line 72: Same here. If we get `key=""`, it will raise a `KeyError`. This should be as:
```suggestion
                if key and (
                    key not in latest_videos_by_created_at
                 or created > datetime.fromisoformat(
                    latest_videos_by_created_at[key].get("created", "1970-01-01")
                )):
```

umar8hassan (changes requested):

umar8hassan (approved):
Looks Good 👍

PR #1200, opened 2025-05-08
fixed path issue in retranscoding

umar8hassan:
part of https://github.com/mitodl/hq/issues/7287

This aims to fix the S3 keys saved in the DB with the incorrect prefix. 
For any video, the S3 key after transcoding/retranscoding should be with the pattern `<VIDEO_S3_TRANSCODE_PREFIX>/<uuid>/video__index.m3u8`. But the prefix is missed in the S3 keys after retranscoding.

### How can this be tested?
This PR should be tested alongwith https://github.com/mitodl/ol-django/pull/244.
Create a dist package as in the instructions and move it here.

1. Switch to this branch and spin up containers
2. Add a video via Dropbox, it should trigger transcoding for it. wait for it to complete.
3. Check from Django admin that you have the related video files.
4. Verify that S3 key pattern for files is `<VIDEO_S3_TRANSCODE_PREFIX>/<uuid>/video__index.m3u8`
5. Re-transcode the same video and verify that it still has the prefix value in the S3 keys
6. Verify in django admin that we populate thumbnail dimensions
7. Verify that the video player does not have duplicate formats 
8. Verify from job results that the thumbnail is not added for re-transcoding.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,6 @@ RUN pip install "poetry==$POETRY_VERSION"
 COPY pyproject.toml /src
 COPY poetry.lock /src
 WORKDIR /src
-COPY mitol_django_transcoding-2025.4.23.tar.gz /src/mitol_django_transcoding-2025.4.23.tar.gz
 RUN poetry install --only main
 
 # Add project
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ RUN pip install "poetry==$POETRY_VERSION"
 COPY pyproject.toml /src
 COPY poetry.lock /src
 WORKDIR /src
+COPY mitol_django_transcoding-2025.4.23.tar.gz /src/mitol_django_transcoding-2025.4.23.tar.gz
 RUN poetry install --only main
 
 # Add project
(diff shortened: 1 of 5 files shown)


marslanabdulrauf:
@umar8hassan  there is an issue while installing dependencies in [github check](https://github.com/mitodl/odl-video-service/actions/runs/14919646518/job/41912464475?pr=1200) can you look into that ?

umar8hassan:
> @umar8hassan there is an issue while installing dependencies in [github check](https://github.com/mitodl/odl-video-service/actions/runs/14919646518/job/41912464475?pr=1200) can you look into that ?

That is due to the change in the Dockerfile for local testing. I had updated it for the reviewer's convenience; otherwise, testing this locally would be difficult and lengthy. I will remove Docker changes after you have verified the working of this PR

marslanabdulrauf (changes requested):
I have tested the functionality and getting some sort of error while re-transcoding a video. Attaching SS for more information on the error

image

image

marslanabdulrauf (approved):
Thanks for the PR @umar8hassan 
LGTM 👍

PR #1209, opened 2025-05-22
Updated transcoding app in OVS

umar8hassan:
part of https://github.com/mitodl/hq/issues/7328

This update transcoding app to latest version that does not supresss client errors

### How can this be tested?
1. Switch to this branch and spin up containers
2. Add a video via Dropbox and wait for it to complete
3. Smoke test the transcoding workflow

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3371,4 +3371,4 @@ xmlsec = ["xmlsec (>=0.6.1)"]
 [metadata]
 lock-version = "2.1"
 python-versions = "^3.12"
-content-hash = "fe27f2d99451911b81f9577907d06869218a6776cfa974465d3b76b73ba42eb0"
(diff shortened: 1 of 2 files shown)


poetry.lock
```diff
@@ -1902,14 +1902,14 @@ typing-extensions = "*"
 
 [[package]]
 name = "mitol-django-transcoding"
-version = "2025.5.9"
+version = "2025.5.21"
```
marslanabdulrauf, line 1905: I think we need to update the version in `pyproject.toml` file

marslanabdulrauf (changes requested):

poetry.lock
umar8hassan: We have `^2025` in `pyproject.toml`, which accepts all versions in the year 2025. Updating the pyproject is not essential, but I have added the change for clarity

marslanabdulrauf (approved):
LGTM 👍

PR #1224, opened 2025-05-30
feat: users API added

marslanabdulrauf:
https://github.com/mitodl/hq/issues/7320#issuecomment-2913022138

This PR:
1. adds users endpoint 
2. updates collection endpoint to return owner and owner_info
3. Adds option on the frontend in edit collection dialog for the user to update the owner of the collection

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

### DIFF
--- a/ui/serializers.py
+++ b/ui/serializers.py
@@ -2,6 +2,7 @@
 serializers for ui
 """
 
+from django.contrib.auth import get_user_model
 from django.utils.translation import gettext_lazy
 from rest_framework import serializers
 from rest_framework.relations import RelatedField
@@ -11,6 +12,8 @@ from ui import models
 from ui import permissions as ui_permissions
 from ui.encodings import EncodingNames
 from ui.utils import get_moira_client, has_common_lists
+
+User = get_user_model()
 
 
 def validate_moira_lists(lists):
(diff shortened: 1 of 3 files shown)


umar8hassan (commented):
Code works fine
However, I have noticed that when i changed the `owner` for a pre-existing collection, the UI turned blank for a second until the `changes saved` notification appeared. Factoring in that it took 1-2 seconds ln local setup, it might take longer with users

marslanabdulrauf:
> Code works fine However, I have noticed that when i changed the `owner` for a pre-existing collection, the UI turned blank for a second until the `changes saved` notification appeared. Factoring in that it took 1-2 seconds ln local setup, it might take longer with users

Its same on production. We can create improvement ticket for this.

umar8hassan:
> > Code works fine However, I have noticed that when i changed the `owner` for a pre-existing collection, the UI turned blank for a second until the `changes saved` notification appeared. Factoring in that it took 1-2 seconds ln local setup, it might take longer with users
> 
> Its same on production. We can create improvement ticket for this.

@marslanabdulrauf, as discussed, this will be addressed in a follow-up ticket: https://github.com/mitodl/hq/issues/7598.

PR #1265, opened 2025-07-01
fix: Collection page when editing collection

marslanabdulrauf:
https://github.com/mitodl/hq/issues/7598#event-18383018652

This PR:
1. Shows the collection data behind Dialog as it is instead of showing blank page

### How can this be tested?
1. Checkout to this branch and restart docker
2. Go to any collection page and click on `gear` button on top right
3. Update any detail of the collection and hit `save`
4. Back page should stay same while the update is being performed.

### DIFF
--- a/static/js/reducers/collections.js
+++ b/static/js/reducers/collections.js
@@ -29,8 +29,10 @@ export const collectionsEndpoint = {
     [CLEAR_COLLECTION_ERRORS]: R.dissoc("error"),
     [CLEAR_COLLECTION_DATA]:   state => ({
       ...state,
-      loaded: false,
-      data:   null
+      processing: false,
+      loaded:     false,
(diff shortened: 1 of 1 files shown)


static/js/containers/CollectionDetailPage.js
```diff
@@ -47,6 +47,10 @@ export class CollectionDetailPage extends React.Component<*, void> {
     this.updateRequirements()
   }
 
+  componentWillUnmount() {
```
umar8hassan, line 50: A nitpick - We should add a test for this dispatch

umar8hassan (approved):
Looks Good 👍
Code works as expected. I have tested the fix locally. Also tested with a delay in API response to check for stale collection data.

PR #1289, opened 2025-07-15
feat: keycloak service added with pre-defined realm and configurations

marslanabdulrauf:
### What are the relevant tickets?
https://github.com/mitodl/hq/issues/6821

### Description (What does it do?)
This PR:
1. Adds `keycloak` service
2. `ol-local` realm configuration for keycloak 
3. Adds management command to transfer all MOIRA lists to keycloak and django and group them under the same MOIRA list name

### How can this be tested?
1. Checkout to this branch
2. Run `docker-compose up --build -d`
4. If the db is already setup then you might need to create keycloak db manually -- follow the steps mentioned in [init-db file](https://github.com/mitodl/odl-video-service/pull/1289/files#diff-0238634ee66943cdea27e176024b624c6271273ead20da3a1898158497805ac0) in db container
5. Make sure you have a running keycloak container now by using docker desktop or `docker ps` command
6. Now follow [keycloak integration README](https://github.com/mitodl/odl-video-service/pull/1289/files#diff-c0ab59f153b41fe300ae16e9bd2fb9da5c33e8d1112d4ed05bd830d6bbbd7256) to configure keycloak. Also set USE_KEYCLOAK = True in env file to start enable Kyecloak
7. Now you should be able to login via keycloak service.

~### To Test MOIRA transfer~
We can test this along with [MOIRA migration to Keycloak PR](https://github.com/mitodl/odl-video-service/pull/1324)
1. I have added a management command that will create users from MOIRA list into keycloak and django and assign them group in keycloak
2. To test this add a MOIRA list in local setup
3. Run the management command `python manage.py migrate_moira_to_keycloak.py`
4. It should create group and users in keycloak and django as well with temporary passwords.

PR #1324, opened 2025-08-04
Moira lists to Keycloak groups

marslanabdulrauf:
https://github.com/mitodl/hq/issues/6821#issuecomment-3149755740

This PR:
1. Migrate Moira lists usage with Keycloak groups
2. Fix/update Management command to migrate moira lists to keycloak

### How can this be tested?

1. Checkout to this branch and run `docker compose up --build -d`
2. See if you have any `moira lists` or you can add one for testing
3. Run the management command to migrate all users from that moira list to keycloak with same name group `python manage.py migrate_moira_to_keycloak`
4. Go to [Keycloak](http://kc.odl.local:7080/admin/master/console/#/ovs-local/users) and verify if the users are created or not under the same group
5. Add/Update moira list from frontend in the collection and see if that works fine.
6. Try getting the group info from APIs
    - http://ovs.odl.local:8089/api/v0/moira/list/repalce-me-with-group-name
    - http://ovs.odl.local:8089/api/v0/groups/user/replace-me-with-user-email
7. Transcoding a video should also send email to users added in the collection `admin_lists`
8. Make sure the list contains all the users from that group/list


marslanabdulrauf:
> What's the status of this PR?

Its ready for review

config/keycloak/realms/ovs-local-realm.json
```diff
@@ -988,25 +988,99 @@
...
         "realm_client": "true",
         "post.logout.redirect.uris": "+"
       },
       "authenticationFlowBindingOverrides": {},
-      "fullScopeAllowed": false,
+      "fullScopeAllowed": true,
       "nodeReRegistrationTimeout": 0,
+      "protocolMappers": [
+        {
+          "id": "a1b2c3d4-e5f6-7890-abcd-ef1234567891",
```
mbertrand: Keycloak fails to start properly because of the dupe id here:

>  org.hibernate.NonUniqueObjectException: A different object with the same identifier value was already associated with this persistence context for entity [org.keycloak.models.jpa.entities.ProtocolMapperEntity with id 'a1b2c3d4-e5f6-7890-abcd-ef1234567891']

mbertrand (changes requested):
I ran into a roadblock while reviewing this, couldn't get keycloak to start properly.  Initial errors were caused by dupe ids in the realm configuration.  After making each of those unique, I ran into this error:

> keycloak-1 | 2025-12-02 16:45:52,919 ERROR [org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (main) Error details:: java.lang.RuntimeException: Script upload is disabled

Which keycloak version do you have installed as the "latest" in docker?  It could be that we are on different versions and mine is incompatible with the current keycloak/realm config (something similar happened with apisix for mit-learn: https://github.com/mitodl/mit-learn/pull/2626)

mbertrand:
I made a [new commit](https://github.com/mitodl/odl-video-service/pull/1324/changes/ef75cbc336f39ee632e97cefe68b63967cadd0e0) to split the one management command into 3:

- `migrate_moira_to_keycloak` - create Keycloak groups for all or a specified subset of `ui.models.KeycloakGroup` objects
- `migrate_users_to_keycloak` - upsert all or a specified subset of Django users to Keycloak
- `assign_group_users` - specify which existing keycloak users should be added to a single keycloak group (by username or email)

I haven't had a chance to try any of them out yet, they might need to be tweaked (or replaced altogether if there's a better way of handling this that I failed to think of).

@umar8hassan @marslanabdulrauf  I think this PR might also need to be rebased with the older keycloak PR here? https://github.com/mitodl/odl-video-service/pull/1289

arslanashraf7:
> I made a [new commit](https://github.com/mitodl/odl-video-service/pull/1324/changes/ef75cbc336f39ee632e97cefe68b63967cadd0e0) to split the one management command into 3:
> 
> * `migrate_moira_to_keycloak` - create Keycloak groups for all or a specified subset of `ui.models.KeycloakGroup` objects
> * `migrate_users_to_keycloak` - upsert all or a specified subset of Django users to Keycloak
> * `assign_group_users` - specify which existing keycloak users should be added to a single keycloak group (by username or email)
> 
> I haven't had a chance to try any of them out yet, they might need to be tweaked (or replaced altogether if there's a better way of handling this that I failed to think of).
> 
> @umar8hassan @marslanabdulrauf I think this PR might also need to be rebased with the older keycloak PR here? #1289

@umar8hassan could you take a look at this since Arslan is out this week?

ui/tasks_test.py
```diff
@@ -1,34 +1,26 @@
...
 
 from ui import tasks
-from ui.constants import VideoStatus
-from ui.encodings import EncodingNames
-from ui.factories import VideoFactory, VideoFileFactory
 
 
 @pytest.mark.django_db
-@factory.django.mute_signals(signals.post_save)
 def test_post_video_to_edx(mocker):
```
umar8hassan, line 10: The test was rewritten to use mocks instead of factories. It no longer exercises real DB behaviour. There should be a comment or separate PR note explaining why this was changed.

ui/tasks.py
```diff
@@ -103,3 +107,76 @@ def post_collection_videos_to_edx(video_ids):
...
+    return {
+        "created": 0,
+        "existing_skipped": 0,
+        "invalid_skipped": 0,
+        "failed": 0,
+        "errors": [],
+    }
+
+
+@shared_task(bind=True)
```
umar8hassan, line 122: `bind=True` with unused `self` is a bit misleading. We should either remove it or implement retries if needed.

ui/management/commands/migrate_moira_to_keycloak.py
```diff
@@ -0,0 +1,127 @@
...
+
+        async_results = [
+            migrate_keycloak_groups_chunk.delay(chunk, keycloak_config)
+            for chunk in chunks
+        ]
+
+        total_chunks = len(chunks)
+        for index, result in enumerate(async_results, start=1):
+            try:
+                payload = result.get(propagate=False)
```
umar8hassan, line 87: this makes the execution synchronous to some extent and might timeout if a leading task takes longer than the default timeout. is there a better way to handle this? maybe use `first in` result aggregation? or maybe remove `.delay()` in `migrate_keycloak_groups_chunk.delay(chunk, keycloak_config)` and execute synchronously.

umar8hassan (changes requested):
can we add tests for the commands?

PR #1340, opened 2025-09-03
fix: mp4 format added when re-syncing videos to edx

marslanabdulrauf:
https://github.mit.edu/mitxonline/mitxonline-issues/issues/567

This PR:
1. Adds all formats except ORIGINAL while re-syncing video files to edX endpoint

### How can this be tested?
1. Checkout to this branch and restart celery container
2. Make sure you have working edX setup
3. Create a new Collection or create new one and for now don't add and configure endpoint for this collection
4. Upload Videos
5. After successfully uploading videos configure endpoint for this collection and add course_id to sync with
6. Run `Sync Videos with edX` by clicking on this button top right -- You should see this after configuring the endpoint and course_id
7. Verify that you are getting both HLS and MP4 formats in LMS/CMS
8. You can verify this by navigation to this url: /admin/edxval/video/ and check the videos are there using client_id/video_name


umar8hassan:
I was able to fix my integration issue. The API URL in edX endpoint needed to be `http://host.docker.internal:8000/` instead of `http://localhost:8000/` because of isolation in OVS and edX setup.
cc @marslanabdulrauf

PR #1360, opened 2025-10-01
fix: avoid saving multiple video files for same format

marslanabdulrauf:
https://github.com/mitodl/hq/issues/3444

This PR:
1. Avoid creating duplicate Video files for same format

### How can this be tested?
1. Checkout to this branch and make sure your setup is fresh/up to date -- `docker-compose up [--build] -d`
2. Transcode any video; Go to OVS admin -> and transcoded video at `/admin/ui/video/`
3. Update the `S3_object_keys` and and change the name of the file of any format, either HLS or DESKTOP_MP4 ( for testing purpose only)
4. Now re-transcode the same video; you should not see two format for any of the encoding and `S3_object_key` should contain the latest transcoded file path

### Testing the management command
1. Checkout to `master`; Repeat the above ☝️ process and you will notice 2 VideoFiles for same format
2. Now run the management command `python manage.py remove_duplicate_encodings --all [--video-ids] [--collection-ids]`
3. **NOTE:** you can also add `--dry-run` to get the stats without deleting the actual content/data

### DIFF
--- a/cloudsync/api.py
+++ b/cloudsync/api.py
@@ -103,6 +103,32 @@ def process_hls_outputs(file_paths: list, video: Video) -> None:
             file_path = Path(file_path)
             bucket_name = file_path.parts[1]
             s3_path = str(Path(*file_path.parts[2:])).replace(RETRANSCODE_FOLDER, "")
+
+            # Check if VideoFile with HLS encoding already exists for this video
+            existing_hls_files = VideoFile.objects.filter(
+                video=video, encoding=EncodingNames.HLS
+            )
+
+            # If an existing HLS file has different s3_object_key, delete it from S3
+            for existing_file in existing_hls_files:
+                if existing_file.s3_object_key != s3_path:
+                    try:
+                        log.info(
+                            "Deleting existing HLS VideoFile with different s3_object_key",
+                            video_id=video.id,
+                            old_s3_key=existing_file.s3_object_key,
+                            new_s3_key=s3_path,
+                        )
+                        existing_file.delete_from_s3()
+                        existing_file.delete()
+                    except Exception as exc:
+                        log.error(
+                            "Failed to delete existing HLS VideoFile",
+                            video_id=video.id,
+                            s3_object_key=existing_file.s3_object_key,
+                            error=str(exc),
+                        )
+
             VideoFile.objects.update_or_create(
                 s3_object_key=s3_path,
                 defaults={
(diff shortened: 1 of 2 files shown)


cloudsync/api.py
```diff
@@ -103,6 +103,33 @@ def process_hls_outputs(file_paths: list, video: Video) -> None:
...
+            # Check if VideoFile with HLS encoding already exists for this video
+            existing_hls_files = VideoFile.objects.filter(
+                video=video, encoding=EncodingNames.HLS
+            )
+
+            # If an existing HLS file has different s3_object_key, delete it from S3
+            for existing_file in existing_hls_files:
+                if existing_file.s3_object_key != s3_path:
+                    try:
+                        log.info(
```
mbertrand, line 116: Maybe change to `log.debug`?

cloudsync/api.py
```diff
@@ -129,6 +156,32 @@ def process_mp4_outputs(outputs: list, video: Video) -> None:
...
+                            # odl-video-service/ui/signals.py
+                            existing_file.delete()
+                        except Exception as exc:
+                            log.error(
+                                "Failed to delete existing MP4 VideoFile",
+                                video_id=video.id,
+                                s3_object_key=existing_file.s3_object_key,
+                                error=str(exc),
+                            )
+
```
mbertrand, line 184: This code is repeated twice, might be better to move it into its own function:

```python
    def delete_existing_video_files(encoding: str):
            # Check if VideoFile with encoding already exists for this video
            existing_video_files = VideoFile.objects.filter(
                video=video, encoding=encoding
            )

            # If an existing HLS file has different s3_object_key, delete it from S3
            for existing_file in existing_video_files :
                if existing_file.s3_object_key != s3_path:
                    try:
                        log.debug(
                            "Deleting existing VideoFile with different s3_object_key",
                            video_id=video.id,
                            old_s3_key=existing_file.s3_object_key,
                            new_s3_key=s3_path,
                        )
                        # S3 object(s) will be deleted through pre-delete signal
                        # odl-video-service/ui/signals.py
                        existing_file.delete()
                    except Exception as exc:
                        log.error(
                            "Failed to delete existing VideoFile",
                            video_id=video.id,
                            s3_object_key=existing_file.s3_object_key,
                            error=str(exc),
                        )

            VideoFile.objects.update_or_create(
                s3_object_key=s3_path,
                defaults={
                    "video": video,
                    "bucket_name": bucket_name,
                    "encoding":encoding,
                    "preset_id": "",
                },
            )
```

mbertrand (changes requested):
Code looks good, a couple minor suggestions.  I wasn't able to test the functionality because my env variables (AWS, maybe certificates) seem to be out of date and I couldn't find OVS on kubernetes to retrieve them.

PR #1450, opened 2026-04-17
feat: add MOIRA_ENABLED toggle to gracefully disable Moira integration

blarghmatey:
N/A

The Moira SOAP service has been retired (mTLS authentication is no longer functional). This PR adds a `MOIRA_ENABLED` boolean environment variable (default `True`) so the integration can be disabled without breaking the running application.

**When `MOIRA_ENABLED=False`:**

- `MIT_WS_CERTIFICATE` and `MIT_WS_PRIVATE_KEY` are no longer required at startup, and the x509 cert/key files are not written to disk
- `get_moira_client()` raises `MoiraException` with a clear message rather than attempting a dead SOAP connection
- `user_moira_lists()` returns an empty set immediately — all list-based permission checks degrade gracefully to "no access via lists" instead of raising 500 errors
- `validate_moira_lists()` in serializers skips the SOAP existence check, allowing collections/videos with list names to be saved (names are preserved in the DB for a potential future re-enable)
- `_get_recipients_for_video()` in mail tasks skips the mailing-list SOAP check and falls back to owner-only email notification

**Nothing changes when `MOIRA_ENABLED=True` (the default).** The `MoiraList` model and all existing list assignments are untouched — no migration needed.

**To disable Moira in production**, set:
```
MOIRA_ENABLED=False
```
and remove (or leave unset) `MIT_WS_CERTIFICATE` and `MIT_WS_PRIVATE_KEY`.

Also includes:
- Fix for a pre-existing typo in `pytest.ini`: `MIT_WS_PRIVATRE_KEY` → `MIT_WS_PRIVATE_KEY`
- Regenerated DRF lint baseline (an added import line shifted existing violation line numbers by 1)

### How can this be tested?
The new `MOIRA_ENABLED=False` path is covered by 4 new tests (439 total, up from 435):
- `test_get_moira_client_disabled` — verifies `get_moira_client()` raises with a clear message
- `test_user_moira_lists_moira_disabled` — verifies empty set returned without calling Moira
- `test_collection_serializer_validation_skipped_when_moira_disabled` — verifies serializer saves without SOAP call
- `test_get_recipients_for_video_moira_disabled` — verifies owner-only email without SOAP call

To manually verify the toggle end-to-end, set `MOIRA_ENABLED=False` in the environment and confirm the app starts without Moira credentials and that authenticated users can still access public/owned content normally.


mbertrand (approved):
👍

PR #1458, opened 2026-04-24
feat: added resizing for thumbnail upload using PIL

umar8hassan:
https://github.com/mitodl/hq/issues/10960

Added resizing for uploaded thumbnails using PIL with max 640*360 dimensions before uploading them to S3. Also added a max cap of 1 MB for uploads

### How can this be tested?
1. Edit Video from either collection detail page or Video detail page
2. You should see previous thumbnail in the form (preview if available)
3. Upload new thumbnail in PNG/JPEG format with size larger than 10 MB.
4. Verify that images larger than 1 MB are not allowed. And verify from admin that image and its dimensions were not updated.
5. Now upload another image smaller than 1 MB. it should be uploaded with error.
6. verify that the thumbnail dimensions are also updated in admin.
**7. NOTE:** you might not get immediate new image in the same thumbnail url due to Cloudfront caching


cloudsync/api.py
```diff
@@ -565,61 +565,85 @@ def upload_subtitle_to_s3(caption_data, file_data):
...
-                # Return the original bytes to avoid a lossy re-encode.
+            # Normalise EXIF orientation so pixel data matches display orientation
+            # before any size computation or re-encode.
+            img = ImageOps.exif_transpose(img)
+            # Downscale if either dimension exceeds the configured limit.
+            orig_w, orig_h = img.size
+            if orig_w > max_width or orig_h > max_height:
+                img.thumbnail((max_width, max_height), Image.LANCZOS)
+            final_w, final_h = img.size
+            if img.format == "JPEG" and (orig_w, orig_h) == (final_w, final_h):
```
mbertrand: `img.format` will be `None` here after `img = ImageOps.exif_transpose(img)` is run.  So might be best to store the original format in a variable beforehand.

```python
img_format = img.format
if img_format not in ("JPEG", "PNG"):
    ...etc
img = ImageOps.exif_transpose(img)
..etc
if img_format == "JPEG" and (orig_w, orig_h) == (final_w, final_h):
```

static/js/components/dialogs/EditVideoFormDialog.js
```diff
@@ -27,6 +27,9 @@ import type { Video, VideoUiState } from "../../flow/videoTypes"
 import { calculateListPermissionValue } from "../../util/util"
 import { videoHasError, videoIsProcessing } from "../../lib/video"
 
+const THUMBNAIL_MAX_FILE_SIZE = 10 * 1024 * 1024 // 10 MB
```
mbertrand: This value should come from the backend instead of being hardcoded.

You could do it this way:

```typescript
const maxSize = SETTINGS.thumbnail_upload_max_size
```

and add this to `default_js_settings` view in `ui/views.py`:

```python
        "thumbnail_upload_max_size": settings.THUMBNAIL_UPLOAD_MAX_SIZE,
```

static/js/components/dialogs/EditVideoFormDialog.js
```diff
@@ -181,6 +184,15 @@ class EditVideoFormDialog extends React.Component<*, DialogState> {
...
       return
     }
+    if (file.size > THUMBNAIL_MAX_FILE_SIZE) {
+      this.setState({
+        thumbnailError:      `This image is too large (max ${THUMBNAIL_MAX_FILE_SIZE / (1024 * 1024)} MB). Please reduce the file size and try again.`,
+        thumbnailFile:       null,
+        thumbnailPreviewUrl: null
+      })
+      event.target.value = ""
+      return
```
mbertrand, line 201: If the value of THUMBNAIL_MAX_FILE_SIZE  is greater than the nginx `client_max_body_size` value, no error is shown when the upload fails with a 413 error.

When I upload an image within the size limits, I am seeing this error message for some reason even though the request succeeds: "Thumbnail upload failed (undefined)"

Image

mbertrand (changes requested):

odl_video/settings.py
```diff
@@ -361,6 +361,9 @@
 VIDEO_S3_TRANSCODE_BUCKET = get_string("VIDEO_S3_TRANSCODE_BUCKET", "")
 VIDEO_S3_THUMBNAIL_BUCKET = get_string("VIDEO_S3_THUMBNAIL_BUCKET", "")
 VIDEO_S3_SUBTITLE_BUCKET = get_string("VIDEO_S3_SUBTITLE_BUCKET", "")
+THUMBNAIL_UPLOAD_MAX_WIDTH = get_int("THUMBNAIL_UPLOAD_MAX_WIDTH", 640)
+THUMBNAIL_UPLOAD_MAX_HEIGHT = get_int("THUMBNAIL_UPLOAD_MAX_HEIGHT", 360)
+THUMBNAIL_UPLOAD_MAX_SIZE = get_int("THUMBNAIL_UPLOAD_MAX_SIZE", 10 * 1024 * 1024)
```
mbertrand: This value should be identical to whatever max is configured for nginx (locally it is 25 MB in `config/nginx.conf`, but on RC/production I think it is 1 MB or perhaps less).

umar8hassan: I agree. Having a different limit for local dev also creates inconsistency between the tested and actual scenario. I would update the local dev environment to use the default max limit as well.

cloudsync/api.py
```diff
@@ -565,61 +565,93 @@ def upload_subtitle_to_s3(caption_data, file_data):
...
-                    f"Unsupported image format: {img.format!r}. Only JPEG and PNG are supported."
+                    f"Unsupported image format: {img_format!r}. Only JPEG and PNG are supported."
                 )
-            if img.format == "JPEG":
-                # Return the original bytes to avoid a lossy re-encode.
+            # Normalise EXIF orientation so pixel data matches display orientation
+            # before any size computation or re-encode.
+            transposed = ImageOps.exif_transpose(img)
+            exif_changed = transposed is not img
+            img = transposed
```
mbertrand: I think this might fix the issue that sentry recently flagged:

```python
            orientation = img.getexif().get(ExifTags.Base.Orientation, 1)
            exif_changed = orientation != 1
            img = ImageOps.exif_transpose(img)
```

mbertrand (approved):
LGTM, just 1 potential suggestion to address the issue that sentry flagged today.

PR #1470, opened 2026-05-08
feat: replace video button added on collection list and video detail page

marslanabdulrauf:
https://github.com/mitodl/hq/issues/11116

This pull request introduces a new "Replace Video" feature that allows admins to replace an existing video with a new file from Dropbox. It includes backend API changes, permission and serializer additions, and frontend UI updates to enable video replacement from both the video list and detail views. The implementation ensures proper permissions, triggers reprocessing workflows, and handles cache invalidation so users see the updated video immediately.

**Backend changes:**

* **API and Permissions:**
  * Added `replace_video_from_dropbox` function in `ui/api.py` to update a video's source, reset its status, and start the reprocessing workflow (`stream_to_s3` + `retranscode_video`).
  * Introduced the `CanReplaceVideo` permission class to ensure only admins can replace videos in their collections.
  * Added `ReplaceVideoSerializer` for validating replace video API requests.

* **CloudFront Invalidation Improvements:**
  * Refactored CloudFront cache invalidation logic into `_invalidate_cloudfront_paths` and `_collect_output_keys` helpers, ensuring that all new/updated video and thumbnail files are properly invalidated after retranscoding or replacement. [[1]](diffhunk://#diff-1060724a69b1a3acf1698f94b51c3c1d9bcab78f02f2ac1174a6638d964af027R49-R105) [[2]](diffhunk://#diff-1060724a69b1a3acf1698f94b51c3c1d9bcab78f02f2ac1174a6638d964af027R148-R150) [[3]](diffhunk://#diff-1060724a69b1a3acf1698f94b51c3c1d9bcab78f02f2ac1174a6638d964af027L683-R746)

**Frontend changes:**

* **Video Replacement UI:**
  * Updated `VideoCard` and `VideoList` components to add a "Replace" menu item for admins, which opens a Dropbox chooser and triggers the replacement workflow. [[1]](diffhunk://#diff-b6eab2d58cf786116b5aaf9121895641ba1c523344521327532ef13548820219L21-R44) [[2]](diffhunk://#diff-b6eab2d58cf786116b5aaf9121895641ba1c523344521327532ef13548820219L52-R88) [[3]](diffhunk://#diff-704b9a2f7ef18466d16bf1099b6fcb9b767f8f5fa37dbf59aeee769c7bb44da6L20-R21) [[4]](diffhunk://#diff-704b9a2f7ef18466d16bf1099b6fcb9b767f8f5fa37dbf59aeee769c7bb44da6R49-R53)
  * Integrated similar "Replace" functionality in the `VideoDetailPage` with Dropbox chooser, and provided user feedback via toasts. [[1]](diffhunk://#diff-4b2735dd2abe9e041c70a557c1bd21d6de849fc4a21875d5033da9ae528fa219R230-R235) [[2]](diffhunk://#diff-4b2735dd2abe9e041c70a557c1bd21d6de849fc4a21875d5033da9ae528fa219R248-R258) [[3]](diffhunk://#diff-4b2735dd2abe9e041c70a557c1bd21d6de849fc4a21875d5033da9ae528fa219R143-R169)

* **API Integration:**
  * Added `replaceVideoFromDropbox` to `lib/api.js` to call the new backend endpoint.
  * Hooked up the replacement handlers in `CollectionDetailPage` and `VideoDetailPage` containers, including error handling and UI updates. [[1]](diffhunk://#diff-54a95a2217488f25763e7bb8e0614d0390c7d57348de2a57f48bb29f6d4b06b2R195-R219) [[2]](diffhunk://#diff-54a95a2217488f25763e7bb8e0614d0390c7d57348de2a57f48bb29f6d4b06b2R284) [[3]](diffhunk://#diff-4b2735dd2abe9e041c70a557c1bd21d6de849fc4a21875d5033da9ae528fa219R143-R169)

**Other improvements:**

* Minor refactoring and dependency imports to support the new Dropbox chooser and API calls. [[1]](diffhunk://#diff-b6eab2d58cf786116b5aaf9121895641ba1c523344521327532ef13548820219R10) [[2]](diffhunk://#diff-54a95a2217488f25763e7bb8e0614d0390c7d57348de2a57f48bb29f6d4b06b2R32) [[3]](diffhunk://#diff-4b2735dd2abe9e041c70a557c1bd21d6de849fc4a21875d5033da9ae528fa219R25-R32) [[4]](diffhunk://#diff-bbfa6639828646ec23f964590fbd28a20acbfaa8fee57e15bb34f473c9a18091R15) [[5]](diffhunk://#diff-88667797e9c16318c990a49002d5643208598f653a69e5dd1f59ba4fc05cf029L12-R12)

These changes collectively provide a robust, admin-only workflow for replacing videos, ensuring that both the backend and frontend are aligned and that users always see the latest content.

image

image

### How can this be tested?
1. Checkout to this branch and run `docker-compose up --build -d` to make sure docker images /containers are up to date
2. Go to collection detail page by clicking on any of your collection
3. Click on the 3-dot menu (each video card have it) and choose "Replace" to replace the video
4. You can also replace video on VideoDetail page (see SS)
5. After replacing the video should see new video in the player


ui/api.py
```diff
@@ -57,6 +58,30 @@ def process_dropbox_data(dropbox_upload_data):
...
+        dropbox_file (dict): A single validated DropboxFile dict with 'link', 'name', etc.
+
+    Returns:
+        dict: Basic info about the video being re-processed
+    """
+    video = get_object_or_404(models.Video, key=video_key)
+    with transaction.atomic():
+        video.source_url = dropbox_file["link"]
+        video.save(update_fields=["source_url"])
+        video.update_status(VideoStatus.CREATED)
```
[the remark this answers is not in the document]
mbertrand: I don't think this comment is correct, the functionality works as expected.

ui/permissions.py
```diff
@@ -153,3 +153,23 @@ def has_permission(self, request, view):
             ) from exc
         collection = Collection.objects.filter(key=collection_key)
         return len(collection) > 0 and has_admin_permission(collection.first(), request)
+
+
+class CanReplaceVideo(BasePermission):
```
mbertrand: I think the existing `HasVideoPermissions`  class can be used instead of creating this new class, it should `return has_admin_permission(obj.collection, request)`

static/js/components/VideoCard.js
```diff
@@ -18,79 +19,113 @@ type VideoCardProps = {
...
+  onReplaceVideo?: Function
 }
 
-const VideoCard = (props: VideoCardProps) => {
-  let videoDisplay
-  const hasError = videoHasError(props.video),
-    isProcessing = videoIsProcessing(props.video)
-  const headerClass = isProcessing || hasError ? "message" : "thumbnail",
-    videoUrl = makeVideoUrl(props.video.key)
+class VideoCard extends React.Component<VideoCardProps, void> {
```
mbertrand: Not sure about converting `VideoCard` from functional to class (to capture a ref for the hidden Dropbox trigger?)  — I think newer MIT OL frontends (e.g. mit-learn) are all hooks-based functional components. Could we keep this functional and either use `useRef`, or skip the hidden trigger entirely and render `DropboxChooser` directly as the menu item?  cc @ChristopherChudzicki or @daniellefrappier18

ui/api.py
```diff
@@ -57,6 +58,30 @@ def process_dropbox_data(dropbox_upload_data):
     return response_data
 
 
+def replace_video_from_dropbox(video_key, dropbox_file):
```
mbertrand, line 61: Could use a unit test for this function

ui/api.py
```diff
@@ -57,6 +58,30 @@ def process_dropbox_data(dropbox_upload_data):
...
+    """
+    Replace an existing video with a new file from Dropbox.
+    Updates the source URL, resets the video status to Created,
+    and kicks off the stream_to_s3 + transcode_from_s3 chain.
```
mbertrand: `and kicks off the stream_to_s3 + retranscode_video chain`

ui/api.py
```diff
@@ -57,6 +58,30 @@ def process_dropbox_data(dropbox_upload_data):
...
+
+    Args:
+        video_key (UUID): The key of the existing Video to replace
+        dropbox_file (dict): A single validated DropboxFile dict with 'link', 'name', etc.
+
+    Returns:
+        dict: Basic info about the video being re-processed
+    """
+    video = get_object_or_404(models.Video, key=video_key)
```
mbertrand, line 79: Should we guard against replacing a video that's currently being processed? If status is `UPLOADING` / `TRANSCODING` / `RETRANSCODING` / `RETRANSCODE_SCHEDULED`, kicking off another chain could collide with the in-flight one.

cloudsync/api.py
```diff
@@ -46,6 +46,63 @@
...
+            },
+        )
+    except Exception:
+        log.exception(
+            "CloudFront invalidation failed",
+            distribution_id=dist_id,
+            path_count=len(paths),
+        )
+
+
```
mbertrand, line 109: These new functions could use some tests.  Also I keep seeing the old video after replacement, maybe a delay in cloudfront cache invalidation (or my cloudfront keys are expired, will double check that).

mbertrand (changes requested):

static/js/components/VideoCard.js
daniellefrappier18: Agreed — no need for a class here. `useRef` works, but rendering `DropboxChooser` directly as the menu item is cleaner if its API allows it. I'll check and go with whichever avoids the hidden trigger.

marslanabdulrauf: Mostly our OVS frontend is based on class based components that's why AI must have taken this decision and I accepted it. ~I will revert it back to functional component and use `useRef` instead.~

EDIT: We are on `React15` and it does not have `useRef`

ui/api.py
marslanabdulrauf: Yes, the race is real.
If a second replace fires while any of those states is active, both chains run concurrently and whichever finishes last wins, potentially overwriting the newer replacement with the older one.

I will disable "Replace" option if the video is in `in-flight` state.

cloudsync/api.py
marslanabdulrauf: yes I have raised this [issue with DevOps](https://github.com/mitodl/hq/issues/11116#issuecomment-4406726736) the invalidation is not working.

cloudsync/api.py
```diff
@@ -46,6 +46,63 @@
...
+def _invalidate_cloudfront_paths(keys: list) -> None:
+    """Create a single CloudFront invalidation batch for a list of S3 object keys."""
+    dist_id = getattr(settings, "VIDEO_CLOUDFRONT_DIST", "")
+    if not dist_id or not keys:
+        return
+    paths = [f"/{k}" for k in keys]
+    try:
+        cf_client = boto3.client("cloudfront")
```
mbertrand, line 94: Two issues here that together explain why the cache invalidation hasn't been working in practice. The first is pre-existing (the thumbnail invalidation added earlier had the same bug); the second is new in this PR.

### 1. Wrong setting passed as `DistributionId`

`VIDEO_CLOUDFRONT_DIST` holds the lowercase CloudFront **subdomain** (e.g. `du3yhovcx8dht`) — that's how it's used at [settings.py:322](odl_video/settings.py#L322) and [models.py:612](ui/models.py#L612) to build `https://<dist>.cloudfront.net/...` URLs, and how Vault populates it (`{{ index .Secrets "cloudfront" "subdomain" }}` in the IaC repo). But CloudFront's `CreateInvalidation` API needs the **distribution ID**, which is a different value (I can DM you the RC value). Calling it with the subdomain returns `NoSuchDistribution` and the invalidation silently fails (the `except Exception` swallows it after logging).

Fix — introduce a separate setting for the distribution ID:

```suggestion
def _invalidate_cloudfront_paths(keys: list) -> None:
    """Create a single CloudFront invalidation batch for a list of S3 object keys."""
    dist_id = getattr(settings, "VIDEO_CLOUDFRONT_DISTRIBUTION_ID", "")
    if not dist_id or not keys:
        return
```

…plus add `VIDEO_CLOUDFRONT_DISTRIBUTION_ID = get_string("VIDEO_CLOUDFRONT_DISTRIBUTION_ID", "")` in `odl_video/settings.py` next to `VIDEO_CLOUDFRONT_DIST`, an entry in `.env.example`, and a corresponding Vault secret + template in `ol-infrastructure` (the existing `cloudfront` Vault path would just need a `distribution_id` key alongside `subdomain`).

### 2. Retranscode paths still carry the `retranscode/` prefix

Even with #1 fixed, the keys passed in here for retranscoded jobs come from `_collect_output_keys` reading `outputGroupDetails`, where MediaConvert wrote them under `retranscode/transcoded/...` ([transcode_video at line 341](cloudsync/api.py#L341)). `move_s3_objects` in `process_transcode_results` then moves them to `transcoded/...`, which is what users actually fetch — `process_hls_outputs`/`process_mp4_outputs` already strip `RETRANSCODE_FOLDER` for the same reason ([lines 171, 191](cloudsync/api.py#L171)). The CloudFront distribution serves these via two separate path-pattern behaviors (`/transcoded*` vs `/retranscode*`), so an invalidation against `/retranscode/transcoded/...` doesn't touch what the player is loading.

Fix — strip `RETRANSCODE_FOLDER` as keys are collected:

```suggestion
def _collect_output_keys(output_groups: list) -> list:
    """
    Extract S3 object keys from MediaConvert outputGroupDetails.

    For HLS groups a directory wildcard (``<dir>/*``) is appended so that
    individual ``.ts`` segment files — which are not listed in the job output —
    are also invalidated.

    Retranscode jobs write to ``retranscode/transcoded/...`` and are then
    moved to ``transcoded/...`` (the URL CloudFront actually serves), so we
    strip ``RETRANSCODE_FOLDER`` from each key before returning.
    """
    keys = []
    seen_wildcards = set()
    for group in output_groups:
        group_type = group.get("type", "")
        for path in group.get("playlistFilePaths", []):
            if key := _s3_uri_to_key(path):
                keys.append(key.replace(RETRANSCODE_FOLDER, ""))
        for output in group.get("outputDetails", []):
            for path in output.get("outputFilePaths", []):
                if key := _s3_uri_to_key(path):
                    key = key.replace(RETRANSCODE_FOLDER, "")
                    keys.append(key)
                    if "HLS_GROUP" in group_type:
                        directory = key.rsplit("/", 1)[0]
                        wildcard = f"{directory}/*"
                        if wildcard not in seen_wildcards:
                            keys.append(wildcard)
                            seen_wildcards.add(wildcard)
    return keys
```

Worth adding a test case with `retranscode/transcoded/...` URIs in the output groups to lock #2 in (current tests use bare `transcoded/...` paths so they don't catch it).

ui/api.py
```diff
@@ -57,6 +58,30 @@ def process_dropbox_data(dropbox_upload_data):
...
+    with transaction.atomic():
+        video.source_url = dropbox_file["link"]
+        video.save(update_fields=["source_url"])
+        video.update_status(VideoStatus.CREATED)
```
mbertrand, line 83: There's a separate issue here that surfaced in testing: if the new Dropbox file's extension differs from the original (e.g. original was `.mp4`, replacement is `.m4v`), the new bytes never replace the source on S3 and retranscoding silently re-runs against the stale file.

`Video.get_s3_key()` derives the extension from `source_url` ([ui/models.py:478](ui/models.py#L478)):

```python
_, extension = os.path.splitext(self.source_url.split("/")[-1])
newkey = "{uuid}/video{ext}".format(uuid=str(self.hexkey), ext=extension)
```

So during the chain:
1. `stream_to_s3` writes to `<uuid>/video.<new_ext>` (a different S3 key from the existing record).
2. `retranscode_video` does `video.videofile_set.get(encoding="original")` and gets the existing `VideoFile` still pointing at `<uuid>/video.<old_ext>`.
3. `transcode_video` uses that VideoFile as input — re-transcoding the stale source. The old object also lingers in S3 forever.

I confirmed this in QA: a test video had `original` at `…/video.mp4` (3 days old) but transcoded outputs from 11 minutes ago — the retranscode used the old source because the new bytes had landed at `…/video.m4v` and the VideoFile was never repointed.

Suggested fix — realign the existing original VideoFile's key before kicking off the chain, and clean up the old object:

```suggestion
def replace_video_from_dropbox(video_key, dropbox_file):
    """
    Replace an existing video with a new file from Dropbox.
    Updates the source URL, resets the video status, realigns the existing
    "original" VideoFile to the new source's S3 key (since `get_s3_key()` is
    derived from the source URL's extension), and kicks off the
    stream_to_s3 + retranscode_video chain.

    Args:
        video_key (UUID): The key of the existing Video to replace
        dropbox_file (dict): A single validated DropboxFile dict with 'link', 'name', etc.

    Returns:
        dict: Basic info about the video being re-processed
    """
    video = get_object_or_404(models.Video, key=video_key)
    with transaction.atomic():
        video.source_url = dropbox_file["link"]
        video.save(update_fields=["source_url"])

        # `get_s3_key()` derives the extension from source_url, so a replacement
        # whose extension differs (e.g. .mp4 → .m4v) would land at a new key
        # while the existing original VideoFile still pointed at the old one,
        # causing retranscode_video to re-transcode the stale source.
        new_key = video.get_s3_key()
        original_vf = video.videofile_set.filter(encoding="original").first()
        old_key = None
        if original_vf and original_vf.s3_object_key != new_key:
            old_key = original_vf.s3_object_key
            original_vf.s3_object_key = new_key
            original_vf.save(update_fields=["s3_object_key"])

        video.update_status(VideoStatus.CREATED)

    if old_key:
        models.delete_s3_objects.delay(original_vf.bucket_name, old_key)

    chain(tasks.stream_to_s3.s(video.id), tasks.retranscode_video.si(video.id)).delay()
    return {"key": video.hexkey, "title": video.title}
```

A test that exercises a replacement with a *different* extension would catch this regression — the existing test uses `.mp4` → `.mp4`, where the keys happen to coincide.

ui/api.py
```diff
@@ -57,6 +58,30 @@ def process_dropbox_data(dropbox_upload_data):
...
+
+    # Considering Retranscode here as that will handle replacing the existing video
+    chain(tasks.stream_to_s3.s(video.id), tasks.retranscode_video.si(video.id)).delay()
```
mbertrand, line 96: Not strictly part of the replace-video flow, but surfaced while testing this PR — the same retranscode chain creates new `VideoThumbnail` records via [process_mp4_outputs](../cloudsync/api.py#L197), so videos that have been retranscoded (or had user-uploaded thumbnails over time) end up with multiple `VideoThumbnail` rows. After replacement, refreshing the page shows a *random* thumbnail — sometimes the new one, sometimes a stale one.

### Root cause

`VideoThumbnail` has no `Meta.ordering`. Without an `ORDER BY`, PostgreSQL is free to return rows in any order — typically heap order, which gets perturbed every time `replace_thumbnail_in_s3` updates a row's `max_width`/`max_height`/`updated_at`. So:

- The frontend picks `video.videothumbnail_set[0].s3_object_key` ([static/js/lib/urls.js:14-18](../static/js/lib/urls.js#L14-L18)) — non-deterministic.
- The upload endpoint picks the thumbnail to overwrite via `video.videothumbnail_set.first()` ([ui/views.py:547](../ui/views.py#L547)) — also non-deterministic.

So uploads can hit a *different* row than the one the frontend renders, and even when they hit the same one, subsequent page loads may show a different row entirely. Browser cache + CloudFront aren't involved — this is the DB returning rows in shifting order.

### Fix

Add `Meta.ordering = ["-created_at"]` on `VideoThumbnail` so newest-first is guaranteed everywhere `videothumbnail_set` is iterated:

```python
# ui/models.py — inside class VideoThumbnail(VideoS3):
class Meta:
    # Newest first so consumers picking [0] (e.g. the frontend's
    # `video.videothumbnail_set[0].s3_object_key`) and `.first()` callers
    # like `upload_thumbnail` always see the most recently created record.
    ordering = ["-created_at"]
```

Plus a one-line migration:

```python
# ui/migrations/0042_alter_videothumbnail_options.py
from django.db import migrations


class Migration(migrations.Migration):
    dependencies = [("ui", "0041_collection_for_shorts")]
    operations = [
        migrations.AlterModelOptions(
            name="videothumbnail",
            options={"ordering": ["-created_at"]},
        ),
    ]
```

After this, `upload_thumbnail` and the serializer agree on which row is "current," and the frontend's `[0]` is deterministic. Stale rows still linger in the DB and S3 — worth a separate cleanup / "single canonical thumbnail" follow-up, but ordering alone resolves the user-visible flicker.

mbertrand, line 96: Not strictly part of the replace-video flow, but surfaced while testing this PR — videos that have been retranscoded (or had user-uploaded thumbnails over time) end up with multiple `VideoThumbnail` rows. After replacement, refreshing the page shows a *random* thumbnail — sometimes new, sometimes stale.

### Root cause

`VideoThumbnail` has no `Meta.ordering`. Without an `ORDER BY`, PostgreSQL is free to return rows in any order — typically heap order, which gets perturbed every time `replace_thumbnail_in_s3` updates a row's `max_width`/`max_height`/`updated_at`. So:

- The frontend picks `video.videothumbnail_set[0].s3_object_key` ([static/js/lib/urls.js:14-18](static/js/lib/urls.js#L14-L18)) — non-deterministic.
- The upload endpoint picks the thumbnail to overwrite via `video.videothumbnail_set.first()` ([ui/views.py:547](ui/views.py#L547)) — also non-deterministic.

Uploads can hit a *different* row than the one the frontend renders, and even when they hit the same one, subsequent page loads may show a different row entirely. Browser cache + CloudFront aren't involved — this is the DB returning rows in shifting order.

### Fix

Add `Meta.ordering = ["-created_at"]` on `VideoThumbnail` so newest-first is guaranteed everywhere `videothumbnail_set` is iterated:

```python
# ui/models.py — inside class VideoThumbnail(VideoS3):
class Meta:
    # Newest first so consumers picking [0] (e.g. the frontend's
    # `video.videothumbnail_set[0].s3_object_key`) and `.first()` callers
    # like `upload_thumbnail` always see the most recently created record.
    ordering = ["-created_at"]
```

Plus a one-line migration:

```python
# ui/migrations/0042_alter_videothumbnail_options.py
from django.db import migrations


class Migration(migrations.Migration):
    dependencies = [("ui", "0041_collection_for_shorts")]
    operations = [
        migrations.AlterModelOptions(
            name="videothumbnail",
            options={"ordering": ["-created_at"]},
        ),
    ]
```

After this, `upload_thumbnail` and the serializer agree on which row is "current," and the frontend's `[0]` is deterministic.

static/js/components/VideoCard.js
```diff
@@ -18,79 +19,115 @@ type VideoCardProps = {
...
-          key={props.video.key}
-          showMenu={props.showVideoMenu}
-          closeMenu={props.hideVideoMenu}
-          open={props.isMenuOpen}
-          menuItems={menuItems}
-        />
-      </div>
-    </Card>
-  )
 }
```
mbertrand, line 122: One potential way to keep `VideoCard` as a function component even without `useRef`:  a plain `let` binding in the function scope works as a ref cell here: each render creates a fresh binding, and the ref callback + the menu-item action both close over the same binding within that render. Refs commit before the menu can be clicked, so by the time `triggerReplaceDropbox` runs the binding has been assigned. No class, no instance state, no `this`-binding gymnastics.

```jsx
// @flow
/* global SETTINGS: false */
import React from "react"
import _ from "lodash"

import Menu from "./material/Menu"
import Card from "./material/Card"
import { makeVideoThumbnailUrl, makeVideoUrl } from "../lib/urls"
import { videoIsProcessing, videoHasError, saveToDropbox, videoIsInFlight } from "../lib/video"
import DropboxChooser from "react-dropbox-chooser"

import type { Video } from "../flow/videoTypes"

type VideoCardProps = {
  video: Video,
  isAdmin: boolean,
  isMenuOpen: boolean,
  showDeleteVideoDialog: Function,
  showEditVideoDialog: Function,
  showShareVideoDialog: Function,
  showVideoMenu: Function,
  hideVideoMenu: Function,
  onReplaceVideo?: Function
}

const VideoCard = (props: VideoCardProps) => {
  const { video, isAdmin, isMenuOpen, showShareVideoDialog,
    showEditVideoDialog, showDeleteVideoDialog,
    showVideoMenu, hideVideoMenu, onReplaceVideo } = props

  let dropboxTriggerEl: ?HTMLElement = null
  const triggerReplaceDropbox = () => {
    if (dropboxTriggerEl) {
      dropboxTriggerEl.click()
    }
  }

  let videoDisplay
  const hasError = videoHasError(video),
    isProcessing = videoIsProcessing(video)
  const headerClass = isProcessing || hasError ? "message" : "thumbnail",
    videoUrl = makeVideoUrl(video.key)

  if (isProcessing) {
    videoDisplay = (
      <div>
        <h5>In Progress...</h5>
        <p>You should receive an email when the video is ready.</p>
      </div>
    )
  } else if (hasError) {
    videoDisplay = (
      <div className="error">
        <h5>Upload failed.</h5>
        <p>
          Please{" "}
          <a href={`mailto:${SETTINGS.support_email_address}`}>contact</a> the
          ODL Video Services team.
        </p>
      </div>
    )
  } else {
    videoDisplay = (
      <a href={videoUrl}>
        <img src={makeVideoThumbnailUrl(video)} alt="" />
      </a>
    )
  }

  let menuItems = [{ label: "Share", action: showShareVideoDialog }]

  const inFlight = videoIsInFlight(video)

  if (isAdmin) {
    menuItems = _.concat(
      menuItems,
      { label: "Edit", action: showEditVideoDialog },
      { label: "Save To Dropbox", action: () => saveToDropbox(video) },
      ...(!inFlight && onReplaceVideo
        ? [{ label: "Replace", action: triggerReplaceDropbox }]
        : []),
      { label: "Delete", action: showDeleteVideoDialog }
    )
  }

  return (
    <Card className="video-card">
      <div className={headerClass}>{videoDisplay}</div>
      <div className="video-card-body">
        <h2 className="mdc-typography--subheading2">
          <a href={videoUrl} title={video.title}>
            {video.title}
          </a>
        </h2>
      </div>
      <div className="actions">
        <Menu
          key={video.key}
          showMenu={showVideoMenu}
          closeMenu={hideVideoMenu}
          open={isMenuOpen}
          menuItems={menuItems}
        />
        {isAdmin && onReplaceVideo && !inFlight && (
          <div style={{ display: "none" }}>
            <DropboxChooser
              appKey={SETTINGS.dropbox_key}
              success={files => onReplaceVideo(files[0])}
              linkType="direct"
              multiselect={false}
              extensions={["video"]}
            >
              <button ref={el => { dropboxTriggerEl = el }}>replace</button>
            </DropboxChooser>
          </div>
        )}
      </div>
    </Card>
  )
}

export default VideoCard
```

mbertrand (changes requested):

ui/api.py
marslanabdulrauf: When we `retranscode` a video we [exclude_thumbnails](https://github.com/mitodl/odl-video-service/blob/master/cloudsync/api.py#L276-L286)
I can’t recall for sure, but I think I have also seen multiple thumbnails associated with a single video before, so it’s probably worth supporting this case.

marslanabdulrauf: Duplicate https://github.com/mitodl/odl-video-service/pull/1470#discussion_r3219347891

mbertrand (approved):
👍

PR #1475, opened 2026-05-20
fix: don't auto public learn collection videos

marslanabdulrauf:
https://github.com/mitodl/hq/issues/11353

This pull request updates both the frontend and backend logic for handling the public visibility of videos, with a special focus on collections intended for MIT Learn. The changes ensure that videos in collections marked as "include_in_learn" are not automatically made public, requiring explicit action from editors. The frontend now provides users with a checkbox to set video visibility when appropriate, and the backend and signals have been updated to enforce the new rules. Comprehensive tests have been added to verify this behavior.

**Frontend changes for video visibility:**
- Added an `isPublic` state and a "Make this video publicly accessible" checkbox to `EditVideoFormDialog`, shown only if the parent collection is public and video permissions are not enabled. The checkbox state is included when saving the video. [[1]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbR8) [[2]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbL42-R44) [[3]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbL67-R70) [[4]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbR111) [[5]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbL223-R228) [[6]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbR264-R266) [[7]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbR283-R287) [[8]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbR390-R411) [[9]](diffhunk://#diff-2f2579574533c1805348421d1be1fbca700372bb1179388ae11b735e5c3e52fbR559-R563)

**Backend logic for public video propagation:**
- Updated `CollectionAdmin.save_model` so that when a collection's `is_public` is toggled on, videos are only auto-made public if `include_in_learn` is False; otherwise, editors must explicitly set each video's visibility. Turning off `is_public` still propagates to all videos.
- Updated the `set_video_is_public_from_collection` signal to prevent videos in "include_in_learn" collections from being auto-set to public on creation.

image

### How can this be tested?
1. Checkout to this branch, make sure the docker is refreshed
2. Create a public collection
3. if the collection have `include_in_learn = True` => Adding new video should not be marked as public automatically and if you change the `is_public` flag from admin it will not migration to its videos
4. if the collection have `include_in_learn = False` => adding new video should be marked as public automatically and if you change the `is_public` flag it will migrate to all of its videos.

### DIFF
--- a/static/js/components/dialogs/EditVideoFormDialog.js
+++ b/static/js/components/dialogs/EditVideoFormDialog.js
@@ -465,6 +465,7 @@ class EditVideoFormDialog extends React.Component<*, DialogState> {
               value={PERM_CHOICE_PUBLIC}
               selectedValue={editVideoForm.viewChoice}
               onChange={this.handleVideoViewPermClick}
+              disabled={defaultPerms}
               className="wideLabel"
             />
           )}
--- a/ui/admin.py
+++ b/ui/admin.py
@@ -84,7 +84,11 @@ class CollectionAdmin(admin.ModelAdmin):
     def save_model(self, request, obj, form, change):
         """Propagate is_public changes to all videos in the collection."""
         super().save_model(request, obj, form, change)
-        if change and "is_public" in form.changed_data:
+        if (
+            change
+            and "is_public" in form.changed_data
+            and not (obj.is_public and obj.include_in_learn)
+        ):
             obj.videos.update(is_public=obj.is_public, is_private=not obj.is_public)
 
 
--- a/ui/admin_test.py
+++ b/ui/admin_test.py
@@ -1,0 +1,95 @@
+"""Tests for UI admin customizations"""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+from django.contrib.admin.sites import AdminSite
+from django.test import RequestFactory
+
+from ui.admin import CollectionAdmin
+from ui.factories import CollectionFactory, VideoFactory
+from ui.models import Collection
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.fixture()
+def collection_admin():
+    """Return a CollectionAdmin instance bound to the default admin site."""
+    return CollectionAdmin(model=Collection, admin_site=AdminSite())
+
+
+@pytest.fixture()
+def mock_request():
+    """Return a minimal GET request for use in admin calls."""
+    return RequestFactory().get("/admin/")
+
+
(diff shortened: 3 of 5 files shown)


static/js/components/dialogs/EditVideoFormDialog.js
```diff
@@ -465,6 +465,7 @@ class EditVideoFormDialog extends React.Component<*, DialogState> {
               value={PERM_CHOICE_PUBLIC}
               selectedValue={editVideoForm.viewChoice}
               onChange={this.handleVideoViewPermClick}
+              disabled={defaultPerms}
```
mbertrand, line 468: I noticed something on this form that could  potentially trip people up - I can choose the "Publicly accessible" radio button, but this does not move the selected parent radio button from "Same as collection permissions (default)" to "Override collection permissions for this video" - and when I click "Save Changes" the change is not applied (because the wrong parent radio button is still selected).

All the other choices are disabled until  the parent radio button is changed, the "Publicly accessible" button should behave the same way.

Image

mbertrand (changes requested):
Overall looks great, just one UI improvement suggestion

static/js/components/dialogs/EditVideoFormDialog.js
marslanabdulrauf: oh, the 4 override permissions should be disabled by default until user clicks on "Override collection permissions for this video"
Let me check if that's not the case.

marslanabdulrauf: Here is the recording @mbertrand. Could you please build the frontend just to be sure.

Uploading Screen Recording 2026-05-21 at 11.54.18 PM.mov…

mbertrand: Working correctly for me now

mbertrand (approved):
👍

PR #1491, opened 2026-05-26
Retain existing django permissions for users not in keycloak groups

mbertrand:
Closes https://github.com/mitodl/hq/issues/11474

Updates the `assign_user_groups` social-auth pipeline step so that users who are not in the Keycloak `Admin` or `Staff` groups keep their existing Django `is_superuser` / `is_staff` flags instead of having them reset to `False` on each login.

### How can this be tested?
**On `master`:**
- Log in as `admin@ovs.local` — you will be staff/superuser.
- In an incognito tab, log in as `user@ovs.local` — you will be a regular user.
- In your admin tab, go to Django admin and set `user@ovs.local` as staff & superuser.
- In your incognito tab, go to Django admin — you should be able to access it.
- Close the incognito tab, open a new one, log in as `user@ovs.local` again. Try going to Django admin — your staff/superuser permissions will be gone.

**On this branch:**
- Repeat the above. On the last step, `user@ovs.local` should still have staff/superuser permissions.

### DIFF
--- a/odl_video/pipeline.py
+++ b/odl_video/pipeline.py
@@ -36,7 +36,7 @@ def assign_user_groups(strategy, details, backend, user=None, *args, **kwargs):
             extra_groups = extra_data["user_groups"]
             groups.extend(extra_groups)
 
-    logger.info(f"Groups found for user {user.username}: {groups}")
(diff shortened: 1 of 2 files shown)


marslanabdulrauf (approved):
LGTM 👍

PR #1497, opened 2026-06-07
Authenticated Dropbox API downloads, permanent URLs

mbertrand:
Partially addresses mitodl/hq#11665

Related infrastructure PR: https://github.com/mitodl/ol-infrastructure/pull/4736

Dropbox throttles anonymous shared-link downloads, which stalls and times out `stream_to_s3` when several videos are uploaded at once. This switches the server-side download to Dropbox's authenticated `sharing/get_shared_link_file` endpoint, using a service-account refresh token, so transfers ride our Enterprise throughput instead of the anonymous-hotlink path. The Dropbox Chooser is also switched to durable `preview` links so the source URLs don't expire.

Requires three settings — `DROPBOX_KEY`, `DROPBOX_SECRET`, and `DROPBOX_REFRESH_TOKEN`. See the updated README for how to obtain them, or ping me on Slack for the values.

### How can this be tested?
Set the Dropbox settings (see README, or ask me for working values), then select one or more videos to upload via the frontend. They should download into S3 and transcode as usual.


cloudsync/tasks.py
```diff
@@ -368,22 +365,18 @@ def parse_content_metadata(response):
...
-    content_length = response.headers["Content-Length"]
-    if content_length:
-        content_length = int(content_length)
 
+    For authenticated Dropbox downloads the metadata arrives in the
+    ``Dropbox-API-Result`` header with a generic octet-stream content type, so
+    the MIME type is derived from the file name.
+    """
+    metadata = json.loads(response.headers["Dropbox-API-Result"])
+    file_name = metadata["name"]
```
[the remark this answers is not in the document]
mbertrand: That is the point - if it's not there something is wrong.

cloudsync/tasks.py
```diff
@@ -107,9 +105,8 @@ def stream_to_s3(self, video_id):
 
     task_id = self.get_task_id()
     try:
-        response = requests.get(video.source_url, stream=True, timeout=60)
-        response.raise_for_status()
-    except requests.HTTPError:
+        response = dropbox_api.stream_shared_link(video.source_url)
+    except (requests.HTTPError, dropbox_api.DropboxAuthError):
```
mbertrand: Fixed in e5f2345 — `stream_to_s3` now catches `dropbox_api.DropboxAuthError` alongside `requests.HTTPError`, so an auth failure marks the video `UPLOAD_FAILED` instead of leaving it stuck in `UPLOADING`. Added `test_upload_auth_failure` as a regression test.

cloudsync/dropbox_api.py
```diff
@@ -0,0 +1,82 @@
...
+        access_token = payload["access_token"]
+    except (ValueError, KeyError) as exc:
+        raise DropboxAuthError(
+            f"Dropbox token response missing access_token: {resp.text}"
+        ) from exc
+    _token_cache["access_token"] = access_token
+    _token_cache["expires_at"] = (
+        now + payload.get("expires_in", 14400) - _TOKEN_EXPIRY_BUFFER_SECONDS
+    )
+    return access_token
```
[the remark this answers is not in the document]
mbertrand: The cache is only used by the stream_to_s3 Celery task, and the worker runs the default prefork pool (celery -A odl_video worker -B, no --pool). Concurrent uploads run in separate processes, each with its own copy of this module-level _token_cache, and each prefork process handles tasks serially (single-threaded). So:

A threading.Lock only guards threads within a single process — it can't coordinate the separate processes that actually run concurrent uploads, so it wouldn't prevent simultaneous token refreshes here.
There's no in-process thread concurrency to race on the cache in the first place.
Worst case is each process refreshing its own token once — independent, valid, and harmless (bounded by worker count).

cloudsync/dropbox_api.py
```diff
@@ -0,0 +1,82 @@
...
+        SHARED_LINK_FILE_URL,
+        headers={
+            "Authorization": f"Bearer {get_access_token()}",
+            "Dropbox-API-Arg": json.dumps({"url": url}),
+        },
+        stream=True,
+        timeout=_DOWNLOAD_TIMEOUT,
+    )
+    response.raise_for_status()
+    return response
```
marslanabdulrauf: If we get 401, for any reason, may be we should give it another try after refreshing the access token.

cloudsync/dropbox_api.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def _require(name):
+    """Return a required Dropbox setting or raise if it is unset."""
+    value = getattr(settings, name, "")
+    if not value:
+        raise ImproperlyConfigured(
+            f"{name} must be set for authenticated Dropbox downloads"
+        )
+    return value
```
marslanabdulrauf: we have a list of [MANDATORY_SETTINGS](https://github.com/mitodl/odl-video-service/blob/master/odl_video/settings.py#L403) in settings, we can add these values there and no need to raise here.

cloudsync/dropbox_api.py
mbertrand: Added

marslanabdulrauf (approved):
A minor comment, every thing else looks good, I tested it out locally as well and it worked just fine.

PR #1517, opened 2026-06-15
Add retries to stream_to_s3 upload task

mbertrand:
N/A

Makes the Dropbox→S3 upload (`cloudsync.tasks.stream_to_s3`) resilient to transient failures and worker loss:

- **`acks_late=True` + `reject_on_worker_lost=True`** — an in-flight upload now survives a worker crash/redeploy (e.g. OOM-killed child) via broker redelivery, instead of silently dying in `UPLOADING` until the janitor reaps it.
- **Per-video upload lock (Redis)** — because `acks_late` redelivery can hand the same upload to a second worker while the first is still streaming, `stream_to_s3` takes a non-blocking per-`video_id` lock; a worker that can't acquire it gives up, since another worker already owns the upload. The lock is a short lease (`CLOUDSYNC_STREAM_S3_LOCK_TTL`, default 120s) refreshed by the progress callback as a heartbeat, so it frees itself shortly after a worker dies. This is what lets `visibility_timeout` be short without risking concurrent double-uploads to the same S3 key.
- **Retries (2 max, jittered exponential backoff)** on *transient* errors only: connection/timeout/chunked-encoding errors, HTTP **5xx/429/408**, and S3 `ClientError`/`BotoCoreError`. Permanent failures (HTTP 4xx like a revoked/missing link, malformed metadata, auth errors) still fail fast. The video is marked `UPLOAD_FAILED` only after retries are exhausted.
- **Janitor reconciliation** — the upload progress callback now throttle-refreshes `updated_at` (default 60s), so `fail_stuck_uploading_videos` acts as an *idle* timeout and no longer reaps healthy long-running uploads. This decouples it from the broker `visibility_timeout`.

New settings (all env-overridable): `CLOUDSYNC_STREAM_S3_MAX_RETRIES`, `CLOUDSYNC_STREAM_S3_RETRY_BACKOFF`, `CLOUDSYNC_STREAM_S3_RETRY_MAX_BACKOFF`, `CLOUDSYNC_UPLOAD_PROGRESS_REFRESH_SECONDS`, `CLOUDSYNC_STREAM_S3_LOCK_TTL`, `CELERY_BROKER_VISIBILITY_TIMEOUT`.

### How can this be tested?
1. Upload a video via the UI — confirm the video upload completes & transcodes normally.
2. **Retry path:** temporarily make `dropbox_api.stream_shared_link` raise `requests.ConnectionError`. Trigger an upload and confirm the worker logs `"Retrying stream_to_s3 after transient error"` with a jittered countdown, retries twice, then logs `"stream_to_s3 retries exhausted"` and the video flips to `UPLOAD_FAILED`.
3. **Immediate fail path:** temporarily make `dropbox_api.stream_shared_link` raise a `ValueError` → confirm it fails immediately with no retry.
4. **Worker-loss path:** set `CELERY_BROKER_VISIBILITY_TIMEOUT=300`, keeping the default `CLOUDSYNC_STREAM_S3_LOCK_TTL=120` (it must stay below the visibility timeout). Start a large upload, hard-kill the celery worker mid-stream (SIGKILL, so it can't release its lock), then restart it. The dead worker's lock auto-expires after ~120s; the broker re-queues the task at ~5 minutes; confirm the redelivered task re-acquires the lock and restarts the upload (status returns to `UPLOADING`, progress climbs from 0). Recovery requires lock-TTL expiry before redelivery — if the TTL were ≥ the visibility timeout, the redelivered task would find the stale lock still held, skip, and leave recovery to the idle janitor.

### Additional Context
- `visibility_timeout` defaults to 1h and is broker-wide (delays redelivery of *any* genuinely-dead task). It no longer needs to exceed the longest upload — the per-video lock prevents a redelivered copy from racing a still-running one — so it's free to be short for faster crash recovery. The only floor is the longest task ETA/countdown (the `stream_to_s3` retry backoffs, max ~10m); a countdown longer than this would be duplicated. Tune `CELERY_BROKER_VISIBILITY_TIMEOUT` as needed.
- `reject_on_worker_lost` redelivery is not bounded by `max_retries`; a file that reliably OOMs its worker could redeliver repeatedly (rate-capped by `visibility_timeout`, and the idle janitor still fails it).

### DIFF
--- a/cloudsync/tasks.py
+++ b/cloudsync/tasks.py
@@ -23,6 +23,7 @@ from celery.utils.time import get_exponential_backoff_interval
 from django.conf import settings
 from django.db import connection
 from googleapiclient.errors import HttpError
+from redis.exceptions import LockError
 from urllib3.exceptions import ProtocolError as Urllib3ProtocolError
 from urllib3.exceptions import TimeoutError as Urllib3TimeoutError
 
@@ -181,6 +182,20 @@ def fail_stuck_uploading_videos():
             log.exception("Failed to update stuck video status", video_id=video.id)
 
 
+def _video_upload_lock(app, video_id):
+    """
+    Non-blocking redis lock serializing concurrent uploads of one video.
+
+    thread_local=False: acquired on the task thread but reacquired/released from
+    s3transfer callback threads.
+    """
+    return app.backend.client.lock(
+        f"stream_to_s3:lock:{video_id}",
+        timeout=settings.CLOUDSYNC_STREAM_S3_LOCK_TTL,
+        thread_local=False,
+    )
+
+
 @shared_task(
     bind=True,
     base=VideoTask,
@@ -200,6 +215,15 @@ def stream_to_s3(self, video_id):
     except (Video.DoesNotExist, Video.MultipleObjectsReturned):
         log.error("Exception retrieving video", video_id=video_id)
         raise
+
+    lock = _video_upload_lock(self.app, video_id)
+    if not lock.acquire(blocking=False):
+        log.debug(
+            "stream_to_s3 skipped; another worker holds the upload lock",
+            video_id=video_id,
+        )
+        return False
+
     video.update_status(VideoStatus.UPLOADING)
 
     task_id = self.get_task_id()
@@ -245,6 +269,15 @@ def stream_to_s3(self, video_id):
                 meta={"uploaded": uploaded, "total": content_length},
             )
             if should_refresh:
+                # Heartbeat the lease so a short TTL survives a long upload.
+                # Best-effort: if already lost, another worker owns it now.
+                try:
+                    lock.reacquire()
+                except LockError:
+                    log.warning(
+                        "stream_to_s3 lost upload lock during heartbeat",
+                        video_id=video.id,
+                    )
                 # Keep updated_at fresh so the fail_stuck_uploading_videos janitor
                 # treats an actively streaming upload as alive rather than stuck.
                 # Throttled (above) to avoid a DB write on every chunk callback.
@@ -292,6 +325,14 @@ def stream_to_s3(self, video_id):
         # successful-upload path where nothing else closes it.
         if response is not None:
             response.close()
+        # Release the upload lock.
+        try:
+            lock.release()
+        except LockError:
+            log.warning(
+                "stream_to_s3 upload lock already released or expired",
+                video_id=video_id,
+            )
 
 
 @shared_task(bind=True, base=VideoTask)
--- a/cloudsync/tasks.py
+++ b/cloudsync/tasks.py
@@ -8,9 +8,11 @@ import re
 from datetime import timedelta
 
 import boto3
+import requests
 from boto3.s3.transfer import TransferConfig
-from botocore.exceptions import ClientError
+from botocore.exceptions import BotoCoreError, ClientError
 from celery import Task, group, shared_task, states
+from celery.utils.time import get_exponential_backoff_interval
 from django.conf import settings
 from googleapiclient.errors import HttpError
 
@@ -25,6 +27,24 @@ from ui.models import Collection, EncodeJob, Video, VideoSubtitle, YouTubeVideo
 from ui.utils import get_bucket, now_in_utc
 
 log = structlog.get_logger(__name__)
+
+
+def _should_retry_upload(exc):
+    """
+    True for transient upload errors worth retrying. Excludes permanent failures
+    like malformed Dropbox metadata (KeyError/ValueError), auth errors, and HTTP
+    4xx responses (e.g. a revoked or missing shared link), which won't fix
+    themselves on retry.
+    """
+    if isinstance(exc, (BotoCoreError, ClientError)):
+        return True
+    if isinstance(exc, requests.HTTPError):
+        # Only server-side (5xx) and throttling/timeout statuses are transient.
+        status = getattr(exc.response, "status_code", None)
+        return status is not None and (status >= 500 or status in (408, 429))
+    # Connection/timeout/chunked-encoding level errors are transient. Checked
+    # after HTTPError since HTTPError is also a RequestException.
+    return isinstance(exc, requests.exceptions.RequestException)
 
 
 class VideoTask(Task):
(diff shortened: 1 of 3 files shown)


cloudsync/tasks.py
```diff
@@ -158,15 +299,40 @@ def callback(bytes_uploaded):
...
+                full_jitter=True,
+            )
+            log.warning(
+                "Retrying stream_to_s3 after transient error",
+                video_id=video_id,
+                attempt=self.request.retries + 1,
+                countdown=countdown,
+                error=str(exc),
+            )
+            raise self.retry(exc=exc, countdown=countdown)
```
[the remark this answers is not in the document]
mbertrand: RETRY_MAX_BACKOFF (600s) is far below STUCK_UPLOADING_THRESHOLD_HOURS (2h). If someone cranked the max backoff into the hours or dropped the janitor threshold to minutes, the concern would become real. With defaults there's a 40× cushion, so the proposed fix is unnecessary complexity.

cloudsync/tasks.py
```diff
@@ -167,6 +327,14 @@ def callback(bytes_uploaded):
         # successful-upload path where nothing else closes it.
         if response is not None:
             response.close()
+        # Release the upload lock.
+        try:
+            lock.release()
+        except LockError:
+            log.warning(
+                "stream_to_s3 upload lock already released or expired",
+                video_id=video_id,
```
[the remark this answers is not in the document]
mbertrand: The lock here is short (TTL 120s, reacquired every 60s during upload), not a lifecycle-long lock — it's deliberately short so a crashed worker's lock expires fast.

That means holding it through the retry isn't effective: nothing heartbeats during the countdown, so a 120s TTL expires well before the up-to-600s backoff regardless of the finally release. The fix would only cover the first 120s of the window.

Concurrent uploads are already prevented by the non-blocking acquire guard at the top of the task — any redelivered/duplicate copy (including two copies of a retry, the real Redis + acks_late countdown case) finds the lock held, suppresses the chain, and gives up. Redis locks are atomic, so at most one task uploads at a time. Releasing between attempts is correct since no upload is running during the countdown.

The default config is also safe: visibility_timeout defaults to 3600s, well above the 600s max backoff.

cloudsync/tasks.py
```diff
@@ -5,14 +5,27 @@
...
+import requests
 from boto3.s3.transfer import TransferConfig
-from botocore.exceptions import ClientError
+from botocore.exceptions import (
+    BotoCoreError,
+    ClientError,
+    ConnectionClosedError,
+    HTTPClientError,
+)
+from botocore.exceptions import ConnectionError as BotoConnectionError
```
marslanabdulrauf:
```suggestion
from botocore.exceptions import (
    BotoCoreError,
    ClientError,
    ConnectionError as BotoConnectionError,
    ConnectionClosedError,
    HTTPClientError,
)
```

cloudsync/tasks.py
```diff
@@ -5,14 +5,27 @@
...
 from celery import Task, group, shared_task, states
+from celery.utils.time import get_exponential_backoff_interval
 from django.conf import settings
+from django.db import connection
 from googleapiclient.errors import HttpError
+from redis.exceptions import LockError
+from urllib3.exceptions import ProtocolError as Urllib3ProtocolError
+from urllib3.exceptions import TimeoutError as Urllib3TimeoutError
```
marslanabdulrauf:
```suggestion
from urllib3.exceptions import (
    ProtocolError as Urllib3ProtocolError,
    TimeoutError as Urllib3TimeoutError,
)
```

cloudsync/tasks.py
```diff
@@ -27,6 +40,65 @@
...
+TRANSIENT_STREAM_ERRORS = (
+    ConnectionError,  # builtin: ConnectionReset/BrokenPipe/etc.
+    TimeoutError,  # builtin
+    Urllib3ProtocolError,
+    Urllib3TimeoutError,
+)
+
+UPLOAD_PROGRESS_REFRESH_SECONDS = 60
+# TTL must exceed the refresh interval; doubled for headroom
+UPLOAD_LOCK_TTL = max(120, UPLOAD_PROGRESS_REFRESH_SECONDS * 2)
```
marslanabdulrauf: PR description mentioned these settings as configurable, could we add entries for these in the settings and update accordingly ?

cloudsync/tasks.py
```diff
@@ -137,18 +243,54 @@ def stream_to_s3(self, video_id):
...
+                    last_progress_refresh is None
+                    or (now - last_progress_refresh).total_seconds()
+                    >= UPLOAD_PROGRESS_REFRESH_SECONDS
+                )
+                if should_refresh:
+                    last_progress_refresh = now
+            self.update_state(
+                task_id=task_id,
+                state="PROGRESS",
+                meta={"uploaded": uploaded, "total": content_length},
```
marslanabdulrauf: `uploaded` is not being used anywhere else, may be we can use `total_bytes_uploaded` here as well ?

marslanabdulrauf (changes requested):
Some minor changes and my AI is giving me this scenario, not sure how to produce this:
1. Worker A starts upload and acquires the Redis lock.
2. During a long upload, lock heartbeat fails once or lock TTL expires (network hiccup, Redis blip, callback delay)
3. Worker B gets the same task (retry/redelivery/manual requeue) and acquires the now-free lock.
4. Now A and B can both stream to the same S3 key and both continue downstream logic.

Possible fix:
- Heartbeat failure only logs and continues, which allows two workers to upload concurrently after lock expiry.
- If callback treats lock-loss as a hard failure, Worker A exits instead of continuing after ownership is gone.
- Retry/redelivery then re-enters normal lock acquisition, so only one worker proceeds.

cloudsync/tasks.py
mbertrand: I initially made them configurable, but then decided against it because certain values need to be lower than others.  I will make them configurable again but with documentation about that and raise ImproperlyConfigured errors if they are not.

mbertrand:
@marslanabdulrauf it should handle the AI scenario now and I also made one last commit to respect "Retry-After" values for 429 exceptions.

marslanabdulrauf (approved):
LGMT

PR #1531, opened 2026-07-01
Add opt-in django-aqueduct settings module

blarghmatey:
## Summary

Adds a parallel, **opt-in** settings entry point based on [django-aqueduct](https://github.com/mitodl/django-aqueduct) **0.9.0** (codegen v2). `odl_video/settings.py` and everything under the existing settings/envs path is untouched — nothing changes for anyone not explicitly opting in via `DJANGO_SETTINGS_MODULE`.

- `odl_video/aqueduct_settings.py` — a Pydantic `BaseSettings` model (`AqueductSettings`) **generated by codegen v2's static AST discovery** of `odl_video/settings.py` plus the mitol EnvParser registry. The generator never imports the settings module, so generation is deterministic and secret-safe. The file is split into managed regions:
  - `# >>> aqueduct:generated:*` regions are rewritten verbatim by `manage.py generate_aqueduct_settings` and guarded by `--check` drift detection. These include `container_decoders` (`NoDecode` + `field_validator` handlers that parse list/dict settings from comma-separated / Python-literal / JSON / `[*]` env strings).
  - Hand-written refinements (required fields, the cross-field timing-window checks, the Keycloak URL derivation, the conditional `INSTALLED_APPS`/`MIDDLEWARE`/`AUTHENTICATION_BACKENDS`/`CELERY_BEAT_SCHEDULE` branches, and the derivations) live in the `# >>> aqueduct:preserved:*` region and survive regeneration.
- `DevAqueductSettings` layers a Vault source configured entirely from `VAULT_*` env vars (`django_aqueduct.sources.dev`); absent `VAULT_ADDR` it degrades to plain env settings, and misconfiguration raises an actionable error, not a `KeyError`.
- `odl_video/settings_aqueduct.py` / `odl_video/settings_aqueduct_dev.py` — thin shims that call `configure_django_settings(...)` using the `pre_configure` hook to initialize Sentry from the validated model before Django settings exist.
- `[tool.aqueduct]` in `pyproject.toml` pins the generation and parity configuration; `AGENTS.md` documents regeneration, drift, and parity commands.

### Codegen v2 recovered what v1 lost

Static discovery reads the settings *source*, so it recovers env-var aliases the old snapshot never honoured (`ODL_VIDEO_EMAIL_*`, `ODL_VIDEO_SECURE_SSL_REDIRECT`, `STATIC_CLOUDFRONT_DIST`, …), required-ness, the inline `get_int` reads in `CELERY_BEAT_SCHEDULE`, and the conditional `INSTALLED_APPS`/`MIDDLEWARE` branches — all of which v1's live-value introspection had frozen.

### Review-flagged bugs fixed

- **`_derive_databases` sslmode/OPTIONS bug** — applied `sslmode: require` unconditionally (a `TypeError` under the SQLite fallback) and discarded `OPTIONS` parsed from `DATABASE_URL`. Replaced with `django_aqueduct.derivations.database_config`, which is SQLite-safe and merges `OPTIONS` by design.
- **`CLOUDFRONT_DIST` alias** — documented its `STATIC_CLOUDFRONT_DIST` env var only in a comment. Static discovery now derives that `validation_alias` directly from the `get_string` call, so it can't drift.
- **Dev-only `update-statuses` beat task** — was baked into the static schedule and would have fired every 10s in production. It's conditional on `ENVIRONMENT=dev` again (and the `FEATURE`-gated `schedule_retranscodes` entry is restored).
- **Conditional `webpack_loader`** — the static `INSTALLED_APPS` omitted the conditionally-added `webpack_loader`, breaking bundle serving in prod configs. All four conditional app/middleware/auth branches are reproduced in the validator.

### Follow-up review fixes

- **`CELERY_REDBEAT_REDIS_URL` now respects an explicit value.** It was read from the environment and then unconditionally overwritten with `CELERY_BROKER_URL` in the validator, silently discarding an explicitly-set value (e.g. a distinct TLS Redis URL). It now derives via `first_url(CELERY_REDBEAT_REDIS_URL, CELERY_BROKER_URL)` — explicit wins, broker is the fallback when unset (unchanged from legacy, so parity holds).
- **Restored the `validation_alias` on the `SOCIAL_AUTH_KEYCLOAK_ACCESS_TOKEN_URL` override.** It still loaded from the env via field-name fallback, but the explicit alias matches the generated field and makes the Keycloak server/realm derivation unambiguously read the real env value.
- Removed a literal fallback-secret string from a code comment (GitGuardian).

### Boilerplate replaced by reusable derivations

`database_config` (DATABASES), `first_url` + `redis_cache` (the Redis/Celery URL chain and CACHES mega-validator), `admins_from_csv` (ADMINS), and `feature_flags` (the `FEATURE_*` scan). Feature flags flow through the source chain via a small env source that injects `FEATURE_*` keys as model extras with legacy `get_any` coercion, so **Vault-supplied flags are seen too** — the old `os.environ` scan bypassed every non-env source.

### List/dict env parsing (0.7.1+)

The generator emits `Annotated[..., NoDecode]` + a per-field before-validator for every list/dict field, parsing comma-separated / Python-literal / JSON / bare-`[*]` env strings before pydantic's default `json.loads` decoding can raise. This replaced an earlier custom env-source subclass.

### 0.9.0 clean-ups (this bump)

The prior revision carried three local workarounds for upstream gaps; 0.9.0 (and 0.8.1) fix all three, so they are removed:

- **`*_URL` fields are kept as `str`; `enrich_url_types` is left OFF.** 0.9.0's opt-in AnyUrl promotion serializes top-level fields back to `str`, but a promoted field still holds an `AnyUrl` *object* at runtime, which breaks two consumers here (verified): the in-validator Redis/Celery fallback calls `.strip()` on `CELERY_BROKER_URL` (`AnyUrl` has no `.strip()` → `AttributeError` at instantiation when it is set), and that value is nested into the `CACHES` dict where the field serializer never reaches, leaving an `AnyUrl` where redis expects a `str`. Plain `str` is the correct type for a settings-injection model.
- **Parity subset-compares dicts.** `DATABASES` and `DEFAULT_DATABASE_CONFIG` are dropped from `parity_ignore` (9 → 7 ignores) — 0.9.0 tolerates Django-injected dict keys (`ATOMIC_REQUESTS`, `AUTOCOMMIT`, …), and `database_config` is SQLite-safe, so they compare clean.
- **Generated file is now linted.** 0.8.1 made output `ruff format`-stable and dropped the blanket `# ruff: noqa`, so the full `extend-exclude` is replaced by a narrower `[tool.ruff.format] exclude` — lint now covers the file, and only *formatting* is skipped. One field still defeats format-stability: `CELERY_BEAT_SCHEDULE`'s default is a verbatim multi-line dict expression (inline `get_int` reads) that ruff-format re-nests, which would trip the drift gate since it lives in a generated region.

### Security behavior retained

`KEYCLOAK_SVC_ADMIN_PASSWORD` remains a *required* field (no default) — the legacy module falls back to a hardcoded `"odl-video-secret-2025"`, which this model deliberately does not replicate.

## Dependency note

Bumps `django-aqueduct[vault,mitol,derivations]` to `>=0.9.0` (published to PyPI; resolves from the index — no local path override).

## Test plan

- [x] `uv lock` / `uv sync` resolve `django-aqueduct==0.9.0` cleanly
- [x] `manage.py check` passes under `odl_video.settings_aqueduct` (prod shim) with fake env vars
- [x] `manage.py check` passes under `odl_video.settings_aqueduct_dev` (dev shim) with no `VAULT_ADDR`; with `VAULT_ADDR` set it only fails at the live Vault network call (expected — no Python-level construction error)
- [x] `manage.py check_aqueduct_settings` reports full parity against `odl_video.settings` under default, prod-like (feature flags, CDN, admins, Postgres), and keycloak-off environments — 7 documented ignores in `[tool.aqueduct] parity_ignore`
- [x] `manage.py generate_aqueduct_settings --check` reports no drift in the managed regions
- [x] `*_URL` fields resolve to `str` (including `CELERY_BROKER_URL` nested in `CACHES`); list/dict env parsing handles comma / JSON / Python-literal / `[*]`
- [x] `ruff check` covers the generated file and passes; `ruff format --check --force-exclude` skips it (the one verbatim-expr field is not format-stable); shims lint+format clean
- [x] `git diff --stat` confirms only the aqueduct files + `AGENTS.md` + `pyproject.toml`/`uv.lock` changed; `odl_video/settings.py` untouched
- [ ] Manual verification against a real Vault instance for `DevAqueductSettings` (needs live `VAULT_ADDR`/role, not verifiable in this sandbox)


blarghmatey:
django-aqueduct 0.6.0 is now published on PyPI, so I removed the `[tool.uv.sources]` local-path override and re-locked against the registry package (`uv.lock` now pins `django-aqueduct==0.6.0` from `pypi.org`, no more local/editable path). No longer blocked on the TODO. Re-verified `manage.py check --settings=odl_video.settings_aqueduct` (passes clean) and `...settings_aqueduct_dev` (fails only at the live Vault OIDC network call, as expected with no Vault server available here — no Python/import errors).

blarghmatey:
Closing for now — we'll revisit and recreate this later.

PR #1563, opened 2026-07-28
Adopt ruff 0.16 and its expanded default rule set

AhtishamShahid:
Replaces #1554.

## Why

ruff 0.16.0 turns on **413 rules by default, up from 59**. This repo has **no `[tool.ruff]` config**, so its lint policy was whatever ruff version happened to be installed — and the routine hook autoupdate in #1554 became a 277-violation lint change nobody chose. The bot autofixed what it safely could and left 106 violations with no safe fix, so #1554 cannot go green on its own.

This PR adopts the new defaults deliberately, fixes the fallout, and pins the config so a future ruff release can't do this again.

Two setup bugs that caused #1554's red checks are fixed too:

- **CI and the hook ran different ruff versions.** `uv.lock` had 0.15.5, the hook had v0.15.22, and the `<0.16` cap made them impossible to reconcile.
- **The hooks ran in the wrong order.** `ruff-format` ran before `ruff --fix`, so lint fixes un-formatted the code and nothing reformatted after. That is what broke the `Code formatting` CI step.

## What changed

| # | Commits | Files | Change | Review |
|---|---|---|---|---|
| 1 | 1 | 3 | **Config.** ruff pinned `>=0.16,<0.17`, `uv.lock` relocked, hook → v0.16.0, `ruff` → `ruff-check`, hook order swapped, `[tool.ruff.lint]` added | **Read this** |
| 2 | 1 | 43 | **Safe autofixes.** `ruff check --fix` + `format`: import sorting, f-strings, PEP 585/604 typing, `IOError`→`OSError`, `logger.warn`→`warning`. 168 fixes | Skim |
| 3 | 1 | 22 | **Unsafe autofixes.** `--unsafe-fixes` + `format`: `dict()`→`{}`, `list(x)[0]`→`next(iter(x))`, implicit `Optional`, collapsed `if`s. 52 fixes | Skim |
| 4 | 12 | 12 | **Hand fixes**, one commit per file — the 23 findings with no autofix | **Read these** |

51 files, +355 −418.

### The 12 hand-fix commits

| File | Rule | Approach |
|---|---|---|
| `ui/views.py` | BLE001 | **Behaviour change** — narrowed `except Exception` to `ValidationError`. Malformed collection keys still 400; unexpected errors now 500 instead of being mislabelled |
| `s3_sync/s3_sync.py` | PLE1205, TRY401, EXE001 | **3 real bugs** — log calls silently dropping their argument. Plus a `sys.exit()` `TypeError` on the adjacent line |
| `retranscode_collections.py` | DTZ007 | Real fix — parse UTC-aware directly, drops a redundant `make_aware` |
| `ui/permissions.py` | SIM103 | Real fix, plus rewrote an autofix that produced a double negative |
| `ui/factories.py` | UP031 | Real fix — `%d` → f-string in `Sequence` lambdas |
| `ui/views_test.py` | PLC0206 | Real fix — iterate `.items()` |
| `mail/utils_test.py` | PERF402 | Real fix — `list(chunks(...))` |
| `cloudsync/api.py` | BLE001, DTZ007 | Suppressed with rationale |
| `mail/api.py` | BLE001 | Suppressed with rationale |
| `odl_video/pipeline.py` | BLE001 | Suppressed with rationale |
| `remove_duplicate_encodings.py` | BLE001 | Suppressed with rationale |
| `sync_video_key_with_edx.py` | BLE001 | Suppressed with rationale |

The 5 suppressed `BLE001` sites are all best-effort resilience boundaries — the email batch collector, the login pipeline, and cleanup loops. Narrowing any of them would turn a partial success into a hard abort. Each has a one-line reason in the code.

Config also ignores `RUF012` globally (26 hits, all Django `ModelAdmin`/`Model` class attrs, which are mutable by design — mitxonline already ignores it) and `B017`/`DTZ001` in `*_test.py`.

## How to test

```bash
git fetch origin ruff-0.16-adoption && git checkout ruff-0.16-adoption
uv sync --locked

uv run ruff check .  --exclude '*/migrations/'          # -> All checks passed!
uv run ruff format --check . --exclude '*/migrations/'  # -> 105 files already formatted
./scripts/test/python_tests.sh                          # -> 667 passed
pre-commit run --all-files                              # -> all 15 hooks pass
```

Already verified: all four commands above, tests re-run individually after the riskier commits, and CI on this branch is green (`python-tests`, `javascript-tests`, `pre-commit.ci`, CodeQL, GitGuardian).

Nothing needs manual/browser testing — no runtime behaviour changes except `ui/views.py`, which is covered by `ui/views_test.py` (107 passed).

## Two things worth a second opinion

**`PLE1205` in `s3_sync.py` — 4 of its 7 hits are false positives.** That file logs through **logbook**, whose `Logger` uses `str.format` `{}` placeholders, while `PLE1205` counts printf `%s`. Verified against the library:

```
log.info('brace {} and {}', 'A', 'B')     ->  "brace A and B"          correct
log.info('no placeholder', 'DROPPED')     ->  "no placeholder"         arg lost
```

The 3 calls with no placeholder were the real bugs. The rule is disabled for this one file, reason recorded. (This script has no test coverage and isn't imported by the Django app — it runs standalone on lecture-capture machines.)

**8 `# noqa` comments were deleted by the new `RUF100`.** Four `BLE001` ones are genuinely unused now (0.16 suppresses `BLE001` when the exception is logged below error level). The three `SLF001` and one `ARG002` are unused because `SLF`/`ARG` aren't in the 0.16 defaults and were never selected here — they look copied from mitxonline, where those categories are on. Flagging in case you expected those protections to still apply.

## Out of scope

`s3_sync/s3_sync.py:110` passes a command *string* to `subprocess.run` without `shell=True`, which raises `FileNotFoundError`. Pre-existing, not flagged by ruff, left for a separate ticket.


s3_sync/s3_sync.py
```diff
@@ -107,11 +107,9 @@ def verify_s3_bucket_exists(s3_bucket_name):
...
     """
-    ls_s3_bucket_cmd = "aws s3api head-bucket --bucket {}".format(s3_bucket_name)
+    ls_s3_bucket_cmd = f"aws s3api head-bucket --bucket {s3_bucket_name}"
     try:
-        subprocess.run(
-            ls_s3_bucket_cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-        )
+        subprocess.run(ls_s3_bucket_cmd, check=True, capture_output=True)
     except subprocess.SubprocessError:
         logger.exception("Failed to list specified s3 bucket: {}", s3_bucket_name)
```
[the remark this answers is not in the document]
AhtishamShahid: As mentioned in the description, this is out of scope :
s3_sync/s3_sync.py:110 passes a command string to subprocess.run without shell=True, which raises FileNotFoundError. Pre-existing, not flagged by ruff, left for a separate ticket.

mail/api_test.py
```diff
@@ -64,7 +64,9 @@ def test_send_batch(self, sender_name, mock_post):
         )
         assert mock_post.called
         called_args, called_kwargs = mock_post.call_args
-        assert list(called_args)[0] == "{}/{}".format(settings.MAILGUN_URL, "messages")
+        assert next(iter(called_args)) == "{}/{}".format(
+            settings.MAILGUN_URL, "messages"
+        )
```
marslanabdulrauf: Let's use f-str to keep everything consistent

odl_video/settings.py
```diff
@@ -221,7 +221,7 @@
 # https://github.com/kennethreitz/dj-database-url
 DEFAULT_DATABASE_CONFIG = dj_database_url.parse(
     get_string(
-        "DATABASE_URL", "sqlite:///{0}".format(os.path.join(BASE_DIR, "db.sqlite3"))
+        "DATABASE_URL", "sqlite:///{}".format(os.path.join(BASE_DIR, "db.sqlite3"))
```
marslanabdulrauf: missing f-str here as well

s3_sync/s3_sync.py
```diff
@@ -40,7 +40,7 @@
 config = ConfigParser(interpolation=ExtendedInterpolation())
 try:
     config.read(settings_file)
-except IOError:
+except OSError:
```
marslanabdulrauf: `ConfigParser.read` is also excepting `OSError` and not raising it again.
Any reason why we shifted from IO to OS Error ?

marslanabdulrauf (changes requested):
Overall it looks good, just left some minor feedback.

marslanabdulrauf:
@AhtishamShahid local build is also not working for me:

image

mail/api_test.py
AhtishamShahid: Done in `09e3f43` — applied to all six occurrences in this file, not just this one, since they were identical and all six came from this branch.

I also dropped the `next(iter(...))` while I was in there:

```python
assert called_args[0] == f"{settings.MAILGUN_URL}/messages"
```

That `next(iter(called_args))` was RUF015's autofix for `list(called_args)[0]`. But `called_args` is already a tuple from `mock.call_args`, and plain indexing doesn't trigger RUF015 — only `list(x)[0]` does. So this satisfies the rule and reads better than either the original or the autofix.

Worth recording why ruff didn't catch this itself: **UP032 skips any `.format()` whose arguments are or contain a string literal**, because the fix would need nested quotes. Verified against 0.16.0 — it holds regardless of `target-version`, so the pinned config will never flag this class. Manual by necessity.

odl_video/settings.py
AhtishamShahid: Done in `09e3f43`:

```python
get_string("DATABASE_URL", f"sqlite:///{os.path.join(BASE_DIR, 'db.sqlite3')}")
```

Same UP032 quote-nesting reason as above. The formatter then collapses the call onto one line — confirmed `ruff format` leaves the inner single quotes alone, so there's no lint/format ping-pong.

Since "as well" suggested a wider sweep, I scanned all 102 non-migration files with an AST walk: **25 interpolation sites left, 19 convertible, 6 not** (template stored in a variable, so there's no literal to interpolate). ruff flags none of the 19.

I converted the **12 that live in files this PR already touches** (`af02ba0`), including the last `%`-format in the repo at `ui/utils.py:344`. The other 7 are in files this PR never opened — `ui/apps.py`, `ui/templatetags/render_bundle*.py`, `createpresets.py` — left for a follow-up rather than widening a 51-file diff. Say the word if you'd rather have them all at once.

s3_sync/s3_sync.py
AhtishamShahid: Two answers, and you're right on the second one.

**On the rename:** `IOError` has been a plain alias of `OSError` since Python 3.3, so UP024 flags it. Zero behaviour change.

**On `ConfigParser.read`:** correct — and it means the whole `except` was dead code either way. `read()` catches `OSError` per filename and continues:

```python
for filename in filenames:
    try:
        with open(filename, encoding=encoding) as fp:
            self._read(fp, filename)
    except OSError:
        continue
```

So a missing settings file never reached that `sys.exit()`. It proceeded with an empty config and died later at `config["Logs"]["logfile"]` with `KeyError: 'Logs'` instead of the intended message.

Fixed in `f875aad` using the return value, which is the list of files successfully parsed:

```python
if not config.read(settings_file):
    sys.exit("[-] Failed to read settings file")
```

| | before | after |
|---|---|---|
| missing file | proceeded, `sections=[]` | exits with the message |
| unreadable path | proceeded, `sections=[]` | exits with the message |
| valid file | `sections=['Logs']` | `sections=['Logs']` |

Good catch — this was pre-existing. The file has no test coverage, so shout if you'd rather it went to its own ticket than rode along here.

AhtishamShahid:
> local build is also not working for me:

This is caused by misc renovate PRs that were merged recently and it is not related to current PR; it requires a full Image rebuild, i.e delete node_modules and rebuild.


```bash
# In Docker (the usual case)
docker compose stop watch
rm -rf node_modules
docker compose up -d --force-recreate watch
```

marslanabdulrauf (approved):
LGTM

PR #1564, opened 2026-07-31
RTL harness and React upgrade prep 

AhtishamShahid:
## Related issues

- Implements **mitodl/hq#12636** — *OVS React 18: Phase E0+R0 — RTL harness and React upgrade prep*
- Part of epic **mitodl/hq#7958** — *React 18 upgrade* (sub-issue 1 of 8)

> Closing keywords don't work across repositories, so mitodl/hq#12636 needs to be closed manually
> when this merges.

---

## What

Phase E0 + R0 of the React 15 → 18 / Enzyme → RTL migration. Stands up React Testing Library
alongside Enzyme with a shared render helper and correct teardown, converts one pilot test file,
and clears the React-upgrade prerequisites that don't depend on Enzyme.

**React stays on `^15.6.1` and Enzyme stays installed.** Nothing here upgrades React.

## Why this ordering

RTL 12 and Enzyme coexist in one mocha process, so the migration can proceed file-by-file with the
suite green throughout. This was verified directly rather than assumed: RTL 12.1.5 renders, fires
events, resolves async queries, and unmounts correctly on `react@15.6.1` — despite `act` being
`undefined` on `react-dom@15`. RTL degrades gracefully instead of throwing.

That matters because it means **no unofficial Enzyme bridge adapter is needed at any point.**
Enzyme is removed while still on React 15, then React moves in later phases.

## Changes

| Commit | Change |
|---|---|
| `bcfad9a` | Remove unused `react-addons-shallow-compare` (zero imports; doesn't exist for React 16+) |
| `bf8e775` | Add `--exit` to the mocha runner (default/coverage/codecov; watch excluded) |
| `c023294` | `componentWillReceiveProps` → `componentDidUpdate` in `Dialog`/`Drawer`/`Menu` |
| `cbcdf77` | Add RTL 12.1.5 / dom 8.20.1 / user-event 14.6.1, pinned exact, + harness test |
| `e80dfb6` | Repo-wide `prettier-eslint` reformat (isolated, no behaviour change) |
| `86b1e29` | Call RTL `cleanup()` in global teardown; drop dead `#integration_test_div` branch |
| `7fd9380` | Add shared `renderWithProviders` RTL helper |
| `f8284e6` | Convert `ErrorPage_test.js` from Enzyme to RTL (pilot / reference pattern) |

New files under `static/js/testUtils/`: `renderWithProviders.js` plus `harness_test.js`,
`teardown_test.js`, `renderWithProviders_test.js`.

## Verification

- **Suite: 480 → 492 passing**, run twice with identical counts
- `npm run lint` clean; `npm run fmt:check` clean
- `NODE_ENV=production npm run build` compiles (7 pre-existing Sass warnings from `@material` 0.33)
- Enzyme test files: **32 → 31**
- No new entries in the `js_test.sh` stderr allowlist (still 7)

**The pilot conversion was mutation-checked by hand.** Two bugs were injected into `ErrorPage.js`
(corrupted title text, renamed `.message` class) and each killed all 3 tests — confirming the
converted test detects breakage rather than merely passing.

**The three lifecycle conversions were verified in a browser**, since no test covers them (they're
reached only via `.instance()`, the API being removed). Each of the drawer, Edit Collection dialog,
and per-video `more_vert` menu was opened, closed, **and reopened** — the reopen being the case
that matters, as `componentDidUpdate` doesn't fire on mount and the prop comparison inverted. All
passed, with zero console errors across a full cycle.

## Reviewer notes

- **`e80dfb6` is the noisy one.** The repo had drifted from its own formatter (last repo-wide
  format was #398; `fmt:check` isn't a CI gate), so `npm run fmt` rewrote 36 unrelated files.
  Landed as its own commit so the other seven stay reviewable. Of the 36, 25 differ only in
  whitespace; the other 11 were hand-checked as prettier normalizations (quote style, trailing
  commas, import/expression reflow). **Review it with whitespace hidden.**
- `@testing-library/react` is pinned to exactly `12.1.5` with no caret — v13+ requires React 18.
  Please don't let a caret in until React reaches 18.

## Out of scope / follow-ups

Two pre-existing environment bugs surfaced while running this. Neither is caused by these changes
and neither is fixed here:

1. **`Dockerfile-node` ships node 24.17.0 while `package.json` pins 24.16.0.** Yarn 1 hard-fails on
   the mismatch, so the `watch` container crash-loops and every bundle 404s — `docker-compose up`
   currently yields a blank app. Workaround is `YARN_IGNORE_ENGINES=1` in a local `.env`.
2. `docker-compose.yml` still carries the obsolete `version:` attribute, warned on every command.

Also noted for a future ticket: `global_init.js` appends a `<script>` tag at module load "to appease
videojs-youtube", but the `afterEach` wipes `document.body.innerHTML` — so it's gone after the first
test. Not addressed here because fixing it changes behaviour for tests this PR doesn't touch.

## Next

| Issue | Phase |
|---|---|
| mitodl/hq#12637 | Stryker mutation baseline + CI ledger |
| mitodl/hq#12638 | Convert Tier 1 + Tier 2 (17 files) |
| mitodl/hq#12639 | VideoPlayer controller extraction |
| mitodl/hq#12640 | Remaining conversions, **remove Enzyme** |
| mitodl/hq#12641 | React 16.14 |
| mitodl/hq#12642 | React 17.0.2 |
| mitodl/hq#12643 | React 18.3.1 |

Recommend landing the mutation-testing baseline (mitodl/hq#12637) before the bulk conversions in
mitodl/hq#12638 — it's the only objective check that converted tests still assert anything.

### DIFF
--- a/static/js/containers/ErrorPage_test.js
+++ b/static/js/containers/ErrorPage_test.js
@@ -2,16 +2,16 @@
 /* global SETTINGS */
 import React from "react"
 import sinon from "sinon"
-import { mount } from "enzyme"
 import { assert } from "chai"
+import { screen } from "@testing-library/react"
 import configureTestStore from "redux-asserts"
-import { Provider } from "react-redux"
 
 import * as api from "../lib/api"
 import ErrorPage from "./ErrorPage"
 import { actions } from "../actions"
 import rootReducer from "../reducers"
 import { makeCollection } from "../factories/collection"
+import renderWithProviders from "../testUtils/renderWithProviders"
 
 describe("ErrorPage", () => {
   let sandbox, store, collections, listenForActions
@@ -32,7 +32,7 @@ describe("ErrorPage", () => {
   })
 
   const renderPage = async (props = {}) => {
-    let wrapper
+    let result
 
     await listenForActions(
       [
@@ -40,15 +40,10 @@ describe("ErrorPage", () => {
         actions.collectionsList.get.successType
       ],
       () => {
-        wrapper = mount(
-          <Provider store={store}>
-            <ErrorPage {...props} />
-          </Provider>
-        )
+        result = renderWithProviders(<ErrorPage {...props} />, { store })
       }
     )
-    if (!wrapper) throw new Error("Never will happen, make flow happy")
-    return wrapper
+    return result
   }
 
   // eslint-disable-next-line no-unused-vars
@@ -72,21 +67,18 @@ describe("ErrorPage", () => {
   ]) {
     it(`renders an error for status=${status}`, async () => {
       SETTINGS.status_code = status
-      const wrapper = await renderPage()
-      assert.equal(
-        wrapper
-          .find(".error-page .title")
-          .text()
-          .trim(),
-        title
-      )
-      assert.equal(
-        wrapper
-          .find(".error-page .message")
-          .text()
-          .trim(),
-        message
-      )
+      const { container } = await renderPage()
+
+      // The title is a single text node, so the accessible query works.
+      assert.isNotNull(screen.getByText(title))
+
+      // The message is deliberately not queried with getByText: for status
+      // 500 the copy is split across a <span>, an <a>, and a trailing text
+      // node, so no single element holds the whole string. textContent on
+      // the container element is the documented RTL fallback for that case.
+      const messageEl = container.querySelector(".error-page .message")
+      assert.isNotNull(messageEl, "expected a .message element to render")
+      assert.equal(messageEl.textContent.trim(), message)
     })
   }
 })
--- a/static/js/testUtils/harness_test.js
+++ b/static/js/testUtils/harness_test.js
@@ -1,0 +1,88 @@
+import React from "react"
+import { assert } from "chai"
+import { render, screen, fireEvent, cleanup } from "@testing-library/react"
+
+class Greeter extends React.Component {
+  render() {
+    return <h1>Hello {this.props.name}</h1>
+  }
+}
+
+class Counter extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { n: 0 }
+  }
+
+  render() {
+    return (
+      <div>
+        <span data-testid="n">{String(this.state.n)}</span>
+        <button onClick={() => this.setState({ n: this.state.n + 1 })}>
+          inc
+        </button>
+      </div>
+    )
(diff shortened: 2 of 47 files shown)


static/js/containers/ErrorPage_test.js
```diff
@@ -72,21 +67,18 @@ describe("ErrorPage", () => {
...
-        wrapper
-          .find(".error-page .message")
-          .text()
-          .trim(),
-        message
-      )
+      const { container } = await renderPage()
+
+      // The title is a single text node, so the accessible query works.
+      assert.isNotNull(screen.getByText(title))
```
umar8hassan: `screen.getByText(title)` isn't scoped to `.error-page .title` the way the old Enzyme query was. It matches the title text anywhere on the page. maybe something like:

```suggestion
      assert.isNotNull(within(container.querySelector(".error-page")).getByText(title))
```

static/js/testUtils/teardown_test.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+describe("global teardown", () => {
+  it("mounts a component and lets the afterEach hook run", () => {
+    render(<UnmountRecorder />)
+    assert.equal(unmountCount, 0)
+  })
+
+  it("unmounted the previous test's tree before this test started", () => {
```
umar8hassan: This test would fail for a single-test invocation. Render and unmount within the same `it`, rather than relying on execution order across two tests.

static/js/testUtils/harness_test.js
```diff
@@ -0,0 +1,88 @@
...
+
+  it("resolves findBy* after an async setState", async () => {
+    render(<AsyncLoader />)
+    assert.isNotNull(await screen.findByText("loaded"))
+  })
+
+  it("fires componentWillUnmount on cleanup()", () => {
+    let unmounted = 0
+
+    class Recorder extends React.Component {
```
umar8hassan: Small duplication: this `Recorder` class and `teardown_test.js`'s `UnmountRecorder` are the same. Could merge into one shared test util helper to avoid duplication

umar8hassan (approved):
Mostly looks Good. I have suggested a few small changes.

AhtishamShahid:
Hi, @umar8hassan  I have updated the PR. Can you please take a look? Thanks.

